PR #15, opened 2015-03-12
Update Spanish locales

tchx84:
Signed-off-by: Martin Abente Lahaye <tch@sugarlabs.org>


llaske:
Hi Martin,
There is few more strings to translate starting here: https://github.com/llaske/sugarizer/blob/master/locale.ini#L349
Could you have a look on it ?

        Lionel.

tchx84:
Yeah sure!

PR #16, opened 2015-03-13
Deutsch Translation

francocorreasosa:
Germany Locale


bertfreudenberg:
Please no. This looks machine-translated. Franco, I'm sorry, being a native speaker this is worse than keeping the English.

PR #21, opened 2015-04-05
Added Japanese translation

naofum:


llaske:
Thanks. Nice work !

llaske:
Hi,
There is few more strings to translate starting here: https://github.com/llaske/sugarizer/blob/master/locale.ini#L948
Could you have a look on it ?

        Lionel

PR #25, opened 2015-07-24
Calculate.activity : Fixing input reset when clicking on trigo functions, disable user selection of results

mikklfr:


llaske:
Strange: you're sending your pull request as "mikklfr" but you've commited using "ohayon_m". I guess it's the reason why you don't appear in the contributor list.

PR #27, opened 2015-08-16
Added TurtleBlocksJS

mikklfr:


llaske:
You removed the old TurtleJS from the activities.json but you need to remove the whole content of the old activity too.

mikklfr:
It seems that github has trouble handling such a big diff.

llaske:
Right, the old directory and all its content has disappeared but github show only one file removed

PR #30, opened 2016-03-05
More efficient/cleaner rewriting of activity_custom.js of WelcomeWebActivity

lupin012345:
##Modifications on activity_custom.js

I removed the "Actualizar()" function and replaced it with a set of functions to update the #here div. (update(), calling display_slide(), calling setDots()).

There is now less of and some more explicit global variables to work with and I refactored the file to replace big code blocks by small and more efficient functions without repeated code (previous, next, change_topic for example). 

The functions and variables names are now in English and commented.


llaske:
Nice. Merged manually due to update in config.xml for Windows 10.

PR #31, opened 2016-03-06
Fix the vertical position of the search bar to center

erilyth:


llaske:
Thanks. Look better like that.

PR #32, opened 2016-03-06
Add icon hover effects

erilyth:
In Sugar, when we hover over activity icons there is a hover effect that is applied to them. This is not present in Sugarizer, and I think it should be added.
It gives the user a sense of where he is moving his cursor and where he can click to start an activity.


llaske:
Nice idea. It's something I had in mind though only interesting with a mouse.

PR #33, opened 2016-03-07
Changed Android minSdkVersion to avoid compilation errors

lupin012345:
This patch sets the android-minSdkVersion to 14 instead of 13 so it matches the used libs minSdkVersion and to avoid compilation error with cordova


llaske:
Fixed manually due to merge conflict.

PR #36, opened 2016-03-09
Fix issue of journal showing entries and empty error dialog

erilyth:
When we use the journal and then toggle between activity selection, even though activities are being displayed, an error message is being displayed above the activities.

screenshot from 2016-03-09 19 15 14

### DIFF
--- a/js/journal.js
+++ b/js/journal.js
@@ -86,24 +86,29 @@ enyo.kind({
 		this.$.message.show();
 		this.$.nofilter.show();
 		this.empty = (!this.getToolbar().hasFilter() && !this.loadingError && this.journal.length == 0);
+		this.$.empty.hide();
+		this.$.message.hide();
+		this.$.nofilter.hide();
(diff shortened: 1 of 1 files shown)


llaske:
Not sure how to reproduce the issue. Could you explain ?

erilyth:
@llaske To reproduce,
1)Go to the journal
2)Then select some activity in the activity filter
3)Then go back and select "Anything" in the activity filter again.

All the entries should be displayed which works fine, but the journal empty dialog is also present which should not happen.

llaske:
Sorry, not able to reproduce. Could you try to reproduce it on http://server.sugarizer.org and describe exactly how.

erilyth:
@llaske Oh, I am unable to reproduce the error on there. That's strange, does it also run the same code that the repository has currently on the master branch?

llaske:
Don't know I didn't do any update here. May be a drawing glitch

PR #38, opened 2016-03-11
Fix missing closing input tag error

erilyth:
This gives an error message in the console.


llaske:
Hmmm. Need to be fixed also in all activities...

PR #56, opened 2016-04-20
Maze Activity, display a score

erilyth:
When I was using the maze activity, I could not keep track of the number of levels I completed, so I thought adding a score would make it easier for the users to know how many levels they have completed.
A total of 4 scores will be displayed, one for each player.

screenshot from 2016-04-20 19 54 33

### DIFF
--- a/activities/MazeWeb.activity/js/activity.js
+++ b/activities/MazeWeb.activity/js/activity.js
@@ -40,6 +40,11 @@ define(function (require) {
 
         var gameSize = 60;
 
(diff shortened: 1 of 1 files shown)


llaske:
Not a big fan of this idea. It's very intrusive on the screen to have always displayed score.
It will be better to do that in a same way of the Sugar version of the activity: name and buddy color of the winner is displayed shortly at the end of each level with the time (not the score) that the player have done.

PR #57, opened 2016-04-20
Maze Activity, add a 'Help' palette

erilyth:
A palette to give the users instructions on how to play the game would be helpful.
Initially when I used the activity, I didn't know that there were 4 players, until I went through the code.

Would a palette be better for this purpose or should I just show the instructions on the page itself overlapping the canvas until a user performs an action (lets say a click) to hide the instructions?


llaske:
You're right, it's not a documented feature but it works only for device with keyboard. I guess this activity will be mainly use with the mouse or with the finger on touch devices. So I'm not sure it's so important to do help on this. 
Plus, if there is a help palette, it must be localized.

davelab6:
(I think its helpful to leave pull requests open until they are ready to merge; you can set pull requests to "auto-squash" all commits into one if you like, or pick the option at merge time, in https://github.com/llaske/sugarizer/settings :)

davelab6:
> it works only for device with keyboard

Perhaps a multi user mode could work with touch... eg, opposite sides/corners of the screen to tap for each person sat/stood around the screen

Issue #54, opened 2016-04-18
Update etoys in Sugarizer


davelab6:
Looks like there is already a port of etoys to js :) 

http://bertfreudenberg.github.io/SqueakJS/etoys/

davelab6:
http://snap.berkeley.edu/snapsource/snap.html also fun

llaske:
Etoys is already in Sugarizer ! thought not in the favorite view because it's not fully working on all platforms.

davelab6:
Great! And snap? :)

llaske:
Ooops not seen Snap. Will test it.

davelab6:
Also http://lively-kernel.org

davelab6:
l-k moved on to http://lively-web.org and is being actively developed

Great intro: https://www.youtube.com/watch?v=QTJRwKOFddc

davelab6:
Also http://tinlizzie.org/chalkboard/#Home

davelab6:
So, Lively is done by "the" Dan Ingalls, who was the guy who actually implement Smalltalk in the 70s and 80s with Alan Kay. http://lively-web.org/projects.html is a list of examples of it in use.

davelab6:
Snap led me to http://bjc.berkeley.edu - very nice.

davelab6:
http://aprt.us also very nice

davelab6:
https://github.com/cdglabs/moonchild from the appertus org

llaske:
Hmm. Most of these projects are not responsive and for most projects I don't see a clear way to integrate it directly into Sugarizer. I think it's better to let them run side by side with Sugarizer.
Plus, on Android for example, we could find Scratch JR which is most interesting than Snap in my opinion.
It's why I'm expect to have in Sugarizer OS a way to launch directly a native application from Sugarizer. So we could have the Scratch JR icon on the Sugarizer home view without the need to really integrate it into Sugarizer.

davelab6:
>> Etoys is already in Sugarizer
> 
> Most of these projects are not responsive

Is http://server.sugarizer.org/activities/Etoys.activity/index.html responsive? 

> I don't see a clear way to integrate it directly into Sugarizer

Is Etoys integrated directly into Sugarizer?

> It's why I'm expect to have in Sugarizer OS a way to launch directly a native application from Sugarizer. 

:) My idea of #55 is that "Sugarizer Thin Client" ("Sugarizer Web"? :) would also launch directly related non-activities.

[...]

PR #64, opened 2016-05-03
Etoys: update to SqueakJS 0.8.2

bertfreudenberg:
resolves issue #54


llaske:
The activity freeze after displaying the main page :-(
May be it's related to a 404 error I have: "Download failed resources/locale/en/LC_MESSAGES/Monticello.mo".
Plus I'm slightly worried about activity size: 70Mb, the previous was only 2Mb. Is there a way to limit size ?

bertfreudenberg:
Previously, all the Etoys files were served from my webserver, the activity only contained the runtime. This means that running sugarizer would require a constant internet connection to my server (or some server that hosts these files). These files include Etoys itself, example projects, and translation files.

I did forget to include the translation files. I just added them. Please try again.

bertfreudenberg:
Regarding activity size: Abecedarium is 136 MB and KAView is 70 MB so Etoys with 90 MB isn't entirely out-of-line. We could host some files elsewhere but then Sugarizer would depend on another service which is probably not a good idea, right?

davelab6:
@bertfreudenberg do these require all the data to be transferred before they can be used, as etoys does? 

eToys does, but I think it just needs a modest loading progress indicator UI to make it totally acceptable - and since eToys has - from what I can see - unparalleled educational depth, I would love to see it more prominently promoted in Sugarizer :)

bertfreudenberg:
Etoys downloads ~22 MB on the first start and fetches more files on demand. All of these are cached locally in IndexedDB so repeated starts are faster (the "downloading" progress bar should only be visible on first start).

davelab6:
AWESOME!!! It seems every time I have an idea about what should be done around Sugar Labs, I found it was already done by just not well positioned, hehheh :)

llaske:
Regarding KAView it will be removed in the next version. It was not a good idea to embed videos. It will be replaced by VideoViewer a single player without any content.
Regarding Abecedarium, if it don't find sounds/images locally, it automatically access it on line (same for FoodChain now). So when I'm generating the mobile version all sounds are removed from the package.
With that I could today generate a full Sugarizer Android APK package from only 40Mb.
In fact, I have no choice: Google Play Store, Amazon Store and Apple Store have limitation on package size (between 100-150mb max).
So unfortunately 90Mb for EToys is an issue.

davelab6:
>  unfortunately 90Mb for EToys is an issue

Its an issue for the Client only; the Server is fine, even over the regular internet.

davelab6:
Therefore I recommend writing the Client build script to leave out KAView and Abecedarium and eToys.

bertfreudenberg:
@llaske we can split Etoys into a client and a server part. Basically the whole `resources/` directory (88 MB) could remain on the server. How would we go about that? The `index.html` needs t know the server URL.

llaske:
Nice idea to split between server and client. We could consider that the URL will be "http://server.sugarizer.org/activities/EToys.activity/". It's exactly what is done in Abecedarium activity: https://github.com/llaske/sugarizer/blob/master/activities/Abecedarium.activity/config.js#L5

bertfreudenberg:
I think all that is needed to make that work is to replace `"resources/etoys.image"` in https://github.com/bertfreudenberg/sugarizer/blob/master/activities/Etoys.activity/index.html#L30 with `"http://server.sugarizer.org/activities/Etoys.activity/resources/etoys.image"` in your build script. Then you can bundle everything except the `resources` folder into your local installation, that's only about 2 MB (and if you want it even smaller, leave out the `squeak.min.js.map` file, which is only needed for debugging). To start Etoys the first time it will need internet access, however.

llaske:
Nice ! it perfectly works. I'm going to update my packaging process to handle that.

@bertfreudenberg one more question: do you think it's to possible to use the journal in some way in the activity ?

PR #76, opened 2016-10-25
Translation to portuguese Brazil

afmachado:
Hi @llaske here is the pull request with the translation to portuguese Brazil. Thanks!


llaske:
Great. Thanks.

llaske:
There is few more strings to translate started here: https://github.com/llaske/sugarizer/blob/master/locale.ini#L1196 
Could you have a look on this ?
Could you tell me who I have to credit for the translation ?

PR #82, opened 2016-11-30
Added delay adjuster in toolbar of WelcomeWebActivity

tradzik:
user is able to choose time between 0,1-10 seconds and apply it using designated button. Done as a task at Google Code-in 2016.


i5o:
Can you add limits to the adjuster? I was able to go below 0, but when applying it returned me to 0.1, can you fix it? Good job btw

tradzik:
@i5o Thanks. Actually there are limits - when you write anything out of range 0,1-10 seconds and press apply, it will autocorrect you to border value. It is possible to ban arrows to go out of this range, but you still will be able to write value out of range. I can ban it with onchange event of jquery and checking the value of the field after every key press, but I think it would be an overregulation.

tradzik:
@llaske I have implemented it as a palette as you wanted me to make it. I hope it is alright. I didn't squash commits, because they change different things. Screenshot:

Screen

i5o:
Good job. I know that @llaske wants it to be similar/same to Speak activity.
But I think that the palette is too big..

Also >sorry Lionel<  I would like to see the time in the palette, I know you can see it on the tooltip.. but it would look more obvious!

Another thing, it doesnt look "responsive"? Or maybe I did something wrong, but when using it with the brwoser console open, you can not see the text D:
I mean, I'll attach screenshots

welcomeweb_0
welcomeweb_1

tradzik:
@i5o Thanks for your comment and time spent on testing my solution! 

I have reduced the size of palette. Looks much nicer right now. 

The slider is partialy responsible. It generally fits the screen and palette even on small devices, but sometimes overrides the label. It is the same behavior like in Speak activity.

I am still not sure about writing the time near to the slider. I think simpler design, is better design. The current value is being displayed on hover of slider. 

Changes have been squashed to the last commit.

Screenshot:

Screen

i5o:
Yay, thanks.

It looks enough and ok for the GCI Task, can you please give me the link?

tradzik:
@i5o Thanks! Sure, the link is https://codein.withgoogle.com/dashboard/task-instances/5667036470444032/

llaske:
Very good work. Thanks.

PR #83, opened 2016-12-01
Localized ChatPrototype

kiridix:


llaske:
Thanks but some other strings had to be translated: in the HTML page and in the JavaScript file.

kiridix:
How can I localize a string in the HTML page?

llaske:
Guess your question is about Status ? Do the same that for other strings, get element into the document and update the innerHTML value.
If I'm right, it seems that your PR include Arabic translation update. If so, please remove the translation from this PR and do a specific PR for that. And do not forget to claim the related GCI task :-)

kiridix:
Yes, I have to finish this one to claim other one

kiridix:
Ive added localization of Status, maybe it is ready but tell what you think

llaske:
Good work. Two things is missing:
- l10n variable is not declared in activity.js (it should be the second parameter of the function)
- you should init the locale with Sugarizer or browser settings, here how you could do https://github.com/llaske/sugarizer/blob/master/activities/Clock.activity/js/loader.js#L12

llaske:
Merged manually in https://github.com/llaske/sugarizer/commit/25657cba33eaf7a2b77c4814d9b7e1615167bc26

PR #84, opened 2016-12-01
created webL10n.js for clock

samswag:
Starting a my  GCI project with clock activity


llaske:
The webL10n.js is already in the lib directory.

samswag:
the problem is that, i can't figure out what strings i can translate.
Because the activity itself don't display much of text

llaske:
The idea is to translate the time string:

clock

I suggest using the moment.js library, already used in the Moon activity.

samswag:
oh! let me look at it

PR #87, opened 2016-12-01
Localization for Clock activity

tradzik:
I have implemented full localization of Clock Activity, including dates display, using webL10n built into sugar webcore.


i5o:
To be honest I thought this was impossible (lawl!)

Good job. Is this a GCI task or something?

tradzik:
Thanks @i5o ! Then I even more happy because of my job. Yeah, it is based on a GCI task.

llaske:
Very good job. BTW It could be nice to detect automatically the language. You could retrieve code for that in getSettings method of TankOp https://github.com/llaske/sugarizer/blob/master/activities/TankOp.activity/settings.js#L5

tradzik:
Thanks! I have implemented the getSettings() in TankOp, I am gonna add it
also here. It is useful if remembered settings are different from the
current navigator (browser) settings.

llaske:
Don't forget to take the task in GCI !

tradzik:
Thank you, I will claim it soon! I have added getSettings and squashed commits. IMHO it is now mergable :smile:

PR #88, opened 2016-12-02
updated the js file

samswag:
I made it translatable

### DIFF
--- a/activities/GetThingsDone.activity/js/activity.js
+++ b/activities/GetThingsDone.activity/js/activity.js
@@ -82,9 +82,9 @@ define(["sugar-web/activity/activity","sugar-web/datastore","activity/model","ac
 		
         });
         function localize() {
-		        document.getElementById("stop-button").title = l10n.get("stop");
-		        document.getElementById('activity-button').title = l10n.get("GetThingsDone");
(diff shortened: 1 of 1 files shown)


activities/GetThingsDone.activity/js/activity.js
```diff
@@ -79,9 +79,13 @@ define(["sugar-web/activity/activity","sugar-web/datastore","activity/model","ac
             if (target.className.indexOf('toggle') > -1) {
                 todo.controller.toggleComplete(lookupId(target), target);
             }
-
+
```
tradzik: unneeded whitespaces

activities/GetThingsDone.activity/js/activity.js
```diff
@@ -79,9 +79,13 @@ define(["sugar-web/activity/activity","sugar-web/datastore","activity/model","ac
             if (target.className.indexOf('toggle') > -1) {
                 todo.controller.toggleComplete(lookupId(target), target);
             }
-
+		
         });
-
+        function localize() {
```
tradzik: function exists,  but is never called (doesn't take effect)

llaske:
Please, put all files related to the same patch in the same PR.

PR #90, opened 2016-12-02
created locale.ini

samswag:


llaske:
Please, put all files related to the same patch in the same PR.

PR #99, opened 2017-01-17
Four new activities

ccosse:
Here are four new activities.  I would like to work with you to integrate these.  Is there a style guidelines page somewhere?  
The activities are:
1. TuxMathScrabble: math scrabble activity  (original Python version has animated penguins, html/js not yet due to space limitations on screen)
2. MakeItTrue: algebra training activity; choose values to make the given equality true.
3. GeoHunt: a maps-based scavenger hunt
4. Echo: a simple app which requires user to type the expression given; designed for use with idioms dictionary, but also includes chemical elements and wordnet dictionaries.
tuxmathscrabble_sugarizer
geohunt_01
make_it_true_02
echo_01


llaske:
Very nice.
I specifically like GeoHunt and TuxMathScrabble. 
MakeItTrue is globally the same concept than TuxMathScrabble. Not sure to understand the idea of Echo.
However, I think these activities need some adaptation before be integrated into Sugarizer.
About the style but mostly about complexity level. Sugarizer is for children from 6-12 years old so I'm afraid rules are too complex.
If you're okay I'm proposing that we work together to improve your activities and integrate it in Sugarizer.
I suggest to start by GeoHunt. 
Don't do a global PR for the moment, just do a separate repository for each activity and tell me when it's done. You could do a PR later. 

Note that there is an issue on the toolbar on all yout activities. The STOP button should go back to the home view. Today it does nothing/ I guess it's because you do not call activity.setup() - see https://github.com/llaske/sugarizer/blob/master/activities/ActivityTemplate/js/activity.js#L7 - It's here that you have to call your first JavaScript entry point be sure to have the default STOP event handled.

ccosse:
Hi,

>If you're okay I'm proposing that we work together to improve your activities and integrate it in Sugarizer.

That would be great, I will create separate repositories and begin with some changes to GeoHunt, including the Stop button issue.  

Bastien Guerry told me about your work and Sugarizer last week.  I have a project in mind which could build upon your platform and pay activity developers, and thereby provide incentive to attract more activity developers.  I hope we can discuss it when you have some time.  [Here](https://github.com/ccosse/NetDispenserEcoSystem/blob/master/Credit-Meter-Assisted-Learning.pdf) is a link to presentation slides for the idea.

llaske:
Yes Bastien told about me about that, sure we could discuss about your incentive project.

Regarding GeoHunt, here some ideas/remarks I have:
- Limit the number of maps. I guess we could limit it to the world map. Plus it will reduce the full size of activity, it's more than 80Mb today. I'm concerning about the size of Sugarizer because it could be use on small device or with small bandwith.
- So, with only one map we could remove the settings button and the left panel. I think we could move the Play button to the toolbar too.
- We could add a game mode where the user could just color countries as he want. It's important to have creative features in activities. For example the teacher could use the feature to ask children to color countries they visit. Do you think it's easy ? We could reuse the color palette that exist in other Sugarizer activities.
- It's important to plan localization of activities (name of countries), I don't know if you're familiar with that but we're using webL10n.js library to handle that.
- I've seen a JavaScript error on Chrome when the user click on a country. It works on Firefox thought.

ccosse:
Hi, yes I agree with all of your points.  I created a new repo and will build-up from scratch.  I also want to change the name to ColorMyWorld, do you agree that it's a better name?  I have another version 0.1 of such app (screenshot below), but it has a 8M data file and too many small islands included.  The good thing of that app is that I extracted the outline of each country and created an image with it, which serves as a visual hint as to what you are looking for.  So, combining all these ideas and versions should make a nice result.  

- I will try to understand use of the color palette and webL10n.js.  I'm familiar with i18n through Django, but never used webL10n.js.  

- The current GeoJSON file with countries can be reduced and re-factored ... another idea is to tag countries by region, like France would have a Europe tag, or something.  Any kind of meta data can be added to the GeoJSON format, so lots of possibilities, but will keep it simple and small as possible for now.

image

- [Here](http://www.asymptopia.org/colormyworld/) is the v0.1 of ColorMyWorld from 2015, and [here](https://github.com/ccosse/colormyworld) is the new repo starting from scratch.

llaske:
Very cool. And ColorMyWorld is a very good name.
Very good idea to tag countries. We could propose to filter on it. To avoid too complexity, may be we could limit tagging to continents. 
I'm closing this PR and put your new repo as favorite. Tell me how I can help you to avoid to lost too much time on Sugarizer stuff.

PR #104, opened 2017-03-02
Issue 69: Red face visible in the beginning

DarkPrince304:
This should fix the issue 69. Now in the beginning instead of the grey circle which denotes the player, the user will be able to see a red face.

Would like to get this reviewed and tested :)


llaske:
It's better to use the color of the mouse player instead of "red". Else, when the mouse player move, the face color change again.
Please send your PR to the new "dev" branch instead of "master".

PR #113, opened 2017-03-13
Add item reordering in Getting Things Done activity

### LINKED ISSUE #68: Get Things Done: Reorder Items
http://server.sugarizer.org/activities/GetThingsDone.activity/index.html doesn't currently allow for reordering items; I suggest this is the next biggest improvement needed

NikolayOG:
Fixes #68


llaske:
Work on Chrome and Safari but not on Firefox here.
Could you check on your side ?

NikolayOG:
@llaske The problem was in drag function. Basically for Firefox we have to explicitly say what happens with dataTransfer. Line 35 in activity.js.

llaske:
Works now. Cool. Thanks

PR #114, opened 2017-03-15
Blockrain activity version 1.0

DarkPrince304:
This PR has the complete activity Blockrain JS for sugarizer. It will work for both desktop and mobile sugarizer environment. @llaske Would like it to get this tested from you and everyone else. Please comment here if anything needs to be changed.


DarkPrince304:
Thank you @llaske.

I would like to discuss something about the arrow buttons. I have positioned the arrow buttons like this for two reasons.

1. For different, rather smaller screen sizes the buttons may seem to overlap

2. For devices such as an ipad the person usually holds it using both the hands so in such a case using both hands would be easier I guess?

DarkPrince304:
Just need your reviews about the positioning of the arrow keys but apart from that all the other changes have been done and ready. Would like your reviews @llaske @mikklfr :)

llaske:
Very nice.
Still an issue however: when I'm resizing the window, horizontal and vertical scrollbar appears

image

llaske:
Solve now.
Could you set as default the theme on my screen capture ?

llaske:
Cool. I will test it in Android, iOS and Windows 10 to ensure it works then I will merge it.
Thanks for your work.

PR #118, opened 2017-04-02
upgrade to openlayers v4.0.1; address memory issues

ccosse:
Changed the 6.6M world.geojson to a 250k world.geojson.  Zooming in/out with hi-resolution vector layer was consuming a lot of memory (2G).  Only the whole world layer is low-res now.  Others, like Africa et al are still hi-res.  Need a Caribbean region.


llaske:
I've got an issue when reloading from datastore. All colored countries become black when the map is reload.

PR #119, opened 2017-04-15
implemented sorting in List View

sauravpratihar:
current version of sugarizer doesn't display activity list in sorted order. 
This commit will display the list of activities in ascending order of their names in List View.


llaske:
Nice idea. It's already the case in Sugarizer OS.
Merged in https://github.com/llaske/sugarizer/commit/b13c160349eca6b61ef95c748a449d4f0b3c63b1

PR #124, opened 2017-06-17
Contribute Jappy Activity

icarito:
From the rainforest with ❤️


llaske:
Very cool. I've tested successfully on Chrome, Firefox and Safari.
I've got an issue on EDGE (Windows 10) with Memorize (see bellow)

memorize_jappy

Could you test on your side ?

icarito:
I don't have a W10 system at or near my location. I could try some virtual think like Browserstack...

icarito:
It looked like Edge won't let writing a script into the iframe.

icarito:
https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/9291485/ ?

icarito:
Dear @llaske ,
Finally I managed to reproduce.
I think it's a problem with my upstream transpiler for the _random_ function.
Nevertheless this is with Edge 14. With 15 Jappy doesn't work at all because of what [looks like a bug in Edge](https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/9291485/).
I can't support them.

llaske:
Bad news :-(
I'm going to do some more testing and I will probably mask the activity on the Windows 10 version. There is a similar issue on the Speak activity: it works everywhere except on iOS and there is nothing we can do for the moment on that. So the Speak activity is not visible by default on iOS.

llaske:
Don't work neither on ChromeOS. Here is the error message:

image

I guess it's related to dynamic JavaScript eval.

One other issue I've got: I can't minify your files using Grunt. Minifying is interesting in JavaScript because it allow to have better performance. The problem here is due to the "riot.config.js" that use an ES6 syntax not supported by Grunt. I guess this file is not use at runtime: is it possible to rename it without the ".js" extension to avoid this issue ?

icarito:
I also don't have Chrome or ChromeOS but it looks from [this question](https://stackoverflow.com/questions/26242682/unsafe-eval-on-chrome-extension) on StackOverflow suggest that if you want to support them, you need to tweak Sugarizer Chrome [manifest.json](https://github.com/llaske/sugarizer/blob/master/manifest.json) file to allow arbitrary eval, adding ` "content_security_policy": "script-src 'self' 'unsafe-eval'; object-src 'self'" ` to it.


On the topic of Grunt not liking source files, I just committed removing the actual sources and build artifacts, with a note to find them at my repo. Grunt should be happy!

llaske:
Hmmm. Unfortunately the "content_security_policy" is only allowed for Chrome Extension, you can't release CSP for Chrome App: https://developer.chrome.com/apps/contentSecurityPolicy#what .
So I'm afraid Jappy will not work as a Chrome App and/or in Chrome OS.

BTW Grunt is happy now :-)

icarito:
Although Windows 10 or Chrome OS Apps users are not my target audience, perhaps @kovidgoyal who is the author of RapydScript NG would like to know:

Kovid, @llaske is the author of Sugarizer and has been testing Jappy on different platforms. 
- Edge 14 has a problem with the `random` function; for some reason we're turning `return n / 18446744073709551616` into `return n / 02(e+21)` which Edge fails to understand.
- ChromeOS seems to complain about unsafe eval, but I'm not sure why that happens.

Thanks again for writing RapydScript-NG and your support!

kovidgoyal:
Not sure I understand the issues. For instance:

```
echo 'n / 18446744073709551616' | rapydscript -m
```

gives:

```
n / 0x10000000000000000
```

on Chrome, Firefox and nodejs. I assume you see 02(e+21) when actually compiling code on edge itself? I have to say that I've tried using Edge to run the rapydscript compiler. I have no idea what issues might result. In general, I have found Edge to be a fairly buggy browser when it comes to full support of modern web standards. For instance, I had to drop support for it from the new calibre content server because it lacks various things, such as support for srcdoc in iframe and array keys in IndexedDB. I'm not surprised that its JS engine is behind as well. 

As for eval, seems odd, since the rapydscript REPL does not show any errors related to eval when run in Chrome itself. I'm afraid I dont know anything about ChromeOS.

kovidgoyal:
I any case, this should fix the issue with random(), bu avoiding compilation of that expression:

https://github.com/kovidgoyal/rapydscript-ng/commit/ac12d7e202b2816c5596d6e0ad9774cd41017d3c

icarito:
@kovidgoyal you are a wonder, I can confirm this fixes the Edge 14 issue!
@llaske one less issue for merging! Even Google has abandoned ChromeOS, I guess we can ignore it.

llaske:
I confirm too, it now works on Edge. Cool.
Regarding Chrome, I'm not agree with you, Google has not abandoned ChromeOS at all ! Plus Chrome application is a very simple way to install Sugarizer locally when you've got a Chrome browser.
BTW, I've merged Jappy in https://github.com/llaske/sugarizer/commit/818ae02d7ba8a0a1cbbb358335291336f9c5e89e

Thanks to add this nice app to Sugarizer.

PR #125, opened 2017-07-02
Serve static files with Express to simplify.

icarito:
Make development simpler by removing dependency on Apache server.
Reverse proxy fun can still be had but should be left as excercise to
the systems administrator.

This was useful for me to test the Sugarizer server.

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,7 +65,7 @@ If you run Sugarizer this way often, you should create an alias for this command
 
 # Server
 
-Sugarizer Server is the back-end for network features of Sugarizer. It means: allow deployment of Sugarizer on a local server, for example on a school server, so expose locally Web Application (without Internet access). Sugarizer Server can also be used to provide collaboration features for Application and Web Application on the network. Sugarizer Server could be deployed on any computer with Apache2, Node.js and MongoDB, or in a Docker container.
+Sugarizer Server is the back-end for network features of Sugarizer. It means: allow deployment of Sugarizer on a local server, for example on a school server, so expose locally Web Application (without Internet access). Sugarizer Server can also be used to provide collaboration features for Application and Web Application on the network. Sugarizer Server could be deployed on any computer with Node.js and MongoDB, or in a Docker container.
 
 Sugarizer Server features include:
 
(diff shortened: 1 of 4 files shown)


icarito:
To test, try the updated instructions in the wiki. Also, now root is not required :-)

llaske:
It's a place where things are changing. Initially Apache2 was a prerequisite. I'm still using it on my local dev env. However the "production" env (hosting both http://try.sugarizer.org and http://dev.sugarizer.org) is deployed using nginx - so without Apache2. Here's the nginx setup file for this server:

```
# prod server
server {
        listen   80;

        server_name localhost try.sugarizer.org server.sugarizer.org;

        root /var/www/master;
        index index.html index.htm;

        location / {
                try_files $uri $uri/ /index.html;
        }

        location /api {
                proxy_pass http://localhost:8080;
                rewrite ^/api/(.*)$ /$1 break;
        }
}

# dev server
server {
        listen 80;
        server_name dev.sugarizer.org;

        root /var/www/dev;
        index index.html index.htm;

        location / {
                try_files $uri $uri/ /index.html;
        }

        location /api {
                proxy_pass http://localhost:8181;
                rewrite ^/api/(.*)$ /$1 break;
        }
}
```

With nginx, no need to use Apache2 redirection and .htaccess file. Redirection on the node.js port is done by the "proxy_pass" command.
I think this solution is better than using node.js express static route like you're proposing because node.js is not known for its ability to server efficiently static files.
Note also that the Sugarizer server part is currently in heavy rewriting by https://github.com/tarunsinghal92 for the GSoC and will be separated from the Sugarizer client (in https://github.com/llaske/sugarizer-server) in the next version.

icarito:
I understand about nginx, I saw it in the docker-compose file. This is fine. All I'm saying is that a simpler solution for serving static files is required for developers not to have to set it up in complex ways. And also, nginx will happily work as you describe, if a static file is found, never hitting the `express` httpd, so this could go _in addition to_ it.

No problem if you don't take this patch, just consider it a comment on developer usability. I want to add collaboration to Jappy on Sugarizer. Do you know if there's a way to make this happen also in Sugar?

Regards,
Sebastian

llaske:
Your idea is definitively good. I've finally worked with @tarunsinghal92 to update the next version of Sugarizer Server (https://github.com/llaske/sugarizer-server). So, this version will include Client exposition, the API and the new server Dashboard into a single node.js process. No need to use Apache2 or NGinx, the single command:

`node sugarizer.js`

Will do everything.
BTW I'm closing this PR because due to the future split between Client and Server it will be done in the server repo instead. Thanks for your nice suggestion.

llaske:
Hi Sebastian,

 I want to add collaboration to Jappy on Sugarizer. Do you know if there's
> a way to make this happen also in Sugar?
>
> Unfortunately collaboration in Sugar and in Sugarizer don't rely on the
same technology.
It's Gnome Telepathy for Sugar and Web Socket for Sugarizer. And there is
not compatibility or bridge between it.

Regards.

        Lionel.

icarito:
Thanks for the info!

What is the simplest example of collaboration enabled in Sugarizer activity?

Chat prototype? Paint?

Haven't found docs on the subject.

Any pointers welcome.

Regards,

Sebastian

llaske:
Hi Sebastian,

Yes, I'm sorry there is a lack of documentation here. We should have a
tutorial on that.

ChatPrototype is a good start because it's the activity written by Suraj
the GSoC student that wrote the presence API/server.
You could also have a look on the SharedNotes activity where presence is
handle in very few lines of code, starting here [1].
Note that these activities use a slightly updated
sugar-web/activity/activity.js file to handle launching from the network.
Get this file too so it will be easier to handle presence.

To explain in few words the way of integrating presence:
- First get the presence object using : activity.getPresenceObject() -
method exposed by the "new" activity.js
- Then call presence.createSharedActivity to create a new share activity
visible on the neighborhood view
- Watch events onDataReceived (someone sendMessage on the share),
onSharedActivityUserChanged (an user enter/leave the share) and
onConnectionClosed
- Send data to the share using sendMessage
- Check window.top.sugar.environment.sharedId variable to see if the
activity has been launched from the neighborhood view with a "Join".

Do not hesitate to ask me questions if you need more informations.

Best regards from France.

             Lionel.


[1]
https://github.com/llaske/sugarizer/blob/dev/activities/SharedNotes.activity/js/activity.js#L437

2017-07-22 20:30 GMT+02:00 Sebastian Silva <notifications@github.com>:

PR #126, opened 2017-09-06
A few small CSS changes to the search bar

eohomegrownapps:
While using Sugarizer I couldn't help but notice that the icons for the magnifying glass and cancel button were slightly out of alignment. I edited the CSS to increase the margin to the left of the magnifying glass from 5px to 6px, and to vertically center the magnifying glass, input text and cancel button in the search bar. This was done by removing the absolute top/left positions and replacing it with a percentage-based `top: 50%;`. However, this centers the top of the element meaning that the middle of the element is off-centre; to compensate for this I added `transform: translateY(-50%);` which would translate the element up by 50% of its height thus centering it vertically.

### DIFF
--- a/css/styles.css
+++ b/css/styles.css
@@ -103,11 +103,12 @@ input {
 	background-repeat: no-repeat;
 	background-position: center;
 	background-size: 20px 20px;
-	top: 5px;
-	left: 2px;
+	left: 6px;
 	width: 20px;
 	height: 20px;
 	display: inline-block;
+	top: 50%;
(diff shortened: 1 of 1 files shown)


llaske:
You're right it was not centered.
Thanks. Merged in dev branch on https://github.com/llaske/sugarizer/commit/bbd42e2a220f91189da307260f471d2b58258e84
Please send your PR on the dev branch, master branch is now reserved to stable/release.

PR #134, opened 2017-12-03
Added resize feature

matiasmartineeez:


llaske:
Nice start. Two issues:
- The current mode (pen, stamps, ...) stay selected. I guess you have to implement a new mode. See switchMode method.
- It will be nice to start rendering the image or a least a dotted rectangle to show to the user that he could drag the image.

llaske:
Very nice.  I've noted two issues now:
- The image is doubled when a second image is inserted
- There is a small pen trace under the image at the place where the image is inserted - look like there is still an issue on the selected mode
Good work BTW.

llaske:
Seems to work but there is lot of error message in the browser console (see below)

insertimage-button.js:43 Uncaught TypeError: ctx.beginPath is not a function
    at insertimage-button.js:43
(anonymous) @ insertimage-button.js:43
insertimage-button.js:56 Uncaught TypeError: ctx.putImageData is not a function
    at insertimage-button.js:56
(anonymous) @ insertimage-button.js:70
insertimage-button.js:70 Uncaught TypeError: ctx.putImageData is not a function
    at insertimage-button.js:70

Plus there is an issue when you're in stamp/text mode and you insert an image. Both text and stamp are displayed with the image.

llaske:
Fixed in https://github.com/llaske/sugarizer/commit/1cddc67f4702cf084cf03a0eaa79e187cc10d12a

PR #139, opened 2017-12-07
Added pomodoro activity

sanatankc:
I have created a pomodoro activity for sugarizer, it looks like this.
screen shot 2017-12-07 at 1 28 17 pm

Pomodoro is a productivity method. More about this: https://en.wikipedia.org/wiki/Pomodoro_Technique

What I have done.
- [x] created basic pomodro activity
- [x] An icon for pomodoro
- [x] Ability to create custom timers.


bzg:
Hi @sanatankc, this looks great, thanks a lot!

@llaske let me know if I can take the initiative to merge this kind of PRs.

sanatankc:
I have added controls for Pomodoro to change Work and Break Sessions.
Here's how it looks like after adding controls.

screen shot 2017-12-09 at 8 33 36 pm
screen shot 2017-12-09 at 8 34 15 pm
screen shot 2017-12-09 at 8 35 14 pm
screen shot 2017-12-09 at 8 35 21 pm

I hope it this pull is ready to be merged. 😊

llaske:
Very nice activity. I just wonder if it's pertinent for children between 6 to 12: the target for Sugar platform. Do you have experience on using this method for children ?

Two more technical remarks:
- Use a more Sugar compatible icon for the activity: should be adaptable to the user colors
- Could you integrate localization using the included webL10n.js library

sanatankc:
I don't have experience of using this method on children. It is the productivity method. I have read some articles which suggests it helps kid by giving them reward or something during break.

1. I will look into icon issue.
2. I honestly don't much about localization but I will try to learn about it and implement it.

Thanks!

llaske:
The icon doesn't match prerequisite of Sugar icons to allow an auto colorize with user colors. See [here](https://wiki.sugarlabs.org/go/Development_Team/Almanac/Making_Icons) for more or look how other activities icon are made. May be you could increase size of the tomato border.

Regarding localization we're using [webL10n.js](https://github.com/fabi1cazenave/webL10n) library. Look for example Moon activities to see how you could use it.

One question: not sure to understand how Work and Break work: you've got X minutes on work then Y minutes of break?

sanatankc:
Yes Break and Work works precisely like that. You've got x minutes to work and then you got y minutes of break in which you can relax. and after that again work and so on.

llaske:
But why it's not possible to have more than 1 minute of break?
May be it could be nice to use user colors into the activity: why not for example use the foreground for the work and the background for the break. What do you think ?
Like I said for the Game of Life, it should be interesting to use the datastore to save the context of the activity so the user could find counters in the same state.
May be adding a toolbar button to reinit could be nice too.

sanatankc:
> But why it's not possible to have more than 1 minute of break?
Confirmed, it was a bug.
> May be it could be nice to use user colors into the activity: why not for example use the foreground for the work and the background for the break. What do you think ?
I was thinking about that, and I implemented it.
> May be adding a toolbar button to reinit could be nice too.
Done.
> Like I said for the Game of Life, it should be interesting to use the datastore to save the context of the activity so the user could find counters in the same state.
For some reason every time datastore get gives me null as data, even though I can clearly see it in chrome devtools. Setting and Getting with native localstorage worked just fine, because I thought it might not work in sugarOS.

llaske:
Hmmm, strange. You right, your record seems correct in the datastore. Try to see what happens during datastore.loadText().

BTW it show that something is wrong in your source code.
I'm sorry to say that you should not use the ES6 syntax "=>" in your code. Unfortunately, all browsers do not support this syntax, specifically on iOS. So, I must ban this syntax from Sugarizer to ensure a maximum of portability. Could you make the change?

llaske:
I've tested more. data is not `null` on my side so I've got an issue:

```
activity.js:44 Uncaught ReferenceError: stateFromLocal is not defined
    at datastore.loadAsText (activity.js:44)
    at DatastoreObject.loadAsText (sugarizer.js:131)
    at main (activity.js:41)
    at window.addEventListener (activity.js:11)
    at fireL10nReadyEvent (webL10n.js:114)
    at onResourceLoaded (webL10n.js:319)
    at webL10n.js:280
    at XMLHttpRequest.xhr.onreadystatechange (webL10n.js:131)
```

`stateFromLocal` is not declared in the context. I guess it was something that you've not finished yet.
But I don't understand why data is null in your context.

sanatankc:
Didn't know that. I will use Babel to compile all files in Es5 because I am using many ES6 features.

llaske:
Cool it's nice to use the user color.
BTW still an issue, due to datastore: all values are "undefined" after datastore load.
Guess it's the same issue than in your other activity

image

llaske:
Still few issues:
- Reload of a saved activity doesn't work. It's related to the with the JSON.parse() in the loadText method. If you don't stringify into the saveText, you don't need to parse
- When the counter is reload an the play button is push, the counter restart from the beginning instead of continue 
- When the counter is paused and the reinit button is clicked, nothing happen

sanatankc:
I have fixed almost all visible errors, and dataStore is working fine now.

llaske:
Very good work. I'm going to test it on mobile platforms and merge it.
Thanks.

llaske:
Can't merge it like this. I've tested it on Chrome OS (i.e. Chrome Extension) and unfortunately it don't work.
I've figured out what's going wrong: "window.setInterval" is not supported in Chrome Extension. I've tried to replace it by "chrome.alarms" like this:

```
			if (typeof chrome != 'undefined' && chrome.app && chrome.app.runtime) {
				chrome.alarms.create("pomodoro", {periodInMinutes:(this.refreshRateMS/1000.0)});
				chrome.alarms.onAlarm.addListener(function(){self._updateTime(self)});
			} else {
				this.tickTimer = setInterval(function(){self._updateTime()}, this.refreshRateMS)
			}
```

And:

```
				if (typeof chrome != 'undefined' && chrome.app && chrome.app.runtime) {
					chrome.alarms.clearAll();
				} else {
					clearInterval(this.tickTimer)
				}
```

But ut don't works well: play button is not updated to pause and chrome.alarms period are limited to 1 by minute so it's not a viable solution.
The only solution I see is to rethink your drawing using requestAnimationFrame like in StopWatch activity because it's fully supported in Chrome Extension. See here: https://github.com/llaske/sugarizer/blob/dev/activities/Stopwatch.activity/js/activity.js#L9

Could you fix that?

sanatankc:
I can't run it as chrome extension for some reason in Chrome 63.

 It never finishes unpacking the extension.

llaske:
Hmmm. Strange. In that case try "Load unpacked...". Just use the Sugarizer directory as content of unpacked extension.

llaske:
Merged in https://github.com/llaske/sugarizer/commit/0394f80e7ad1ae06df4894f06bf9deb9c050cd01

PR #147, opened 2017-12-12
Added Game Of Life Activity

sanatankc:
I have created a Conway's Game of life Activity.

Game of Life is 2d grid cellular automaton. It just have these four rules.

1. Any live cell with fewer than two live neighbours dies, as if caused by underpopulation.
2. Any live cell with two or three live neighbours lives on to the next generation.
3. Any live cell with more than three live neighbours dies, as if by overpopulation.
4. Any dead cell with exactly three live neighbours becomes a live cell, as if by reproduction.

Just using these  simple looking rules cells make amazingly complex patterns.
Here is one of those moments:
ezgif com-video-to-gif 4

Amazing video about Game of life by it's creator: https://www.youtube.com/watch?v=E8kUJL04ELA
More about Conway's game of life: https://en.wikipedia.org/wiki/Conway%27s_Game_of_Life

### Why?
1. I think this will be helpful for teaching kids about life.
2. Even with just these 4 simple rules cells makes amazingly complex patterns. It will be helpful to understand how complexity arises from simplicity. (Just like how complexity arises in an ant colony.)
3. It is fun. 😄 

TODO
- [ ] Localisation support (I tried to figure it out by myself but couldn't, any help will be appreciated).


llaske:
Nice idea and cool works.
Few remarks:
- The icon is not compatible with Sugar icons (see my message on your other PR)
- Would be great if grid colors match user colors (see for example Reflection or Abacus activity to see an example)
- The window is not really responsive, should be better if grid size fit window size

sanatankc:
I have worked on all your suggestion and made changes on both pull except localisation, I am getting an error `#Generation missing for [en-GB]`, not sure why.

llaske:
Cool. I do like your activity. 
Still two remarks:
- Regarding the localization error, you have to include "locale.ini" in the HTML file like [here](https://github.com/llaske/sugarizer/blob/master/activities/Moon.activity/index.html#L13)
- It could be nice when you stop the activity to store the current board state (and generation number). So when the user relaunch the activity, it could retrieve the board in the same state. For example to see its creation to friends. See for example [here](https://github.com/llaske/sugarizer/blob/master/activities/Moon.activity/js/activity.js#L17) how to store and [here](https://github.com/llaske/sugarizer/blob/master/activities/Moon.activity/js/activity.js#L10) how to load.

sanatankc:
Thanks for remarks!

I have fixed localisation error and also implemented datastore.

activities/GameOfLife.activity/js/activity.js
```diff
@@ -0,0 +1,191 @@
...
+
+    shouldPlay: ['.fake-selector', function (fakeElem, value, prevValue) {
+      if (value) {
+        generateGeneration();
+      }
+    }]
+  });
+
+  dataStore.loadAsText(function (err, metadata, data) {
+    var dataToLoad = data ? data.state : randomPattern();
```
llaske: it seems that state is in data.state.state so need to be replaced by:
`    var dataToLoad = data && data.state ? data.state.state : randomPattern();`

sanatankc:
You are right. I'll make these changes

sanatankc:
I have updated as requested changes. I am still not able to test these locally (I tried running file from a local python server with no luck). I just can hope that my changes are correct.

llaske:
Sorry, there is still an issue, see error below.
You should ensure first that you could test it locally else it will be impossible for your to fix the issue.

Board.js:32 Uncaught TypeError: Cannot read property 'forEach' of undefined
    at Board.update (Board.js:32)
    at Array.state.subscribe.boardState (activity.js:111)
    at vanilla-state.js:22
    at Array.forEach (<anonymous>)
    at vanilla-state.js:19
    at Array.forEach (<anonymous>)
    at render (vanilla-state.js:18)
    at State.set (vanilla-state.js:40)
    at activity.js:131
    at DatastoreObject.loadAsText (sugarizer.js:131)

sanatankc:
I don't know it just not working for me. How you are running these files?

activities/GameOfLife.activity/js/activity.js
```diff
@@ -32,7 +32,7 @@ function main(Board, State, patterns, color, shadeColor, l10n, dataStore) {
 
   var storeLocally = function storeLocally(state) {
     dataStore.setDataAsText({
-      state: state
+      state: state,
```
llaske: I guess you planned to add generation here ?

sanatankc: Yes I'll make this fix.

sanatankc:
updated.

sanatankc:
I have fixed datastore bug. Just sharing the process I found the bug.

As you said that I should ensure first that I could test it locally else it will be impossible for me to fix the issue.
screen shot 2017-12-16 at 7 25 56 am
So , when I opened my local storage in chrome dev tools and saw this.

It didn't look right as these random values were always changing on each load and create a new local storage. So, i started digging into source code of library.

I first traced back from where it is importing datastore.
`sugar-web/activity/activity.js`, and found this on `Line 46`
`datastoreObject = new datastore.DatastoreObject();`

`sugar-web/datastore/sugarizer.js`
Again I traced back it to it's original source code and found this.  and found out it was taking an argument.
```javascript
    function DatastoreObject(objectId) {
       ....
   }
```
it was taking an objectid as an argument, which later it was using as a hash value to store in local storage. So, I passed an argument in `sugar-web/activity/activity.js`. Like this:
```javascript
datastoreObject = new datastore.DatastoreObject('game_of_life');
```

Which Worked. It hashed my local storage with `game_of_life`. And after fixing errors whole game is working.

## Side Note
Then I thought if it was working for you (not printing null).  Then there might be some other bug.
So I decided to dig into it. In the same file `sugar-web/datastore/sugarizer.js`. I found this.

```javascript
    function DatastoreObject(objectId) {
       ....
       ....
       ....
       if (this.objectId === undefined) {
            var env_objectId = window.top.sugar.environment.objectId;
           ...
        }
    }
```

Which means in case of there is no arguments, it was referencing to some objectId property attached to window. So I searched my whole folder structure for `window.top.sugar.environment.objectId`, and found it in `env.js`.  and there was some code like this.

```javascript
var getUrlParameter = function(name) {
    var match = RegExp('[?&]' + name + '=([^&]*)').exec(window.location.search);
    return match && decodeURIComponent(match[1].replace(/\+/g, ' '));
};
window.top.sugar.environment = {
	activityId: getUrlParameter("aid"),
        activityName: getUrlParameter("n"),
	bundleId: getUrlParameter("a"),
	objectId: getUrlParameter("o"),
	sharedId: getUrlParameter("s")
};
```
 So I put up a console for objectId and it was `null`. As I understand this code this just extracts query Parameter from URL. So I looked at my URL  it was like this.
```
file:///Users/sanatan/Desktop/code/sugarizer/activities/GameOfLife.activity/index.html?aid=00a9107b-3137-49a4-8025-b3b01cb0e3aa&a=org.sugarlabs.gameOfLife&n=Game%20Of%20Life
```

There was no parameter called 'o' (and also 's') in URL. I don't know ehy but the reason why it worked on your computer and not mine is there was no parameter called 'o' in my URL and it is in yours (Can you confirm this?)

llaske:
Very good work of analysis of the way of working of datastore.
The 'o' parameter don't exist when you run the activity for the first time, when your run with menu item "Start new" or when you run it from the activity list view. Shortly it means: create a new instance for this activity.
When you reopen the activity from the Home view or from the Journal the 'o' parameter is set with the id of the instance.
This feature is implemented [here](https://github.com/llaske/sugarizer/blob/master/lib/settings.js#L328). It's called [here](https://github.com/llaske/sugarizer/blob/master/js/homeview.js#L474) from the home view, see function `runActivity`, `runNewActivity`, `runOldActivity`.
I guess your 'o' parameter is always null because you create a new instance each time: may be you always launch activities from the list view ?
BTW fixing the DataStore id as you've done [here](https://github.com/llaske/sugarizer/pull/147/commits/009f6a4322f40c9e4043fccd11149487470d7014#diff-73f14dfc6d73fab07e06a4ddb8438580R46) is not at all a good idea. First because it's never a good idea to patch an internal library, second because it works not in a good way: it's just said that there is only on instance of the activity. So each time you will open the activity (whatever you come from) it will be open with the last context.
Just remove this patch and it will works as expected.

activities/GameOfLife.activity/lib/sugar-web/activity/activity.js
```diff
@@ -43,7 +43,7 @@ define(["webL10n",
         }
         bus.onNotification("activity.stop", sendStopEvent);
 
-        datastoreObject = new datastore.DatastoreObject();
+        datastoreObject = new datastore.DatastoreObject('game_of_life');
```
llaske: Not at all a good idea. It forbid to have multiple instances of your activity.

sanatankc: OK

sanatankc:
I have removed the patch

llaske:
Cool it perfectly works now. Congrats.
I need to do some more testing on Android and iOS and I will merge it.

Could you confirm that your 'o' parameter was null because you created a new instance each time ?

sanatankc:
It was null because I was launching it from List View.

llaske:
Merged in https://github.com/llaske/sugarizer/commit/81e496f09dfbaffec3a89ff6a7a4ad26873c9ddf

PR #153, opened 2017-12-13
Added notification

Youngestdev:
Notification for export to png button.


Hrishi1999:
Not sure why would you add humane.js if it is already available at /lib/humane.js

llaske:
- Please put the humane.js into lib directory instead of js directory
- You should implement localization for the message using [webL10n.js](https://github.com/fabi1cazenave/webL10n) library. See for example how it's used in Moon activity

Youngestdev:
ow. i didnt  notice humane.js was available in lib

PR #157, opened 2017-12-19
Add notify to user!

ccr4b:


llaske:
The message should be localize using webL10n.js library.

PR #167, opened 2018-01-03
Added Journal chooser

kiridix:


llaske:
Same than https://github.com/llaske/sugarizer/pull/168

PR #168, opened 2018-01-03
Handle images in Memorize activity

kiridix:


llaske:
Nice job. 
Two remarks:
- The Journal icon and the Favorite icon are missing in the Journal box. See the browser console where your could see the message. You have to copy these files in the right directory (see the Paint activity)

image

- I'm not a big fan of your "Image" button. May be you could use the same icon than in the Sugar Memorize: https://github.com/sugarlabs/memorize-activity/blob/master/icons/import_picture.svg

Issue #146, opened 2017-12-11
Allow clicking on Abacus balls to move it


satyamchaudharydev:
For now, we have to drag balls which is requires bit effort and is error-prone.

It would be great if we can move balls by just clicking on them and it would also be better for mobile and touch users.

ag1709:
I am interested in working on this. Is anyone already working on it?

llaske:
Cool. Let's go.

ag1709:
Which file in particular is to be modified?

llaske:
Guess it should be in https://github.com/llaske/sugarizer/blob/master/activities/Abacus.activity/js/abacusbead.js#L88
Try to debug to understand how it works.

utkarsh-raj:
https://github.com/llaske/sugarizer/pull/181#issuecomment-370563571

The changes proposed by @llaske are intuitive, more than the current behaviour. I would carry the work forward by @aditigupta17 to incorporate the proposed behaviour. Any ideas for still better behaviour ? (Since the issue has been open for quite a long time)

llaske:
Like I said, we could propose an automatic move on click/tap when there is no doubt about the direction expected and let the drag gesture - since double-tap is not intuitive - when it's not possible to do the deduction. It should be tested.

utkarsh-raj:
https://github.com/llaske/sugarizer/pull/277#issue-255474687
I have opened a PR with some changes. Can you look and see if it feels intuitive @llaske ?

llaske:
Fixed by https://github.com/llaske/sugarizer/issues/146

llaske:
Closed in https://github.com/llaske/sugarizer/commit/dd3c58d56981e4dfa6d94601b95d7117a3bc9241

PR #181, opened 2018-01-15
Enabled clicking on abacus beads to move it #146

aditigupta17:
Earlier, the user had to drag the beads, now a single click sends the beads up and double-clicking on them brings them down.
peek 2018-01-15 20-54

### DIFF
--- a/activities/Abacus.activity/js/abacusbead.js
+++ b/activities/Abacus.activity/js/abacusbead.js
@@ -84,24 +84,16 @@ function AbacusBead(x,y,blockcol,abacus,column,i,value,beadheight){
 			this.resetAge();
 		}
 	}
-
+	
 	this.addClickListeners = function(){
 		var th = this;
 		var col = column;
-		this.bead.on("mousedown", function(evt) {
-		    th.originaly = evt.stageY;
+		this.bead.addEventListener("click", function(evt){
+			col.shuntLeft(th.index);
 		});
-
-		this.bead.on("pressup", function(evt) {
-		    if (th.originaly<evt.stageY){
-		    	//moved down
-		    	console.log("down");
-		    	col.shuntRight(th.index);
-		    } else if (th.originaly>evt.stageY){
-		    	console.log("up");
-		    	col.shuntLeft(th.index);
-		    }
-		});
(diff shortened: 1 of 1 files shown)


llaske:
Nice work but using double-click is a problem because it will not work on touch screen.
Another idea ?

aditigupta17:
Oh yeah, didn't see that..any suggestions from your side?

aditigupta17:
I just tried, it works well on touch screen as well, just double-tapping the bead is causing it to move down.

llaske:
Hmmm. Not sure double-tap is a intuitive gesture on touch screens.
Is there a way to compute automatically (by deduction) the direction expected by an user ? A sort of AI that could know that when you click it's for move up or down.
If it's possible to do that without any error, let's try this. Else we could propose an automatic move on click/tap when there is no doubt about the direction expected and let the drag gesture when it's not possible to do the deduction.
What do you think?

aditigupta17:
Ya, the latter sounds good. Working on it.

ritjoe:
Is there any other behavior associated with a single click or a touch (long or short) on a bead beyond moving it?

aditigupta17:
According to the changes made in the PR, a single click/tap should send the bead up and double click/double tap moves it down.

ritjoe:
On Sun, Mar 04, 2018 at 03:01:11AM -0800, Aditi Gupta wrote:
> According to the changes made in the PR, a single click/tap should send the
> bead up and double click/double tap moves it down.

OK. I'm just wondering why not make a single click do both?

Or rather, the only option in either direction is "move to the other
position."

So why use one for one direction and the other for the other direction?

aditigupta17:
peek 2018-03-05 08-52
There are a few types of abacii, (like the one shown in the gif) in which beads can move in either direction. So to differentiate that, different gestures are needed.

llaske:
Thanks all for your analysis. So my first intuition is the right: we could propose an automatic move on click/tap when there is no doubt about the direction expected and let the drag gesture - since double-tap is not intuitive - when it's not possible to do the deduction.

llaske:
Fixed by https://github.com/llaske/sugarizer/pull/277

PR #189, opened 2018-02-24
Added functionality to calculator activity

hardiked:
I have added modulo functionality in the calculator activity and also modified the pallet that we see beside the calculator screen. I found that these are just piled up beside our calculator and there is no way to get rid of this. So, I have added bin button on top of every pallet graph/result/error. so, that user can remove the unnecessary pallet and thus able to clean up space simultaneously can organize his/her work better.


llaske:
Do not include in the same PR a mix of different features/fixes.
Here's my comments on your work:
- Not a big fan of your result palette. It's better to leave the result always on display.
- Issue on Speak activity name is already fixed here: https://github.com/llaske/sugarizer/pull/184
- Okay for the Blockrain button title

hardiked:
How about keeping track of users score and accuracy in Abecedarium activity.

PR #192, opened 2018-02-25
score and accuracy in Abecederium activity

hardiked:
This activity contains lots of things that are helpful to learn many things. I have enhanced this activity by adding score and accuracy popup so that user may improve his functionality from next time.


llaske:
Thanks for this contribution.
Few remarks:
- Neither the score or popup are responsive. Because Sugarizer should adapt to any screen size, all content must be responsive.
- If you add some text, it should be localized to adapt to the current Sugarizer language (see https://github.com/llaske/sugarizer/blob/dev/docs/tutorial_step5.md for more).
- I'm not sure it's a good idea to display the score during the game. It's better to display it only at end.
- Accuracy should not be displayed with 8 decimal of precision (66.66666666666667% in my test).
- You must not include fix for multiple activities in the same PR (here Abecedarium and Speak are mixed).

hardiked:
Due to some problems, I have to fork the repo again by deleting the old one. so, I have to make pull request again and I have made all the changes prescribed above in PR #193 .

PR #193, opened 2018-03-01
Changes made as specified in #192

hardiked:
changes made as specified in #192 
<ol>
      <li>Score and popup should be responsive</li>
      <div>Both score and popup are now responsive.</div>
      <li>Text must be localized to adapt the current language</li>
      <div>Language support added</div>
      <li>Score during the game</li>
      <div>I have removed the score from the game and it can be seen only at the end of the game.</div>
      <li>Precision in accuracy</li>
      <div>It is neccessary to add the precision as such a big number is not at all good to see for user thus two decimal point precision added.</div>
</ol>


llaske:
- There is still issues on responsive. See for example this screen capture (popup not centered, scrollbar on the left of screen).

image

- The language support is hard coded, should use the webl10n library instead as mentioned earlier.

- Please clean commented lines and trace log in the source code.

- You must not include a fix for Speak activity into the same PR than a fix of Abecedarium if both are unrelated.

hardiked:
webl10n library is used and language is no longer hardcoded.
all commented lines and trace logs are removed.

hardiked:
score popup is now much more responsive.

llaske:
No more pertinent

PR #195, opened 2018-03-07
Sprint Math activity

manki11:
This is my first activity. I have made the UI of the game responsive, i am facing one issue the button in the top-right is not closing the activity. @llaske can you help me with that.

Screenshots of activity:
Desktop:
screen shot 2018-03-07 at 6 51 39 pm

Mobile:
screen shot 2018-03-07 at 6 51 54 pm


llaske:
This symptom often mean that you don't call the "activity.setup()" function at startup.
See [here](https://github.com/llaske/sugarizer/blob/dev/docs/tutorial_step2.md#customize-content) for an explanation. I suggest you to do the whole tutorial.

manki11:
I was implementing localisation from the tutorial, and getting some unexpected errors

```
env.getEnvironment(function (err, environment) {
            currentenv = environment;

            console.log("Language code in env="+language);
});

window.addEventListener("localized", function () {
            // Set current language to Sugarizer
            console.log("Language code in window.addeventlistener="+webL10n.language.code);
            console.log(webL10n.get("Time"));

            document.getElementById("timeString").innerHTML = webL10n.get("Time");
            document.getElementById("scoreString").innerHTML = webL10n.get("Score");
            document.getElementById("instructionString").innerHTML = webL10n.get("Instruction");
            if (play) {
                document.getElementById("gameString").innerHTML = webL10n.get("Reset");
            } else {
                document.getElementById("gameString").innerHTML = webL10n.get("Start");
            }

        });
```

Image of console:
screen shot 2018-03-08 at 6 39 25 pm


@llaske as you can see from the console.log, the language code is changing between getenv function and localized function. Also webL10n.get() function is not returning the strings.

My Language.ini file
```
[es]
Time= Vez
Score= Puntuación
Instruction= Haga clic en la respuesta correcta
Start= Empezar Juego
Reset= reiniciar el Juego
GameOver= Juego terminado, tu puntaje es
```

llaske:
Seems not related to localization. You're using "getElementById()" on HTML elements that don't exist in your HTML file: I don't see any HTML element in your index.html file with id "timeString", "scoreString", ...

manki11:
Actually the new changes haven't been committed i have been able to solve the issue of retrieving strings. However When i am using a language such as french or spanish, it keeps defaulting back to english. In this commit localization doesnt work half of the time.

manki11:
@llaske made the activity consistent with user colors, i implemented a setColor() function in activity.js.
I had a doubt, when you said level complexity, do you mean to implement levels like easy, medium, hard?
The language issue was solved in an earlier commit.

llaske:
It's better.
I think you could leave the background white and use the second colors as foreground for the operation "8+4" (look how the Game of Life activity is). Do not color the Time and Score box, so number to guess will be more visible.
The start/reinit button should be an icon in the toolbar.
Regarding text: less is better because users of the activity could be not able to read. Avoid text if possible so I think "Click on the correct answer" is not useful. Here I think it's obvious. The box with the ending score is not necessary neither.

A small bug: "Correct" box appears at start up even thought no choice has been done.

I think something goes wrong with your CSS. There is some "glitch" around toolbar buttons and activity palette content has wrong size (see below). Did you use the "ActivityTemplate" CSS files for the activity?

image

Yes I mean: easy, medium, hard.

llaske:
Should be nice also to use the journal to store the score/current context.
So when you reopen the activity you could retrieve your last score.

llaske:
Another advice: Reduce the size of the content in your icon. Icons should have a border around the content. See for example icons for Memory, Record or Game of Life activities.

manki11:
@llaske i don't know what you mean by "launch the game immediately" as the game launches as soon as the activity starts and there is no need to click on the restart button.

llaske:
One more thing: think to localize the "xxx joined/left" message (thought it's not localized in the sample I've pointed to you :-)

manki11:
@llaske the reason why i disabled restart is because the questions are the same, so a user can take the quiz and just before his time ends, restart giving him an unfair advantage. Also isn't it a good idea to repeat the history, as if he repeats it, others should know it was that users second time?

llaske:
Okay I understand. But the idea is, at the end of the game, to let the user to relaunch a game with another difficulty/operations. Today it's just not possible. Restart in that case should regenerate a new set of questions and sent it to connected users to restart the game for connected users too.

PR #210, opened 2018-04-22
Brazilian Portuguese translation

afmachado:
Brazilian Portuguese translation pt_BR


llaske:
Thanks a lot for this translation. BTW the FoodChain localization concern both strings and sounds. So until, the sound directory (and audio files inside) is not translated, it's not possible to integrate your localization.
Note also than a dedicated website will allow you to more easily translate Sugarizer core and activities. Look on http://translate.sugarizer.org/languages/pt/sugarizer/

llaske:
Integrated in dev branch starting at
https://github.com/llaske/sugarizer/commit/04b0d0c55d5a35765550c32ad697f403b63ce8fa

PR #211, opened 2018-04-25
Brazilian Portuguese translation audio files

afmachado:
Brazilian Portuguese translation audio files MP3 and OGG


llaske:
Thanks a lot. Sounds are very good.
BTW there is two missing sounds:

- mice
- skunk

Plus one MP3 file is missing (only OGG is present):

- mosquito.mp3

I've started to integrate your change into the branch https://github.com/llaske/sugarizer/tree/pr/211
You could test to see the result.

llaske:
Integrated in https://github.com/llaske/sugarizer/commit/aca76fe178a9357a683aa100b2d47cfdd2e94185

PR #212, opened 2018-04-26
Brazillian Portuguese translation PO

afmachado:
Brazillian Portuguese translation PO


llaske:
Integrated in https://github.com/llaske/sugarizer/commit/33fc8355e03be00e7c27d3f564e693385474adea
BTW po files are generated from .ini file, so you don't need to update po files.

PR #219, opened 2018-07-21
Lunar ephemerides phase for 2018 to 2024

quozl:
Moon will cease to function properly after 2018, because the data set ends.  This patch extends the data set, and removes old data.


llaske:
Thanks a lot @quozl

Issue #204, opened 2018-04-18
Save Clock activity settings in Journal


llaske:
Clock activity settings (clock format and display type) is not saved in the Journal.
It could be nice to save it into the journal so the display could be the same next time the activity will be open.

llaske:
Clock activity source code is here: https://github.com/llaske/sugarizer/tree/dev/activities/Clock.activity
Adding datastore handling to an existing activity is explain here: https://github.com/llaske/sugarizer/blob/dev/docs/tutorial_step4.md

llaske:
Fixed in https://github.com/llaske/sugarizer/commit/9db543cc7264a6d3033b0bb1e32badfcd0283115

PR #228, opened 2018-11-06
Implemented journal to stopwatch

### LINKED ISSUE #201: Save Stopwatch activity context in Journal
The Stopwatch activity don't use Journal to store the activity context.
It could be nice to store values for current counters and intermediate times in the Journal so when the activity is reopened, all values will be retrieve.

image

FreddieN:
This PR resolves #201. It allows users to save their stopwatch data in the Sugarizer journal to be viewed/edited later.

demo

Let me know if any issues :)


llaske:
Very good work. Thanks.
Fixed https://github.com/llaske/sugarizer/issues/204

PR #229, opened 2018-11-07
Fixed #213 Fototoon stop button icon

### LINKED ISSUE #213: Fototoon close icon missing in small screens
It seems that the close icon is missing and does not works in Android.

https://imgur.com/a/fapv550

Thanks/merci!

AndreaGon:
Added CSS media query (see /*For smartphone mode*/ comment for the code).


llaske:
This idea is not to reduce toolbar/button size but to hide other button that could hide Stop button.
See how it works when reducing the window size for example in Paint activity.

Please send PR only on dev branch.

PR #231, opened 2018-11-07
sanitized chat input to prevent XSS

### LINKED ISSUE #225: ChatPrototype activity does not sanitize input
The ChatPrototype activity is susceptible to arbitrary script injection, and by extension, XSS.
Animated GIF of demonstration

To fix this, either sanitize all HTML tags, or select for certain "whitelisted" tags.

FreddieN:
Fixes #225
This PR fixes the issue where a user could enter HTML into the message field and send it to another client for it to be executed.


llaske:
Good job. Thanks.

PR #232, opened 2018-11-08
Aligned Buttons Better in Help Popup

### LINKED ISSUE #135: Fix: Align buttons better
This is how (prev | next | end) buttons looks like now, which is not ideal in any way. Doesn't looks good by any UX perspective.
screen shot 2017-12-04 at 1 56 18 pm

It would look better if it looked something like this.

try sugarizer org_

mchlp:
- Use Flexbox to align buttons
Updated layout:
image


llaske:
Nice work.

PR #233, opened 2018-11-08
Fix #213 Fototoon stop icon fix

### LINKED ISSUE #213: Fototoon close icon missing in small screens

AndreaGon:
Added CSS media query. Also hid splitbar for smartphone mode (on desktop mode it shows.)


llaske:
It solve the issue but it's too aggressive, you start reducing size and hiding items thought there is still enough room in the toolbar. You should only hide item when the size become too small.

llaske:
That's very good now. Thanks.

PR #234, opened 2018-11-11
Add number of play in Flip activity

### LINKED ISSUE #216: Add number of play in Flip activity
It could be nice to see the current number of play in Flip activity. 
The number of play should be count from the starting position. The lower this number is, the better the player is.
Could be displayed in the same way than "Generation count" in the Game of Life activity.

FreddieN:
Fixes #216
There is now a turn counter in the flip activity, similar to the generation count in the Game of Life activity. There is also a coloured skill indicator to let the player know how well they are doing. See below:

demo
demo

Let me know if any issues :)

### DIFF
--- a/activities/Flip.activity/css/activity.css
+++ b/activities/Flip.activity/css/activity.css
@@ -61,3 +61,9 @@ background-size: contain;
 border-radius: 5.5px;
 background-color: #C0C0C0;
 }
+#flip-count {
+  flex: 1;
+  text-align: center;
+  font-size: 72px;
+  font-weight: 800;
+}
--- a/activities/Flip.activity/css/activity.css
+++ b/activities/Flip.activity/css/activity.css
@@ -64,6 +64,7 @@ background-color: #C0C0C0;
 #flip-count {
   flex: 1;
   text-align: center;
-  font-size: 72px;
+  font-size: 8vh;
   font-weight: 800;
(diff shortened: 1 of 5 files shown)


llaske:
Nice start. Few remarks:
- The "Turn" string should be localized, you need to include localization (see https://github.com/llaske/sugarizer/blob/dev/docs/tutorial_step5.md)
- The "Turn" string is not responsive: its size should depend of the window size
- The counter is not stored in the datastore so it restart to 1 when the activity is relaunched from the Journal
- Good idea to change line color depending of number of turn. BTW it should be also dependent of board size: it's easier to flip all board if size is smaller

FreddieN:
Turn string is now localized:
demo
It is also now responsive to suit the screen height. The counter is also stored in the datastore so when opening from the journal the user's session is restored. I've adjusted the indicator to adjust to the board's difficulty.
Let me know if you want anything else changed.

llaske:
It's better.
BTW there is still two issues:
- The line color is green when the game is restored from the Journal, whatever number is,
- When reloaded from Journal the board size is changed (it's smaller) and not optimal.

FreddieN:
I think you might have meant this button:
demo
I have now updated the game to store this in the journal.

llaske:
It works now. Great job.

PR #235, opened 2018-11-12
Save Clock activity settings in Journal (fixes issue #204)

sdziuda:
Update the file to let it save settings in journal

### DIFF
--- a/activities/Clock.activity/js/activity.js
+++ b/activities/Clock.activity/js/activity.js
@@ -1,4 +1,4 @@
-define(["sugar-web/activity/activity","sugar-web/graphics/radiobuttonsgroup","mustache","moment-with-locales.min","webL10n"], function (activity,radioButtonsGroup,mustache,moment) {
+define(["sugar-web/activity/activity","sugar-web/env","sugar-web/graphics/radiobuttonsgroup","mustache","moment-with-locales.min","webL10n"], function (activity,env,radioButtonsGroup,mustache,moment) {
 
     // Manipulate the DOM only when it is ready.
     requirejs(['domReady!'], function (doc) {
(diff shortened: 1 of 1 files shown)


llaske:
Nice.
It works but buttons selection in the toolbar should reflect the current state and it's not the case when settings is read from the journal.
For example: launch the activity, set the date, close the activity, relaunch the activity, we could expect that date button should be selected and when you click on it you remove the date but it's not the case, you should click twice to unset the date.

activities/Clock.activity/js/activity.js
```diff
@@ -16,15 +16,19 @@ define(["sugar-web/activity/activity","sugar-web/env","sugar-web/graphics/radiob
             } else {
                 activity.getDatastoreObject().loadAsText(function(error, metadata, data) {
                     if (error==null && data!=null) {
-                        console.log(data);
-                        clock.changeFace(data[0]);
+                        face=data[0];
+                        if(data[0]=="simple"){
+                            simpleClockButton.onclick();
```
llaske: I'm not sure it's a good idea to do "onclick" it's easier to do the changeFace like you've done before, then just add the CSS class "active" to the toolbar element.  Something like: document.getElementById("simple-clock-button").classList.add("active");

activities/Clock.activity/js/activity.js
```diff
@@ -16,15 +16,19 @@ define(["sugar-web/activity/activity","sugar-web/env","sugar-web/graphics/radiob
...
+                        }else if(data[0]=="nice"){
+                            niceClockButton.onclick();
```
llaske: Same comment than the previous one.

activities/Clock.activity/js/activity.js
```diff
@@ -16,15 +16,19 @@ define(["sugar-web/activity/activity","sugar-web/env","sugar-web/graphics/radiob
...
+                        }
                         if(data[1]=="block"){
-                            clock.changeWriteTime(true);
                             writeTime=true;
+                            writeTimeButton.onclick();
```
llaske: Same comment than the previous one.

activities/Clock.activity/js/activity.js
```diff
@@ -16,15 +16,19 @@ define(["sugar-web/activity/activity","sugar-web/env","sugar-web/graphics/radiob
...
                         }
                         if(data[2]=="block"){
-                            clock.changeWriteDate(true);
-                            writeDate=true;;
+                            writeDate=true;
+                            writeDateButton.onclick();
```
llaske: Same comment than the previous one.

llaske:
Works good now. Thanks.

PR #238, opened 2018-11-15
Blockrain activity  enhancement#203

AndreaGon:
Added css for continue button.

### DIFF
--- a/css/activity.css
+++ b/css/activity.css
@@ -1,26 +1,206 @@
+* {
+  -moz-user-select: none;
+  -webkit-user-select: none;
+}
+
 #main-toolbar #activity-button {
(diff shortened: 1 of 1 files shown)


llaske:
My idea is not to have a continue button: once you've lost the game, you've lost the game. Adding a continue button allow player to cheat.
My idea is to add a new field Hi-Score. This new field that could be displayed at the left of the play board. It will start at 0 and will be upgrade at the end of each play.
When you quit the game, the Hi-Score will be stored in the Journal so you could try each time increase it.
One other important thing: you must not update the global css file (css/activity.css). This file is reserved for Sugarizer Core UI, not for activities. Each activity use its own CSS file.

Issue #203, opened 2018-04-18
Save Blockrain activity context in Journal


llaske:
The Blockrain activity don't store its current state and high score in the Journal. So when you close the activity everything is lost.

llaske:
Blockrain activity source code is here: https://github.com/llaske/sugarizer/tree/dev/activities/Blockrain.activity
Adding datastore handling to an existing activity is explain here: https://github.com/llaske/sugarizer/blob/dev/docs/tutorial_step4.md

llaske:
Fixed in https://github.com/llaske/sugarizer/commit/8f6a6f7fd28c8e7a2b77f653f48fb1421b5d64d1.

PR #239, opened 2018-11-16
Blockrain Activity enhancement #203

AndreaGon:
Added .blockrain-Hscore-holder, .blockrain-Hscore-msg and .blockrain-Hscore-num


llaske:
Nice. That's the idea. But there is an issue. When you start a new game and obtain an high score, then play again the high score is reinitialized to zero.

llaske:
Great. It works now.

PR #240, opened 2018-11-17
Abacus does not display solution with big numbers #140 

### LINKED ISSUE #140: Abacus does not display solution
The sbacus does not display the solution if the numbers get too big. 
image
The solution gets cut off at the end. This problem could be resolved by displaying the solution on the side of the abacus.

FreddieN:
Fixes #140 
The text now dynamically changes size according to the available canvas space and the length of the sum. I thought this solution was best.
Example:
demo

Let me know if you want me to try another solution.


llaske:
Very good. The size decrease when no enough room left but it will be better if the size increase when space left increase. The text should try to have the best possible size.

FreddieN:
@llaske Could you send a screenshot so I can understand what you mean?

llaske:
Here is a screen capture: at end we could expect that the size should be larger.

pr240_issue

FreddieN:
This issue should now be fixed.

llaske:
Very good. Congratulation.

Issue #209, opened 2018-04-20
Save Speak history in Journal


llaske:
In the Sugar version of Speak, history of talk is stored in the Journal. So the user retrieve its history when he reopen the activity. It could be represented by a sort of ComboBox in the input field.

llaske:
Fixed in https://github.com/llaske/sugarizer/commit/acac9e63a5f76b4fee95f1627133560e72a5b754

PR #241, opened 2018-11-21
Save Speak History in Journal #209

AndreaGon:
Added a datalist/combo box so user can access history of talk. I used datalist so it can suggest existing talks to user while the user types.

### DIFF
--- a/activities/Speak.activity/css/activity.css
+++ b/activities/Speak.activity/css/activity.css
@@ -84,6 +84,9 @@
   font-size: 0;
 }
 
+#combo-box option{
+  font-size: 20px;
+}
 
 
 /** Select tag custom css **/
--- a/activities/Speak.activity/css/activity.css
+++ b/activities/Speak.activity/css/activity.css
@@ -59,6 +59,7 @@
 .dropdown:before {
   content: "";
   position: absolute;
+	margin-left: 5px;
   left: 23px;
   top: 27px;
 	width: 0;
@@ -85,14 +86,14 @@
 
 
 
-/** Select tag custom css. Note that the select tag is invisible **/
+/** Select tag custom css **/
 
 select {
     padding:3px;
     margin: 0;
     background: transparent;
     color:#888;
-    border: none !important;
+    border: none !important;;
     outline:none;
     display: inline-block;
     -webkit-appearance:none;
@@ -137,6 +138,7 @@ body {
 	  background-color: black;
 	  border: 2px solid #808080;
 	}
+	
 
 }
 
@@ -165,14 +167,14 @@ body {
 @media screen and (max-width: 740px) {
 	select {
 		position: absolute;
-		right: 660px;
+		right: 670px;
 	}
 
 }
 @media screen and (max-width: 667px) {
 	select {
 		position: absolute;
-		right: 590px;
+		right: 610px;
 	}
 
 }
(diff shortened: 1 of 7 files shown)


llaske:
Nice.
The idea of combo is good but I don't think it's a good thing to filter content while the user type text.

llaske:
It's better but still few issues:
- The triangle overflow the white rectangle (see capture below) and it's not responsive (change size when window is resized). May be it's easier to don't add a white rectangle under the combo arrow
- The combo should be accessible at any time so each time a new text is spoke, it should be added immediately to the combo. Currently you have to close/reopen the activity to see history
- Do not update sugar-96dpi.css which is a CSS from Sugar-Web library, just update the local activity.css file.

llaske:
Very nice.
Few remarks:
- Could you remove all content you changed in sugar96-dpi.css?
- Is it possible to remove the checked sign in the combo on the selected line? If no, could you put the selected mark on the last generated text
- It will be better to add a small margin between the triangle and the input field

AndreaGon:
Hi!
I cannot seem to find the checked sign in the combo box. I am using Mozilla Firefox and it doesn't show any checked sign in the selected line. Is it possible if you can give me a screenshot? I have a feeling that it depends on what kind of browser that you are using.

llaske:
Here's a screen capture.

capture d ecran 2018-11-24 a 15 27 58

It's both on Chrome and Firefox on my side.
I guess it's the selected line in the combo.

llaske:
It do not solve the issue.
I think you should play with the "value" attribute of the "option" tag. The checked line change when you change the "value" attribute of the "select" tag with the value of the element you want to select. So, just set a value for each line and when a line in the combo is selected, force the combo to the value of this line.
Plus you must not add code (script) directly in the HTML file, it's a bad practice and it's not supported on some platform (ChromeOS). So if you want to add and event on the combo, just set  addEventListener in the main Javascript file for activity.

activities/Speak.activity/css/activity.css
```diff
@@ -42,13 +42,67 @@
...
+#combo-box {
+  position:absolute;
+  top: 5px;
+  width:2vw;
+  height:25px;
+  line-height:20px;
+  margin:13px;
+  margin-left:3vh;
+  padding:0;
+  font-size: 0;
```
FreddieN: Think this might be a mistake! The font size is too small.
demo

FreddieN: In Chrome, this causes the actual drop-down text to not appear as well. To fix add a new CSS selector with a normal font size. (see here: https://github.com/llaske/sugarizer/pull/250/files#diff-6b8bcb22150cf3bd907abf1eaa17ba77R170)
demo

FreddieN: This `font-size: 0;` hides the drop-down text but also hides the drop-down options as well in some browsers (chrome). You can fix this by adding a CSS selector to the option element and setting the font size to normal. See here: https://github.com/llaske/sugarizer/pull/250/files#diff-6b8bcb22150cf3bd907abf1eaa17ba77R174
Chrome:
demo

AndreaGon: Oh, I see. I don't use Chrome because for some reason, Sugarizer doesn't work well in my computer if I run it using Chrome. Anyway, thanks for letting me know, I'll work on it :)

AndreaGon: Fixed it, thanks! :)

llaske:
It works very good now. Congratulation.
Could you localize the title of the combo "Saved talk"? It appears on iOS on top of the combo (see below), so it's better to have a localized string here.

capture d ecran 2018-11-26 a 17 34 02

Could you also remove all update you've done in sugar-96dpi.css: there is still some lines about focus. If needed, you should put it into activity.css.

llaske:
Great. Thanks.

Issue #242, opened 2018-11-22
Add an erase button into GridPaint activity


llaske:
It could be nice to add an erase button in GridPaint activity toolbar.
This button should be visible only on the paint screen (not on the global view with all paintings).
When you click on it, it will paint all triangles of the current painting in white. 
The appearance of the button must be the same than the Erase button in Paint or Physics activity.

Code of GridPaint activity is here: https://github.com/llaske/sugarizer/tree/dev/activities/Gridpaint.activity

llaske:
Fixed in https://github.com/llaske/sugarizer/pull/249

PR #249, opened 2018-11-24
Gridpaint activity: Add clear button, clean up code

yummypasta:
I added a Clear button to the Grid Paint activity (solves #242 ). The button only shows in "edit" mode, and `onclick` it turns all triangles white. I can easily change this to a "fill" button that fills the background with the desired color, which I think might make more sense if the user wants, say, a black background.

Additionally, I deleted some unused code left over from what I presume was a copy/paste from another activity.


llaske:
Good work.
BTW, could you hide the button at startup, else there is a small glitch when the activity is launched: the button is visible then disappear.

yummypasta:
@llaske Fixed!

llaske:
Nice. Thanks.

PR #250, opened 2018-11-24
Add an history of last QR Code generated/read in QR Code activity 

### LINKED ISSUE #244: Add an history of last QR Code generated/read in QR Code activity
Add an history of QR Code generated (or read). So the user could retrieve its history when he reopen the activity. The history could be represented by a sort of ComboBox in the input field.
The history should be stored in the datastore. See https://github.com/llaske/sugarizer/blob/dev/docs/tutorial_step4.md for more.
The source of the QR Code activity is located here: https://github.com/llaske/sugarizer/tree/dev/activities/QRCode.activity

FreddieN:
Fixes #244
This PR implements the functionality to restore previously generated barcodes. See below:
Choose QR's with dropdown:
demo
QR's being saved to datastore:
demo

### DIFF
--- a/activities/QRCode.activity/css/activity.css
+++ b/activities/QRCode.activity/css/activity.css
@@ -142,3 +142,31 @@
 	visibility: hidden;
 	z-index: 4;
 }
+#qrtextdropdown {
+	width: 89.7vw;
+	height: 60px;
+	border-radius: 22px;
+	padding-left: 2vh;
+	margin-left: 3vh;
+	-webkit-appearance: button;
+	-moz-appearance: button;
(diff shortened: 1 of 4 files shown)


llaske:
That's the idea but I'm not a big fan to add a new line for the combo. It take lot of space.
I suggest to use and hidden combo, like in https://github.com/llaske/sugarizer/pull/241

FreddieN:
Is this any better?
demo

FreddieN:
I've added these changes.
demo

llaske:
Works very well. Thanks.

Issue #208, opened 2018-04-18
Improve Maze activity to allow playing with presence


llaske:
Integrate [presence](https://github.com/llaske/sugarizer/blob/dev/docs/tutorial_step6.md) handling into Maze activity to allow playing the activity with other users on the network.

llaske:
Source code for Maze activity is located here:  https://github.com/llaske/sugarizer/tree/dev/activities/MazeWeb.activity

llaske:
Fixed in https://github.com/llaske/sugarizer/pull/251

PR #251, opened 2018-11-24
Update Maze activity to handle multi-user with presence. Fixes issue #208

sdziuda:
Updated files to handle multi-user presence.


llaske:
It's a good start but a player should not be allowed to change level if other players have not complete the level.
Today it's already possible to play with multiple players but on the same computer: one player use mouse, one another use arrow keys, one use I/J/K/L keys and the last one use W/A/S/D keys.
I guess you could have a look in the code how this multi-user mode is implemented to try to do the same.

sdziuda:
I've updated the files, moved all the commands from maze.js (and now it is not needed, so I removed it) to activity,js, because it was causing the issue that one player can't move. Now both can. The activity also wait for both players to accomplish the maze but there's a bug, that occurs after one of the player's win, that the other has two faces on start. I'll be working on this.

llaske:
Very good work. Thanks.

llaske:
FYI I've updated the activity here https://github.com/llaske/sugarizer/commit/a906c142e6fba9b1257314c47182112506db0bd3 to handle an non limited number of users and to add some popup during the game.

sdziuda:
Wow, nice addons!

Issue #174, opened 2018-01-08
No San-Marino and Vatican in Color My World app.


nikitaermishin:
There is no such countries, as San-Marino and Vatican in Color My World app (screenshot below).

![2018-01-09 01-24-51](https://user-images.githubusercontent.com/26204565/34695624-a427a9a2-f4dc-11e7-88e2-f0533e6d094b.png)

It's is valueble issue, because San-Marino and Vatican are small, but countries.

llaske:
Both Vatican and San Marino are present into the activity. See [here](https://github.com/llaske/sugarizer/blob/master/activities/ColorMyWorld.activity/locale.ini#L266) and [here](https://github.com/llaske/sugarizer/blob/master/activities/ColorMyWorld.activity/data/europe/europe.geojson). Probably very small due to the scale of the map.
Did you try to click on it?

nikitaermishin:
Yes, I tried, and I changed map size, and I've got nothing.

llaske:
Source code for Color My World activity is located here: https://github.com/llaske/sugarizer/tree/dev/activities/ColorMyWorld.activity

llaske:
Fixed in https://github.com/llaske/sugarizer/pull/253

PR #253, opened 2018-11-28
San Marino and Vatican City #174

AndreaGon:
I added the coordinates for Vatican City and San Marino. They are not in world.geojson that is why they are missing.


llaske:
Nice. 
Two remarks however:
- You should also update others maps (europe, asia) else it don't works when the user select these maps
- You should update the locale.ini if need.

AndreaGon:
Hello,
I only added San Marino to europe.geojson (it was missing there). Vatican City and Singapore are already on their respective maps, they are only missing on the world.geojson. I also checked locale.ini and everything seems to be in order on that file.

PR #254, opened 2018-11-29
Update LOL activity to handle multi-user with presence

sdziuda:


sdziuda:
I've added some functions that were missing in activity/activity.js file. If it wasn't compiling before, now it should work.

sdziuda:
I've resolved some of the issues (for example not the same amount of lines on init for both players) and implemented the new way to count how many lines are left.

sdziuda:
Everything should work fine now.

sdziuda:
I fixed the 2nd and 3rd point of your message.

sdziuda:
This should fix the 1st point of your message.

llaske:
Very good.
Still an issue: when the first player click on the switch button, both side wait for other. 
We could expect than the second player is able to play.

sdziuda:
You mean the Switch player button?

llaske:
Yes

sdziuda:
Everything should work fine now

llaske:
There is a bug. If you click a first time on the switch button then a second time, it seems to work but after on turn both player can play at the same time (see screen capture)

capture d ecran 2018-12-05 a 17 26 39

sdziuda:
Should it be possible to click twice on switch player button? I found this line:
`document.getElementById("switch-player-button").disabled = (this.game.getLength() != this.size);`
and it seems to disable the switch button after you or cumputer (in singleplayer) make a move. I assume that I should make it disabled after one move also in multi. Am I right?

llaske:
Yes you're right it should not be possible to switch first player after the first move. It should be the case both in single player or multi player mode.

activities/LastOneLoses.activity/js/activity.js
```diff
@@ -12,8 +12,8 @@ define(["sugar-web/activity/activity", "sugar-web/graphics/radiobuttonsgroup", "
 	var newgame=false;
 	var onejoined=false;
 	var numberofplayers=0;
-	var playerswitches=0;
-	var switched=false;
+	var switchclicked=false;
+	//var playerswitches=0;
```
llaske: You can't leave code into a comment

activities/LastOneLoses.activity/js/activity.js
```diff
@@ -174,15 +178,15 @@ define(["sugar-web/activity/activity", "sugar-web/graphics/radiobuttonsgroup", "
 				}
 			}
 			
-			if(presence&&playerswitches%2==1){
+			/*if(presence&&playerswitches%2==1){
```
llaske: You can't leave code into comment

llaske:
Nice. Thanks.

llaske:
FYI I've refactored and improved your code here: https://github.com/llaske/sugarizer/commit/3247c803b378ea70095bc297c696d1f65f0389e4
Few advices:
- You've used lot of boolean variables to handle change but most of changes are already handled by the LOLGame object. For the same reason, you didn't need to store item count and current player in global variables, LOLGame already take care of that. More generally, avoid using global variables, it's not a good practice.
- No need to add 'newgame' message: it's the same one that 'init'.
- Avoid doing "if (condition1 && !condition2) {}" followed by "if (condition1 && condition2) {}". It's better to imbricate it in "if (condition1) { if (condition2) {} else {} }" to avoid to evaluate 2 times each condition.

sdziuda:
Ok, thank you for advices!

PR #256, opened 2019-01-06
Added Emoji for ChatPrototype

AndreaGon:
I added an emoji select tag for ChatPrototype so you can use and send out emojis.

image

Feedbacks are appreciated :)


llaske:
It's a good idea. Thanks.
Few issues however:
- Should be better to use a SVG for the button, it's slightly pixelised currently
- It's not responsive, the button overlap the text field depending of the window size

image

- A very long combobox of icons is not very useful to select the good one and it will be very difficult to use on tablet.

image

AndreaGon:
I changed the layout to make it more responsive:

image

I also added 3 new buttons to distribute the emojis and I only use the good emojis. What do you think?

llaske:
It's better with 3 buttons. I've tested it on a tablet and it could be used.
BTW it's better if buttons is on the same line than the input field else a scrollbar appears.

image

Could you reduce the input field size and put buttons at the end of line ?

AndreaGon:
This should do it :)

Desktop mode:
image

Responsive:
image

llaske:
Very nice.
BTW There is an issue: you can't click on the first item on the combobox. If you want to use the first item you first need to choose another one then click again on the first item.
I guess it's related to the fact that at startup the first item of each combobox have a selected mark.

AndreaGon:
This should fix it. Let me know if there are still issues.

llaske:
Nice. It perfectly works now.
Thanks.

Issue #262, opened 2019-01-22
Add a delete button in the new idea input field of Labyrinth activity


llaske:
The new idea input field into the Labyrinth activity miss a delete button to delete the whole content of the field. So, it's difficult to completely remove the content, specifically on tablets.

<img width="404" alt="image" src="https://user-images.githubusercontent.com/981841/51566745-b492f100-1e95-11e9-9060-b367417d28c3.png">

The idea is to add a delete button into the input field: the same button that the one in the Search activity in Sugarizer home view. 
Styles to do that are described in the sample here: http://sugarlabs.github.io/sugar-web-samples/iconinputtext.html

Source code for the Labyrinth activity is available here: https://github.com/llaske/sugarizer/tree/dev/activities/LabyrinthJS.activity

lohani2280:
@llaske I would like to work on this issue.

llaske:
Sure. Feel free to work on it.

llaske:
Closed in https://github.com/llaske/sugarizer/commit/dd3c58d56981e4dfa6d94601b95d7117a3bc9241

PR #267, opened 2019-01-24
Added a delete button in the new idea input field

lohani2280:
Fixes https://github.com/llaske/sugarizer/issues/262


llaske:
That's the idea. Nice.
By the way, the button has not the Sugar UI style. It should be black instead of gray.
Plus it slightly shift when clicked, it should stay on the same position.
You could take inspiration on Sugar CSS samples here: http://sugarlabs.github.io/sugar-web-samples/iconinputtext.html

lohani2280:
@llaske Thanks for the review. I'll update the PR according and get back soon.

lohani2280:
@llaske I have updated the PR as per your suggestion. Please have a look.

PR #272, opened 2019-02-13
Update tutorial.md

precious-adeyinka:
Change the last sentence syntax,

Before:
But let's first with the [Step 0](tutorial_step0.md), set up the development environment.

After:
But first let's start with the [Step 0](tutorial_step0.md), set up the development environment.


llaske:
Thanks for your fix proposal but as explained in https://github.com/llaske/sugarizer/blob/master/CONTRIBUTING.md, contributions should be done on dev branch only.

Issue #263, opened 2019-01-22
Integrate presence in Labyrinth activity


llaske:
Integrate presence handling into Labyrinth activity to allow playing the activity with other users on the network. So users could create a collaborative mind map.

You should add the presence palette and integrate presence like explained in the tutorial here: https://github.com/llaske/sugarizer/blob/dev/docs/tutorial_step6.md
See code for Labyrinth activity here: https://github.com/llaske/sugarizer/tree/dev/activities/LabyrinthJS.activity
You could use the SharedNotes activity source code to see how handle presence in a similar context (Cytoscape framework). See code for the SharedNotes activity here: https://github.com/llaske/sugarizer/tree/dev/activities/SharedNotes.activity

lohani2280:
@llaske I would like to work on this issue.

llaske:
@lohani2280 Sure, you're welcome.

llaske:
Fixed in https://github.com/llaske/sugarizer/pull/274

llaske:
Closed in https://github.com/llaske/sugarizer/commit/dd3c58d56981e4dfa6d94601b95d7117a3bc9241

PR #274, opened 2019-02-13
Integrated presence in Labyrinth activity

lohani2280:
Refers https://github.com/llaske/sugarizer/issues/263


llaske:
Very good work. It perfectly works.

PR #277, opened 2019-02-22
Add click and drag

utkarsh-raj:
The following ideas are implemented

- Single click to move the bead up or down according to the previous state

- Drag feature to move the bead when the user wants to override the click to move logic


llaske:
Seems perfect for me.

Issue #185, opened 2018-01-29
Turtle Blocks Activity hangs on Android App


manki11:
The turtle blocks activity hangs in Android App, when the Load Samples from web icon is clicked (world icon in top right menu). After clicking on the button , any other click after that registers as a load samples from web.

![error](https://user-images.githubusercontent.com/25136650/35530034-dfa2a874-0558-11e8-9933-b88b363562fd.gif)

Not facing any similar behaviour on sugarizer web.

Device Tested on-
Oneplus 5T
Android 7.0.1

llaske:
Seems that load samples don't work on web version.
There is a JavaScript issue here: https://github.com/llaske/sugarizer/blob/dev/activities/TurtleBlocksJS.activity/js/samplesviewer.js#L213

<img width="654" alt="image" src="https://user-images.githubusercontent.com/981841/36635408-856189c8-19b4-11e8-97b7-892f7fbfb19e.png">

divyabaid16:
I would like to work on this issue. Can you plz guide me ?

llaske:
Turtle Blocks activity come from https://github.com/walterbender/turtleblocksjs.git .
I suggest you to fork this repository, and try to understand how it works by studying the code.
Then you could see what's wrong when running on the Sugarizer environment.

utkarsh-raj:
https://github.com/llaske/sugarizer/pull/279#issue-255672546

A part of the error has been fixed by cleaning the String before parsing with the parse function. I cannot test the complete functions because the sugarizer server probably does not allow the request without CORS headers which are not present by default in the request.

Nonetheless, the problem is fixed. Please review @llaske .

utkarsh-raj:
The commit is updated @llaske .

llaske:
Fixed in https://github.com/llaske/sugarizer/pull/279

llaske:
Closed in https://github.com/llaske/sugarizer/commit/dd3c58d56981e4dfa6d94601b95d7117a3bc9241

Issue #106, opened 2017-03-09
Memorize responsive seems broken on phones


mikklfr:
The paint activity button seems to have troubles with small screen.
Some CSS should be able to fix it.

dinesh-sirvi:
Couldnt able to reproduce the issue, Can u elaborate it more or provide me the steps to reproduce it?

mikklfr:
My bad, it's the memorize activity. I'm updating the issue.

You can see the quit button going out of the toolbar on this photo.

https://www.facebook.com/photo.php?fbid=10154521813268510&set=pcb.10154521813863510&type=3&theater

dinesh-sirvi:
I'll Check it....

rajeevravindran:
Can't reproduce the problem. @mikklfr Can you give the viewport dimensions of the browser on your phone?

mikklfr:
I've been able to reproduce using the Galaxy S5 viewport in Google Chrome debug. 640 x 360

rajeevravindran:
![image](https://cloud.githubusercontent.com/assets/10564476/23852351/56a179cc-080d-11e7-8fbc-e5a02cc82af5.png)

`.toolbar{
 padding : 0 50px;
}`

This solves it. Creating a PR

mikklfr:
Better but I still have issues using
 590px  <=  width <= 638px

rajeevravindran:
Yes. Maybe an @media query for this range? 570px - 638px

rajeevravindran:
![image](https://cloud.githubusercontent.com/assets/10564476/23861468/1f5b9268-0830-11e7-900a-807885dd80de.png)
Seems to be working. I'll make a PR? 
`@media screen and (min-width: 590px) and (max-width: 638px){
  .toolbar{
    padding: 0 25px;
  }
}`

utkarsh-raj:
Some insight - 

- The responsiveness is broken across several activities (including Memorize, the original activity in question)

- In some other activities, the Stop button simply goes away on some viewports, which can be verified by the Google Chrome Developer Tools

- The issue is also there for several modern devices like the iPhone X

Probably some CSS for modern viewport breakpoints should be able to fix this. But the important point is that the issue has a much broader scope than only Memorize. I am currently working on this and some reviews will be great :)

![screenshot from 2019-02-21 23-53-25](https://user-images.githubusercontent.com/28586681/53192356-46785000-3634-11e9-8a6b-1124fc5286e0.png)
[...]

PR #279, opened 2019-02-24
Fix unexpected token in JSON Error

utkarsh-raj:
The original mentioned error was caused due to some non printable characters in the JSON String. These characters do not show but can break the parse function in the JSON Object in JS.

This was happening because the string was not cleaned before parsing. Applied standard cleaning techniques.

### DIFF
--- a/activities/TurtleBlocksJS.activity/js/samplesviewer.js
+++ b/activities/TurtleBlocksJS.activity/js/samplesviewer.js
@@ -210,8 +210,18 @@ function PlanetModel(controller) {
 
     this.redoLocalStorageData = function () {
         this.localProjects = [];
-        var l = JSON.parse(localStorage.allProjects);
-        l.forEach(function (p, i) {
+        s = JSON.stringify(localStorage);
+        s = s.replace(/\\n/g, "\\n")  
+               .replace(/\\'/g, "\\'")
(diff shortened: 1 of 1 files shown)


llaske:
Please do not integrate two fixes in the same PR. This one contains #106 and #185. Thanks to split into two PR.

Regarding #185, your change seems to fix partially the issue. BTW CORS should be installed on the server, could you ask to Walter bender to add it?

utkarsh-raj:
I am sorry. I will split the fix into two PR.

#185 is a partial fix, the issue now raised is probably an altogether different one. The console errors are regarding the server configuration, and the App must not hang on any device as there is no critical error.

I think that this fix solves the original problem for which the issue was opened, and the new issue can be opened for server configuration.

I will research a little and ask Walter Bender to review the suggestions.

utkarsh-raj:
I have reverted the fix of the other issue from this branch. Please check @llaske .

utkarsh-raj:
I am little unsure of the root cause. The App does not hang on any device because the JS error is fix. But the request URL does not look right to me. However, discussing this problem will deviate from the issue being raised in #185 . Should a new issue be open for @llaske ?

llaske:
@utkarsh-raj yes it's a good idea to open a new issue on the CORS problem.

PR #284, opened 2019-02-28
Tutorial documentation update

NikhilM98:
I was going through the Pawn Activity tutorial. I found out that on adding every new pawn, whole `pawns` div is being re-rendered.
Suppose if you're adding N pawns one at a time, then the pawn div gets reset N times and a new pawn div is added N! (N factorial) times. It has a significantly higher time complexity than adding one pawn at a time which has a time complexity proportional to N.
When neighbor connects, I've kept in mind to clear the canvas and draw all the pawns in one go.

I have made appropriate changes in the documentation, keeping it simple.
I'm open to suggestions if you find any issue with my implementation.

### DIFF
--- a/docs/tutorial_step4.md
+++ b/docs/tutorial_step4.md
@@ -58,26 +58,24 @@ First we will slightly refactor (i.e. improve) our `activity/activity.js` file. 
 		document.getElementById("user").innerHTML = "<h1>"+currentenv.user.name+" played !</h1>";
 	});
 
-We will refactor it to store pawns in an array `pawns` and to add a new method `drawPawn` to draw all icons at one time. So instead of adding a new `div` element on each button click, we just add a new element in the `pawns` array and we call the `drawPawn` icon to redraw the board. Here is the resulting code:
+We will refactor it to store pawns in an array `pawns` and to add a new method `drawPawn` to draw one icon on the board. So on each button click, we add a new element in the `pawns` array and we call the `drawPawn` icon to draw one icon at a time on the board. Here is the resulting code:
 
 	// Draw pawns
 	var pawns = [];
-	var drawPawns = function() {
-		document.getElementById("pawns").innerHTML = '';
-		for (var i = 0 ; i < pawns.length ; i++) {
-			var pawn = document.createElement("div");
-			pawn.className = "pawn";
 
-			document.getElementById("pawns").appendChild(pawn);
-			icon.colorize(pawn, pawns[i]);
-		}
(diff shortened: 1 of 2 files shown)


llaske:
It's just a pedagogic sample to learn how to develop a Sugarizer activity. It's not intended to learn good programing practice. So the source code should be as simple as possible.
It's not pertinent to add complexity if it's not related to the pedagogic objective.

llaske:
It's just a pedagogic sample to learn how to develop a Sugarizer activity. It's not intended to learn good programing practice. So the source code should be as simple as possible.
It's not pertinent to add complexity if it's not related to the pedagogic objective.

NikhilM98:
@llaske I have gone through the tutorial and I understand that its objective is to make users understand the basics of creating Sugarizer web activity. The tutorial is well written and is easy to understand even by new programmers.
I made changes, keeping the code simple. My objective was to make it clear to the users that they do not need to re-render the whole div again even as in the case of HTML canvas where the whole canvas is reset after every frame and `requestAnimFrame()` function is called. Its not about providing explicit documentation about time complexity and rendering optimization concepts, since this tutorial is one of a kind and it would also be used by Moderate/Advanced users, hints are provided in the code that will be out of sight of users who are not familiar with these concepts. Keeping the changes minimal.

llaske:
You're wrong, the board is  not a Canvas HTML object, it's just a div object so there is nothing like a refresh process on every frame. So no need to add complexity here.

NikhilM98:
Sorry @llaske I think I made a typo in the last comment.

> My objective is to make it clear to the users that they do not need to re-render the whole div again as in the case of HTML canvas

I was trying to say "Unlike Canvas".
I think someone reading the documentation can probably develop an understanding that to update a parent div, he has to clear it and then start creating its child nodes again. I was trying to make it clear to the user that this is not the case.

llaske:
> I think someone reading the documentation can probably develop an understanding that to update a parent div, he has to clear it and then start creating its child nodes again.

I'm not agree. Understanding of JavaScript and DOM is a pre-requisite to write a Sugarizer activity. I don't expect to give good JavaScript practice in this tutorial. Focus is on specific Sugarizer mechanism, not on JavaScript way of working.

NikhilM98:
Thanks for the response @llaske
I think I should close this PR if the understanding of JS and DOM is a pre-requisite for the user.

I was trying to make PR on Exerciser but no one has reviewed them so far. Can you please review [them](https://github.com/llaske/ExerciserReact/pulls)

Issue #285, opened 2019-03-04
Exerciser Activity - Improvement add confirmation delete dialogue Box.


saadpasta:
In Exerciser Activity when we click on delete to delete any test the application deletes it immediately.
I think we should add Confirm box to confirm if they actually want to delete the test.

Scar26:
Working on it

Scar26:
Solved the issue.Can anyone guide me on how to merge it?
Sorry. I'm new here

saadpasta:
I was also working on but it's okay.
First, push your code on your forked repo and then Create MR button and then choose your branch on which you have worked and then click create.

saadpasta:
https://help.github.com/en/articles/creating-a-pull-request-from-a-fork
Read this.

Scar26:
Was that correct?

saadpasta:
Yeah Great 👍

llaske:
Should be posted to Exerciser repo: https://github.com/llaske/ExerciserReact

PR #286, opened 2019-03-04
Addded confirm message for deletion in Exerciser activity

Scar26:
Added a confirmation message for dialogue deletion in Exerciser activity as requested in #285 .
Hopefully this fixes the issue.


Scar26:
@llaske I was thinking of styling the confirm button to match the rest of the UI instead of using the default in electron. Is it fine if I use Jquery UI or should I stick to vanilla javascript?

llaske:
Exerciser activity is written using ReactJS. Only the build code is integrated in Sugarizer.
So for this activity, don't do a PR here, do it on https://github.com/llaske/ExerciserReact

Scar26:
@llaske Done

llaske:
Related to https://github.com/llaske/ExerciserReact/pull/11

Issue #292, opened 2019-03-12
Display language settings in 2 languages


YJDave:
Languages in `Language setting` should not be translated according to local/browser language. So "English" should be "English" even if my local language of browser is "French" or something else.

![image](https://user-images.githubusercontent.com/25907420/54201621-1fef5b80-44f4-11e9-9a00-8b5cc0b11cfc.png)

Why?
Suppose I change my local language to "French", languages list in setting also changed, and I am no longer be able to change language back to "English" as I couldn't understand which one of the option was "English" as all options were translated to "French".

llaske:
I don't think it's a problem. English is at first and name of languages are understandable in most languages. 
I never seen this sort of double translation elsewhere. Did you?

YJDave:
I was not sure whether double translation is the most suitable solution. But I do think that there should be some way that allows user to see languages names without translated to local language. 
I have seen this double-translated language list setting on Twitter, not checked other applications. 
![image](https://user-images.githubusercontent.com/25907420/54252939-d2660380-4572-11e9-9d24-f96197eabd27.png)

What could be the appropriate solution to this problem, according to you?

llaske:
Fixed in https://github.com/llaske/sugarizer/commit/0cd4903a18c6c35a81589e2ac916530a0710ed87

YJDave:
Closing the issue as it is fixed!

llaske:
I don't close issue before release of a new version.

llaske:
Closed in https://github.com/llaske/sugarizer/commit/dd3c58d56981e4dfa6d94601b95d7117a3bc9241

PR #293, opened 2019-03-12
#292: Languages in "Language settings" should be non-translated.

YJDave:
Languages list in User settings->Language settings should
be non-translated to local-browser language.
Fixes issue #292 
Now:
If local language is English:
image

If local language is French:
image

In `()` it shows language name translated to local language and without `()` it shows original language name without being translated to any other language. 

This is my first contribution, so please let me know your feedback.

### DIFF
--- a/js/dialog.js
+++ b/js/dialog.js
@@ -580,16 +580,16 @@ enyo.kind({
 		this.$.message.setContent(l10n.get("ChooseLanguage"));
 		this.initlanguage = this.currentlanguage = preferences.getLanguage();
 		this.languageset = [
-			{code: "en", icon: null, name: l10n.get("English")},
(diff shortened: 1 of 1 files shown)


sum2it:
The idea of using ( ... ) is interesting.
Seems more user friendly than the previous convention.

YJDave:
Thanks @sum2it for the feedback :)
It is surely more easy to understand comparing to previous one, as it shows list in both languages(local one and its original).

llaske:
Thanks.

NikhilM98:
@llaske I think it'll be nice to implement this in Sugarizer-Server too.

ashish0910:
@NikhilM98 agreed

Issue #302, opened 2019-03-20
Translation warnings on Home


Anshuman71:
![Screenshot from 2019-03-20 13-38-49](https://user-images.githubusercontent.com/28081510/54668744-a3710400-4b15-11e9-8ba6-5b0543fd8f5e.png)

AndreaGon:
Can I work on this?

llaske:
You're welcome @AndreaGon. BTW no need to ask to work on an issue, see https://github.com/llaske/sugarizer/blob/master/CONTRIBUTING.md

Anshuman71:
@AndreaGon sure as @llaske said you can contribute any update as long as it fixes the bug.

llaske:
We explore all solutions here: https://github.com/llaske/sugarizer/pull/303
No one is interesting. 
The warning message could just be ignored.

PR #303, opened 2019-03-22
Fix Translation Warnings on Home #302

AndreaGon:
This should fix Issue #302.

### DIFF
--- a/js/homeview.js
+++ b/js/homeview.js
@@ -233,7 +233,7 @@ enyo.kind({
 
 	localize: function() {
 		if (this.otherview && this.otherview.localize) {
-			this.otherview.localize();
+			this.otherview.localize;
 		}
 	},
 
@@ -320,7 +320,7 @@ enyo.kind({
 				} else if (i > 0 && i == this.restrictedModeInfo.start) {
 					this.$.desktop.createComponent({
 							kind: "Sugar.Icon",
-							icon: {directory: "icons", icon: "activity-etc.svg", name: l10n.get("ListView")},
+							icon: {directory: "icons", icon: "activity-etc.svg", name: "Previous"},
 							size: icon_size,
 							x: ix,
 							y: iy,
@@ -331,7 +331,7 @@ enyo.kind({
 				} else if (i >= this.restrictedModeInfo.start+activitiesCount-1 && this.restrictedModeInfo.start + activitiesCount < activitiesList.length) {
 					this.$.desktop.createComponent({
 							kind: "Sugar.Icon",
-							icon: {directory: "icons", icon: "activity-etc.svg", name: l10n.get("ListView")},
+							icon: {directory: "icons", icon: "activity-etc.svg", name: "Next"},
 							size: icon_size,
 							x: ix,
 							y: iy,
@@ -732,7 +732,7 @@ enyo.kind({
 
 	rendered: function() {
 		this.inherited(arguments);
-		this.localize();
+		this.localize;
 	},
 
 	localize: function() {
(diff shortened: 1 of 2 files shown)


js/homeview.js
```diff
@@ -233,7 +233,7 @@ enyo.kind({
 
 	localize: function() {
 		if (this.otherview && this.otherview.localize) {
-			this.otherview.localize();
+			this.otherview.localize;
```
llaske: Very strange fix: removing parenthesis is equivalent to don't call the function at all.

js/homeview.js
```diff
@@ -732,7 +732,7 @@ enyo.kind({
 
 	rendered: function() {
 		this.inherited(arguments);
-		this.localize();
+		this.localize;
```
llaske: Very strange fix: removing parenthesis is equivalent to don't call the function at all.

js/homeview.js
```diff
@@ -743,7 +743,7 @@ enyo.kind({
 		this.$.neighborbutton.setNodeProperty("title", l10n.get("NeighborhoodView"));
 		this.$.helpbutton.setNodeProperty("title", l10n.get("Tutorial"));
 		if (app.localize) {
-			app.localize();
+			app.localize;
```
llaske: Very strange fix: removing parenthesis is equivalent to don't call the function at all.

AndreaGon:
I restored the parenthesis in this latest commit. Let me know if there are still some issues.

js/homeview.js
AndreaGon: Restored parenthesis

AndreaGon:
Please check @llaske

llaske:
That's better but tooltips on buttons (Help, Neighborhood, ...) is missing now.

AndreaGon:
@llaske Are the tooltips only missing when the screen size is small? When I tried it on a normal window size, the tooltips for the buttons show themselves properly.

llaske:
Hmmm, you're right, the tooltip disappear when the window is resized then do not appear again even if the window is enlarged.

AndreaGon:
@Ilaske I guess it's normal for the title tooltip on buttons to disappear when the screen size is small as I do not think that mobile browsers support this, as I tried restoring the original code and the output is the same as the modified code. Also, the tooltip seemed to work properly if the window is enlarged again. What do you think?

llaske:
Yes, you're right.
By the way I've found a side effect of your fix: tooltip and search label disappear when coming from the neighborhood view.
Steps to reproduce:

- Go to the neighborhood view and join an existing activity (created by another user)
- Stop the activity
- Go to the home view
- The tooltip and search label has disappeared 

If you test the same thing on dev.sugarizer.org, it works correctly.
Not proud of that but there is lot of imbrication on the localization functions. Thanks for your help on this.

AndreaGon:
Hello!
Based on the PR on the webL10n library, loading and parsing the locale.ini file is happening after webL10n starts localizing. In my recent PR, I tried a workaround suggested in the webL10n PR, which worked for Sugarizer. I would appreciate your opinion.

llaske:
Hmmm. Interesting but it will replace a (bunch of) warning by another:

> webL10n.js:124 [Deprecation] Synchronous XMLHttpRequest on the main thread is deprecated because of its detrimental effects to the end user's experience. For more help, check https://xhr.spec.whatwg.org/.

Considering that translation warnings could be ignored because the UI is refreshed when the localization is loaded, I wonder if the good solution is not just to remove log by setting `gDEBUG` to 0 [here](https://github.com/llaske/sugarizer/blob/master/lib/webL10n.js#L57).

What do you think?

AndreaGon:
Setting the variable gDEBUG to 0 does remove the translation warnings. However, it might be harder to see possible future localization bugs with gDEBUG turned off. In my opinion, the translation warnings on home should just be ignored, as it doesn't have any bad effects in Sugarizer. 

Thoughts? @llaske

llaske:
Okay. Thanks for your contribution.

PR #315, opened 2019-03-27
Improvements in README

prabhupant:
Improvements in the readme file. Adds proper headings and paragraphs for every information in the readme


llaske:
I don't see improvement here. And the split on Desktop environment is not really pertinent.

Issue #314, opened 2019-03-27
Duplicated text in QR Code Activity history


yog24esh:
I think there is no use of logging the same user in the dropdown menu if he generate his qrcode again. 

Here is the ex-

![Screenshot (38)](https://user-images.githubusercontent.com/16835166/55045249-30124980-5063-11e9-80b1-7afa64ab06e8.png)

Here what to be expected - 

![Screenshot (39)](https://user-images.githubusercontent.com/16835166/55045435-c5154280-5063-11e9-922c-0f512b31dc50.png)

Scar26:
working on this

yog24esh:
Have a look in qrtextdropdown id and attach the if statement to check if previous div value is same or not.

![Screenshot (42)](https://user-images.githubusercontent.com/16835166/55092207-25de6280-50d8-11e9-9d14-c47c3dd1877c.png)


* I think before making/working we should ask one of the maintainers.

Scar26:
This wouldn't work. the names are being stored in a global array called history. If I run the condition check at the div change,the array would still be different every time because it would just contain the same name multiple times.
Instead I attached the statement at the beginning of the pushtohistory function so if the name already exists in the array it's not pushed at all

yog24esh:
Yes but instead of checking the whole array , checking the last pushed element in the array would be useful otherwise it would defeat the purpose of history.
Ex - If a user1 generate qrcode.  arr->[user1]
user2 generate qrcode. arr->[user1,user2]
user1 again generate qrcode. arr->[user1,user2,user1]
user1 is already present in array but it should be logged because it shows the history.

Scar26:
Hm...makes sense,I mean I probably could just let it be logged into history and run the test in the for loop that they are using to append in the div. But it would be a little less efficient since we have to check a higher number of elements every-time.  On the other hand, it makes sense that history should contain everything.

@llaske can we have your opinion on this?

llaske:
Easy access to old generated QR code is more useful than a real history with duplication.

So https://github.com/llaske/sugarizer/pull/317 fixed the issue in my opinion.
A supplemental fix is here: https://github.com/llaske/sugarizer/pull/339

llaske:
Closed in https://github.com/llaske/sugarizer/commit/dd3c58d56981e4dfa6d94601b95d7117a3bc9241

PR #317, opened 2019-03-28
Removed log redundancy in QRCode activity

Scar26:
prevented the same word from being logged multiple times as mentioned in #314


Scar26:
Done @llaske

activities/QRCode.activity/js/activity.js
```diff
@@ -210,9 +210,11 @@ define(["sugar-web/activity/activity","sugar-web/datastore", "sugar-web/env", "w
 
 		// Add entry to history
 		function addToHistory(text) {
+			if(!history.includes(text)){
 			history.push(text);
 			updateHistory();
 		}
+		}
```
llaske: Could you realign code to respect indentation?

Scar26: done @llaske

llaske: No you don't. Of course, line 213 and 214 must be aligned with block.

Scar26: Is it fine now? @llaske

llaske:
Ok thanks. It fix the issue.

BTW the history content is case insensitive but QR code are case sensitive.
So it's a bug to consider history as case sensitive. It should be nice to fill a bug on this.

PR #320, opened 2019-03-28
Audio bar display in media viewer

### LINKED ISSUE #312: audio bar not displaying in chrome when opening a audio file(.wav) through journal
When a user tries to open an audio file in .wav format, no audio bar is displayed in the media viewer activity.

AvinashAgarwal14:
Fixes #312 
Currently, if a user tries to play an audio(.wav) file from the journal media activity viewer starts, but the audio bar is not displayed. This PR solves the issue:

Before:

before_audio

After: 

after_audio


llaske:
Nice. Thanks.

Issue #330, opened 2019-03-31
Users should be able to send images in chat prototype


Scar26:

llaske:
Why not, could be nice.

llaske:
Fixed in https://github.com/llaske/sugarizer/pull/331

llaske:
Closed in https://github.com/llaske/sugarizer/commit/dd3c58d56981e4dfa6d94601b95d7117a3bc9241

PR #331, opened 2019-03-31
Users can now send images in chat prototype

Scar26:
#330 

Now:
imgup


activities/ChatPrototype.activity/js/activity.js
```diff
@@ -124,6 +144,18 @@ define(["sugar-web/activity/activity","webL10n","sugar-web/graphics/palette","su
 				return false;
             }
         };
+
+        imageUpload.onchange = function(){
+        	var image = imageUpload.files[0];
+			var reader = new FileReader();
```
llaske: For compatibility, images should be taken only from the Journal. See how to include the Journal chooser dialog box here: https://github.com/llaske/sugarizer/blob/dev/docs/tutorial_step7.md

Scar26:
Journal integrated.
Now:
journal

Defined a separate function for sending data through the presenceObject making it easier to accommodate future updates.

Also noticed a slight discrepancy in the code given in https://github.com/llaske/sugarizer/blob/dev/docs/tutorial_step7.md

Scar26:
@llaske Are there any other improvements I could make in this PR?

Scar26:
@llaske  Are there any changes I should make with respect to this PR?

Scar26:
@llaske Are there any changes I should make with respect to this PR?

Scar26:
@llaske  I have made the requested changed. Can you please review this PR?

llaske:
Could you solve conflict on last dev branch?

Scar26:
done @llaske

llaske:
Nice work.
Few remarks:
- The icon quality is bad, should be nice to use the "standard" Sugar icon to include an image
- It will be better to put the icon into the toolbar
- The user should not be able to integrate an image before connection
- You broke compatibility with a client that do not include this feature. So an "old" version of the app will not be able to communicate with this new version
- Please fix conflict

Scar26:
Now : 
toolbar

-  Icon changed. This is the icon used in record activity as well as QRcode activity.
-  Image Upload icon moved to toolbar
-  The icon is online visible when the user connected, therefore they cannot integrate an image before connection
-  Conflict fixed

An old version of the app will still be able to exchange text messages with a newer version since I'm only sending a new member "type" in the message object. It just won't be used anywhere. 

In the case that a user on the newer version sends an image, the old version will display the base64 code of the image. Currently, I see no way of dealing with this bug. Even with changes to the Sugarizer server, note that the userInfo object that is sent to the server does not contain the application version so the server won't be able to decipher what to send to a socket. so this is just a glitch that's gonna persist unless everyone in the neighborhood updates to the new version.

llaske:
Nice. The image is good now.
There is still a conflict: If an old client try to send a message to the new client, the new client don't display the old client message (just ignore it). BTW old client is able to display new client message (expect images but it's okay).

Scar26:
Hm...I'll look into it

Scar26:
Done @llaske 

Now older versions can interact with the new one

Scar26:
@llaske I have made the requested changes, can you please Review this PR?

activities/ChatPrototype.activity/css/activity.css
```diff
@@ -136,17 +142,21 @@ textarea {
   background-repeat: no-repeat;
   margin-top: 8px;
   margin-left: 10px;
+  font-size: 0;
```
llaske: Should be removed to follow https://github.com/llaske/sugarizer/pull/323

activities/ChatPrototype.activity/css/activity.css
```diff
@@ -136,17 +142,21 @@ textarea {
...
 }
 
 #sad-button {
   position: absolute;
   margin-top:8px;
   margin-left: 55px;
+  font-size: 0;
```
llaske: Should be removed to follow https://github.com/llaske/sugarizer/pull/323

activities/ChatPrototype.activity/icons/insert-picture.svg
```diff
@@ -0,0 +1,40 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
llaske: Do you still need this file?

main.js
```diff
@@ -119,6 +119,7 @@ function createWindow () {
 		mainWindow.maximize();
 	}
 
+
```
llaske: Do not upgrade this file

Scar26:
Done @llaske

llaske:
Ok. Nice.

PR #337, opened 2019-04-09
Remove calls of postRunActivity when using Native apps

mikklfr:
Remove calls of postRunActivity when using Native apps


llaske:
Reference https://github.com/llaske/sugarizer/issues/338

PR #339, opened 2019-04-10
QR code History case sensitive

ashish0910:
This pr solves the issue mentioned [here](https://github.com/llaske/sugarizer/pull/317#issuecomment-478325854)
Earlier the history content was case insensitive but QR code were case sensitive . In this PR history is also made case sensitive 
Current:
current
Expected:
expected

### DIFF
--- a/activities/QRCode.activity/js/activity.js
+++ b/activities/QRCode.activity/js/activity.js
@@ -38,7 +38,6 @@ define(["sugar-web/activity/activity","sugar-web/datastore", "sugar-web/env", "w
 		var generateCode = function(text) {
 			qrCode.clear();
 			qrCode.makeCode(text);
-			var text = userText.value.toLowerCase();
 			if (text.indexOf("http://") == 0 || text.indexOf("https://") == 0) {
 				document.getElementById("user-text").classList.add("text-url");
 			} else {
--- a/activities/QRCode.activity/js/activity.js
+++ b/activities/QRCode.activity/js/activity.js
@@ -38,12 +38,13 @@ define(["sugar-web/activity/activity","sugar-web/datastore", "sugar-web/env", "w
 		var generateCode = function(text) {
 			qrCode.clear();
 			qrCode.makeCode(text);
+			addToHistory(text);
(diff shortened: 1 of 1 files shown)


activities/QRCode.activity/js/activity.js
```diff
@@ -38,7 +38,6 @@ define(["sugar-web/activity/activity","sugar-web/datastore", "sugar-web/env", "w
 		var generateCode = function(text) {
 			qrCode.clear();
 			qrCode.makeCode(text);
-			var text = userText.value.toLowerCase();
```
llaske: I think removing the toLowerCase is not the solution else the following line will not be able to detect string like "HTTP://" or "HttPs://". The solution is just to put the raw value in history.

Scar26:
I was actually working on this.

[Code_of_Conduct](https://github.com/llaske/sugarizer/blob/master/CODE_OF_CONDUCT.md)

Anyway, as lionel mentioned, removing the lowercase condition isn't the answer.

ashish0910:
> I was actually working on this.

Sorry @Scar26 if that was the case . However from next time please let other fellow contributors know that you are working on an issue as you did not mention anywhere that you are working on this and I did not found any PR from your side to solve that issue . 
Happy coding

ashish0910:
@llaske it should work now

llaske:
Nice. Thanks.

quozl:
@ashish0910, @Scar26, thanks for your contribution.  Yes, it can be helpful to signal that you are working on something, but that does not stop anyone else from working on the same thing.  Signal is usually not mandatory.  What matters most is the problem is fixed.  The _value decisiveness_ part of the code of conduct.

For trivial fixes, the investment is small, the solution obvious, and immediately accepted.  So the chances of a sense of wasted effort is lower.

For complex fixes, the investment can be large, the solutions diverse, and as a result maintainers (such as @llaske) may pick from more than one person.  Again, the chances of a sense of wasted effort is lower.

In the open source environment, we _expect_ your changes to be copied by others.

Issue #344, opened 2019-04-16
No erase button for recordings in Record activity


Scar26:
Currently there is no way to delete a picture,video or audio recording in the record activity.

![record1](https://user-images.githubusercontent.com/41830515/56235695-8a705a00-60a5-11e9-95ac-6079195abb93.jpg)

llaske:
Why not? But the UI should be simple and usable with a mouse or on a touch screen.

llaske:
Fixed in https://github.com/llaske/sugarizer/pull/345

llaske:
Closed in https://github.com/llaske/sugarizer/commit/dd3c58d56981e4dfa6d94601b95d7117a3bc9241

PR #345, opened 2019-04-16
Added delete button for recordings in record activity

Scar26:
Fixes: #344 

Before:
record1

Now:
record

### DIFF
--- a/activities/Record.activity/css/activity.css
+++ b/activities/Record.activity/css/activity.css
@@ -34,6 +34,14 @@
     height:40%;
 }
 
+.delbtn:hover{
+    border-style:none;
+}
+
+.delbtn:focus{
+    border-style: none;
+}
+
 html, body {
     overflow-y: visible;
 }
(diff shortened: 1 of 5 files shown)


llaske:
This UI will be unusable on tablet, the icon is too small.
Could you suggest another UI?

Scar26:
Would it be appropriate if the button appeared at the corner of the image popup?

Kind of like this:
UI

llaske:
Good idea to put it in the detailed view.
BTW, it's not the standard Sugar icon for delete. Use [this one](https://github.com/llaske/sugarizer/blob/dev/activities/SharedNotes.activity/icons/delete-mode.svg) instead.
Plus, should be white, not red.

Scar26:
Done @llaske 
Now:
record2

llaske:
Nice. 
Please remove the confirm dialog box. I think it's not need. Plus if we want to do a confirm dialog box, the UI is incorrect it should be done with icons for non reading children.

Scar26:
Ok, I have removed the confirm button.

llaske:
Ok nice. Thanks.

PR #353, opened 2019-06-07
[WIP] Write activity 

ashish0910:
Write Activity : Created basic rich text editor with sugarizer UI and Journal support .
Features added : Bold , Italic , Underline , Strikethrough , Lists , Justify alignments , foreground and background colour , Font styling and size , Undo , redo , copy , paste , Journal integration .

Limitations : Multiuser collaboration not added , copy and paste feature not working on Firefox , Tables and images not added .

Screenshot 2019-06-07 at 2 53 30 PM


llaske:
Nice. Look very promising!
Two remarks:
- Colors at startup for Foreground/Background palette should be Black and White
- There is a horizontal/vertical scrollbar on the window at startup.

ashish0910:
@llaske , changes done

ashish0910:
->Images Handling Feature added 
->Export to HTML and TXT feature added

ashish0910:
Features added:
-Export to PDF
-Multi user collaboration using presence

llaske:
Good job.
Few remarks:
- Not sure to understand why but the toolbar of your activity is smaller than the standard one
- Look like the default font is not one of font selectable using the font palette
- The Export palette should be at the end of the toolbar
- Export features should not launch a download, it should generate a new record in the journal with the right mime type (see what happens when you're doing export in Moon activity for example)
- I don't think that Export to HTML button is interesting. I think we need Export to TXT, Export to DOC and Export to PDF
- Regarding the image selection, I think that it should be better if there is no selection for the text part (even no cursor flashing) when an image is selected, else it seems that there is several selected element at the same time
- Could you update your activity.js file with the last one in ActivityTemplate to include dynamic favicon change https://github.com/llaske/sugarizer/commit/4b687e90b665ff9945878e9afcf43d844721d75c
- During collaboration, every change made by one user move the cursor to other users to the start of the file.

llaske:
Cool.
Here is my feedback:
- ENTER no longer work, I have to type CTRL-ENTER to do a line feed
- Cursor is lost when the image is unselected. Today the user should explicitly click somewhere to continue typing after unselection.
- Look like export TXT/PDF generate a new Write record in the Journal. The new journal entry should not be related to the Write activity. Like Moon export entry is not related to Moon activity. Today trying to reopen a PDF open Write and generate an error because of this.
- In collaboration mode, cursors for other players are not updated until something has changed
- Guess it's complicated but it will be nice to see something like a cursor (even not flashing) at the position of other players.

ashish0910:
Bugs Fixed :
1 Enter had stopped working since iframe was removed and div was added . It is fixed now .
2 Image selection restores the cursor to initial position on deselecting the image .
3 Export PDF/TXT issue fixed .
4 Cursors of other users are now updated for all cases 

Features added : 
1 A list of users connected are maintained in the presence palette ( similar to paint activity)
2 Giving undo-redo feature access to new master when original master leaves .
3 Cursors with a name flag are added 

Screenshot 2019-07-08 at 11 22 43 AM

llaske:
Cursors are very cool. 
Here's some other remarks:
- I think you should use both buddy background and foreground color for cursors. Plus, the user name should be bigger.
- The default font in the document should be bigger too: In my opinion at least 24px.
- There is still an issue on cursors: when the remote user create new line before the current cursor position it look like, in the remote user view that the cursor is floating in the middle of nothing.
- Export to PDF don't work: nothing is write in the Journal

ashish0910:
Bugs Fixed :
1. Cursor retention in multi user collaboration improved
2. Default font increased
3.Export to PDF now working 
4. Size of cursor and name of other users increased and Both stroke and fill color included

ashish0910:
Feature added : 
1. Export to docx feature added . Works on all versions of word after 2007 . Does not support google docs .
2.Activity Menu made responsive .

llaske:
Look like insertion of image no longer work. 
I've got an issue when trying to insert one:
```
activity.js:370 Uncaught TypeError: Cannot read property 'toString' of null
    at activity.js:370
    at sugarizer.js:184
    at IDBRequest.request.onsuccess (sugarizer.js:412)
```

ashish0910:
Closing this PR and Shifted to #364 Due merge issues

Issue #354, opened 2019-06-07
Add a pause button to the Physics activity


llaske:
Would be nice to add a pause button on the Physics activity to stop current movement of items. 
The button could be transformed into a Play button to continue movement.

Source code for Physics activity is here: https://github.com/llaske/sugarizer/tree/dev/activities/PhysicsJS.activity 

(Suggestion by [Sameer Verma](https://twitter.com/sameerverma/status/1132768962511495168?s=20))

phorical:
Opened a pull request to fulfill this request: #355 .

llaske:
Fixed in https://github.com/llaske/sugarizer/pull/355

llaske:
Closed in https://github.com/llaske/sugarizer/commit/dd3c58d56981e4dfa6d94601b95d7117a3bc9241

PR #355, opened 2019-06-08
Feature: Physics pause button

phorical:
This is to address Issue #354. This adds a pause button to the PhysicsJS activity. Bodies are still visible during pause but set to static, and new bodies can be added as well. They become dynamic bodies on unpause.

### DIFF
--- a/activities/PhysicsJS.activity/css/activity.css
+++ b/activities/PhysicsJS.activity/css/activity.css
@@ -43,9 +43,13 @@
   z-index: 3;
 }
 
-#main-toolbar #pause-button {
-  background-image: url(../icons/pause-play.svg);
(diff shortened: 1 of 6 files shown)


llaske:
Very good. Thanks for your contribution.
BTW, it will be better to have 2 different icons (play and pause) for the play/pause button in the toolbar and switch between play and pause icon depending of the state.
See Color my world in Interactive mode to see an example.

phorical:
@llaske I've added a new commit to have 2 different icons (play and pause) for the play/pause button in the toolbar. The activity switches between the play and pause icons depending on the state.

llaske:
Good job. Thanks.

PR #357, opened 2019-07-02
Added Support for PDF and TXT

ashish0910:
Support added for PDF and TXT files in journal entries .
According to this PR :
For TXT : it will receive text data in form of string from user and will create txt file .
For PDF : It will receive data in form of base64 dataurl (with file type PDF ) and produces PDF file .


llaske:
Miss the Electron part.
Should be updated also [here](https://github.com/llaske/sugarizer/blob/dev/main.js#L75) and [here](https://github.com/llaske/sugarizer/blob/dev/main.js#L166).

ashish0910:
@llaske Done !

PR #364, opened 2019-07-17
Write Activity

ashish0910:
Write Activity Project GSOC - 2019

**Complete Project Summary and Final Report** : [here](https://medium.com/@ashish09101998/gsoc-final-blog-f3570fe0f3fa)
 
Link to Blog Series : [here](https://medium.com/@ashish09101998)

Milestones achieved:

- Basic Editor Set up with major text editing features
- Journal Integration 
- Images Integration
- Multi user Collaboration
- Export to Doc , TXT , PDF , ODT
- Localisation


ashish0910:
Features Till now : 
- [x] Rich text editor with Quilljs having Bold , italic , underline , strikethrough , Lists , paragraph align, Foreground and background colour , undo , redo .
- [x] Images : resize and text wrapping 
- [x] Journal connectivity
- [x] Multi user collaboration
- [x] Export to doc , txt , PDF
- [x] Responsive : Works across different screen size
- [ ] copy-paste feature 
- [ ] Tables Integration 
- [ ] Speak integration

llaske:
I've tested DOCX export.
The format is not supported by LibreOffice thought it works into Word.
May be it could interesting to convert to DOC instead?
Plus it look like there is some issue in encoding character with accent (like é,à,è, ...).

activities/Write.activity/js/activity.js
```diff
@@ -349,20 +350,14 @@ define([
...
-			var content = '<!DOCTYPE html>' + document.getElementById("editor").innerHTML;
-			var orientation = "portrait";
-			var converted = htmlDocx.asBlob(content, {orientation: orientation});
-			var reader = new FileReader();
-			reader.readAsDataURL(converted); 
-			reader.onloadend = function() {
-				var inputData = reader.result;
-				var mimetype = 'application/vnd.openxmlformats-officedocument.wordprocessingml.document';                
+			$("#editor").wordExport(function(inputData){
+				var mimetype = 'application/vnd.ms-word;charset=utf-8';
```
llaske: Guess the mimetype for Word files should be `application/msword`. See my comment on https://github.com/llaske/sugarizer/pull/367

activities/Write.activity/js/activity.js
```diff
@@ -355,7 +355,7 @@ define([
         	var footer = "</body></html>";
 			var sourceHTML = header+content+footer;
 			var inputData = 'data:application/vnd.ms-word;charset=utf-8;base64,' + btoa(unescape(encodeURIComponent( sourceHTML )));
-			var mimetype = 'application/vnd.ms-word;charset=utf-8';
+			var mimetype = 'application/vnd.ms-word;';
```
llaske: That's still not good, should be `application/msword`.

activities/Write.activity/js/activity.js
```diff
@@ -355,7 +355,7 @@ define([
         	var footer = "</body></html>";
 			var sourceHTML = header+content+footer;
 			var inputData = 'data:application/vnd.ms-word;charset=utf-8;base64,' + btoa(unescape(encodeURIComponent( sourceHTML )));
-			var mimetype = 'application/vnd.ms-word;';
+			var mimetype = 'application/ms-word;';
```
llaske: Still not good :laughing: it's `application/msword`

ashish0910: Sorry for the typo 😅

llaske:
Very good job on ODT generation :+1: 
Three remarks:
- Font changing in ODT seems not work, is it a limitation?
- Do you update the DOC generation to avoid using base64 encoding for images?
- Could you change the two icons for ODT/DOC generation: use the same image than TXT/PDF generation with ODT and DOC instead of TXT/PDF.

ashish0910:
- Regarding the font family change in ODT , I will fix it .
- Regarding the doc , It is not the encoding we are doing at the end causing issue , It is because when we add images through journal chooser , we get them in a base64 format which are not supported . Is there a way to avoid that ?
- Regarding icons , do you want to replace the odt and doc icon with icons similar to txt and PDF ? ( I was not able to find similar one's for ODT and DOC ) .

llaske:
- Regarding encoding, the problem is that only recent versions of Word support base64 encoding so it will be better to use binary encoding instead. It's not possible in Journal Chooser to encode in something else than base64 because it's returned as a string. BTW you could try to re-encode it as a Blob object then create an URL that you could affect to the `src` property for the image. You could find an example [here](https://github.com/llaske/sugarizer/blob/dev/lib/util.js#L647).
- Regarding icons, the icon don't exist. You could obtain it by editing the SVG file in InkScape to update the text.

llaske:
Just tested. The level of ODT support is now very good :+1:

ashish0910:
Copy and Paste Issue has been Fixed

llaske:
Few remarks:
- Why you remove quill.min.js instead of quill.js? I guess the first is minimized so it should be better to use this one
- Will be better to put all libraries in the lib directory and put code for the activity in the js directory
- The standard indentation rules for Sugarizer is to use TAB, currently you seem to use a mix of TAB and SPACE. Could you fix it?
- Could you change the welcome text to the one in the screen capture below? The image could be found [here](https://github.com/llaske/sugarizer/blob/dev/activities/Abecedarium.activity/images/class.png). Note that the text should be localized.

image

ashish0910:
Done Changes

activities/Write.activity/js/activity.js
```diff
@@ -320,7 +320,8 @@ define([
 			file_size: 0
 		};
 		datastore.create(metadata, function() {
-			console.log("export done.")
+			console.log("export done.");
+			humane.log("Export to TXT Done");
```
llaske: Message should be localized.

activities/Write.activity/js/activity.js
```diff
@@ -361,7 +362,8 @@ define([
                         file_size: 0
                     };
                     datastore.create(metadata, function() {
-                        console.log("export done.")
+						console.log("export done.");
+						humane.log("Export to PDF Done");
```
llaske: Message should be localized.

activities/Write.activity/js/activity.js
```diff
@@ -389,6 +391,7 @@ define([
 			};
 			datastore.create(metadata, function() {
 				console.log("export done.");
+				humane.log("Export to DOC Done");
```
llaske: Message should be localized.

activities/Write.activity/js/activity.js
```diff
@@ -410,6 +413,7 @@ define([
 			};
 			datastore.create(metadata, function() {
 				console.log("export done.");
+				humane.log("Export to ODT Done");
```
llaske: Message should be localized.

activities/Write.activity/js/activity.js
```diff
@@ -43,19 +43,36 @@ define([
...
-					{ insert:  greet, attributes: { size: "40px" , color : environment.user.colorvalue.stroke , bold: true }},
-					{ insert: name, },
-					{ insert: '\n' }
-				]);
-				var index = greet.length + name.length ;
-				editor.setSelection(index,0);				
+				console.log(window.initialImageDataUrl);
+				// Intentionally added setTimeout to allow locale.ini file to be loaded
+				setTimeout(function(){
+					editor.setContents([
```
llaske: Unfortunately, the way of building a sentence is not the same in all languages. So you can't simply use a list of inserts and be sure that it could work in any language to rebuild the sentence. You can only insert sentence by sentence, not word by word.
So I suggest two things:
- For the first sentence, use a webL10n variable into the .INI string for the username, something like `Welcome {{username}}!`. You should pass the username value to the webL10n get call.
- For other sentences, you should use a substitution into the .INI string, something like `Write is an editor where you could [b]type[/b] and [i]format[/i] your [u]text[/u] and ideas.` Then call the webL10n get function and split the result to add specific QuillJS attributes on each part resulting of the sentence.

llaske:
@ashish0910  It doesn't solve the issue of welcome text construction.
As I said previously, the user name should be pass as parameter in first sentence.
For other sentence, I don't see other way of doing the job to use tag into each sentence to isolate part of sentence to format.

ashish0910:
I have added some fixes for Localisation

llaske:
A very smart way to solve the issue :+1: 
I love the method name `dangerouslyPasteHTML` :smile: 
Hope it's not too dangerous!

PR #367, opened 2019-07-27
Doc and ODT Support

ashish0910:
Adds Support for .doc Files . 

PS : Please Don't remove [this](https://github.com/llaske/sugarizer/pull/367/files#diff-3deb3f32958bb937ae05c6f3e4abbdf5R458) line . It is required for txt support .


llaske:
Do we really need to support both .DOCX and .DOC? If the Write activity export only .DOC, I guess we could remove the .DOCX support.

llaske:
Not sure why you use mimetype `application/vnd.ms-word;charset=utf-8`
Did you try to import/export Word files into Journal with this type?
On my side, it works only with `application/msword`.
I've also create a dedicated icon for ODT files in Journal.

PR #373, opened 2019-09-10
fixing clear icon and save colors buttons 

### LINKED ISSUE #372: Clear icon in XO Editor activity is not consistent
The icon use to clear the colors position in XO Editor is not consistent.
It should be the clear icon like in Paint, Physics and other activities.
Plus it should be align to the left.

Capture du 2019-09-02 22-15-40

prakashujjwal1010:
This PR fixes #372 
Screenshot from 2019-09-10 13-06-03

### DIFF
--- a/activities/XOEditor.activity/css/activity.css
+++ b/activities/XOEditor.activity/css/activity.css
@@ -8,7 +8,7 @@
 }
 
 #reset-button {
-  background-image: url(../icons/clear-button.svg);
+  background-image: url(../icons/reset-clear.svg);
 }
 
 #canvas {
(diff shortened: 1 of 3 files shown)


llaske:
Nice. But I think it will be better to switch buttons: save to the left and erase to the right. Will be more consistent with other activities.

prakashujjwal1010:
@llaske done. just wanna ask how xo editor works means how to play it. how's about if each activity have a small description(so that a child can get it) about it? or the developers intentionally didn't provide any description about activities so that children can learn by themselves.

llaske:
Nice. Thanks.

PR #377, opened 2019-09-18
Fix title overlaying issue in Tank Activity

utkarsh-raj:
The title div was covering the button which was still accessible from top.


sukhdeepg:
@utkarsh-raj I think just removing the <p> tag with the title class would be sufficient because the title is anyways not responsive on mobile and some activities that I check don't have title in the first place.
tank

Issue #380, opened 2019-09-28
Allow to control speed generation in Game of Life activity


llaske:
Add icons to control speed of generation in the game of life activity.
It could be for example a fast forward and fast backward button to a double and divide by 2 the current speed.

prakashujjwal1010:
@llaske , I would like to work upon this issue.
I am thinking to add two toolbar buttons fast backward and fast forward. and I think i should avoid storing the speed value in datastore .

llaske:
Sounds good. Do not hesitate.
You're right, you don't need to store the speed value in the datastore. BTW I think you should display it when it will change.

Note that there is no need to ask permission to work on an issue. See [here](https://github.com/llaske/sugarizer/blob/master/CONTRIBUTING.md).

llaske:
Nice start. By the way I'm not a big fan to have one button only. It means that to slow the generation you should first accelerate it. It's counter intuitive. 
Alternatively may be we could use a slider instead of buttons.
Something like the one used to control rate in the Speak activity.

<img width="244" alt="image" src="https://user-images.githubusercontent.com/981841/66136147-4513fa80-e5fb-11e9-9077-700c9b87d8e6.png">

It's used here: https://bitstorm.org/gameoflife/

llaske:
Fixed in https://github.com/llaske/sugarizer/commit/bfe50cb516df1b85492265bd43f53e7d61d98bb9

PR #385, opened 2019-09-30
adding generation speed button

prakashujjwal1010:
This pr adds a feature in game of life activity to control generation speed( #380 ).
I have made a single toolbar button which shows the current generation speed as well and a user will have a set of allowed speed values (i.e 0.25x, 0.5x, 1x, 2x, 4x) and will have to click that button to switch between different speed values. 

Screenshot from 2019-09-30 23-09-16

Screenshot from 2019-09-30 23-09-29

Screenshot from 2019-09-30 23-09-38

Screenshot from 2019-09-30 23-09-40

Screenshot from 2019-09-30 23-09-46


llaske:
Good job. See my comment here: https://github.com/llaske/sugarizer/issues/380#issuecomment-537972811

prakashujjwal1010:
The single generation speed is replaced by a pallete.
Screenshot from 2019-10-05 16-12-24
Screenshot from 2019-10-05 15-49-45

llaske:
Very nice.
Few remarks:
- The icon is very cool. Could you confirm it use an Open Source license (or Creative Common)?
- The palette don't close when the user click a second time on it. It stay open forever.
- When the speed is low there is a strange "jump" in generation (see video capture below)

golissue

prakashujjwal1010:
after introducing a speed palette in game of life , I checked its working in mozilla firefox and it was working good there was no jumps etc

ezgif com-video-to-gif

But when i checked it in chrome I noticed that the speed palette is not working that is palette don't close user  clicked on it second time and after changing the speed value by the range input there was jumps as speed keeps oscillating between its initial value and new value, as @llaske noticed.

ezgif com-video-to-gif (1)

I add the testpalette(just like in [docs](https://github.com/llaske/sugarizer/blob/master/docs/tutorial_step8.md) in game of life and found that even this test palette got stuck i.e. don't close by clicking second time in google chrome.
But then i add the same testpalette in xor activity but here the test palette is working fine in google chrome.

ezgif com-video-to-gif (2)

i even  tried to add different types of palettes in game of life but all of them were getting struck.

prakashujjwal1010:
i downloaded the speed icon from [here](https://www.svgrepo.com/svg/28222/speedometer) and I think we can use it .

llaske:
I've worked on your issues and it appears that it's not related to your work.
In fact, it's related to a tricky issue related to the way the `main` function is called in the activity. The `main` was called in the `localized` event but this event could be called double times depending of the browser. This double call generated double board generation at the same time and two events listeners on your palette.
I've declared a specific issue on it and fix it here: https://github.com/llaske/sugarizer/issues/388
Retrieve this fix and resolve conflicts and it should work on your side too.

Thanks for the precision on the speed icon.

llaske:
Nice job.
Few things:

- Regarding speed: initial speed should be slower. The minimum speed should be very slow (may be one generation by second). The higher speed is too fast. Note that you don't need to have a linear speed.
- Could you localize the "speed" label in the palette?
- Remove commented code in speedpalette

llaske:
Very good work. Thanks.

Issue #399, opened 2019-11-08
Add a fullscreen button in the Reflection activity


llaske:
Sugarizer activities don't look good on devices with small screen - for example smartphones.
On these devices, the toolbar could sometimes take 1/4 of the screen.

![image](https://user-images.githubusercontent.com/981841/68498330-e1629a00-0256-11ea-8cb7-50e522b65d7f.png)

To increase the size for activities contents, the idea is to add a fullscreen button in the toolbar, like in the QRCode activity.

![image](https://user-images.githubusercontent.com/981841/68498506-4b7b3f00-0257-11ea-8507-b56176b061b2.png)

When clicked, the fullscreen button hide the toolbar and display only an unfullscreen button.

![image](https://user-images.githubusercontent.com/981841/68498790-086d9b80-0258-11ea-9063-cc3630d132cd.png)

When clicked, the unfullscreen button show again the toolbar.

This features is implemented [here](https://github.com/llaske/sugarizer/blob/dev/activities/QRCode.activity/js/activity.js#L144-L160).

The objective of this issue is to implement the same feature in the Reflection activity.

Thinks to check:
* [ ] The fullscreen button should be on the right of the toolbar, near the Stop button.
* [ ] The fullscreen button should be always visible, including in small screen size. Could require to adapt media queries in CSS.
* [ ] The size of the board should be increase in fullscreen mode. So it means that the resize event should be call directly or indirectly when clicking on fullscreen/unfullscreen button.
* [ ] The unfullscreen button should be visible only when the screen is in fullscreen mode.
* [ ] The unfullscreen button should be attach to the right of the screen. It's better if it's not at the same location than the Stop button to avoid side effect on multiple click.

est271:
I would like to work on this issue but i have one question before i begin. 

I cloned the dev branch and I am able to successfully launch the platform on the sugarizer-server and all the features on sugarizer seem to work just fine. 

the only thing is that there is some unit test failures on my dev branch without any modifications being made. Currently running on Linux Mint and I'm wondering if its due to my system and settings or maybe its other issues that are still being worked on? Attached is a screenshot 

![u_test1](https://user-images.githubusercontent.com/56150991/68632408-427ead80-04a3-11ea-8b91-72e702c077a5.png)

Most of the other failures have similar Error messages

est271:
Actually never mind about my previous question. 
I get the same results from the [Sugarizer Web Application](https://try.sugarizer.org),
[...]

PR #420, opened 2019-11-12
Add fs button

est271:
- [x] closes issue #399 
- [x] The fullscreen button should be on the right of the toolbar, near the Stop button.
- [x] The fullscreen button should be always visible, including in small screen size. Could require to adapt media queries in CSS.
- [x] The size of the board should be increase in fullscreen mode. So it means that the resize event should be call directly or indirectly when clicking on fullscreen/unfullscreen button.
- [x] The unfullscreen button should be visible only when the screen is in fullscreen mode.
- [x] The unfullscreen button should be attach to the right of the screen. It's better if it's not at the same location than the Stop button to avoid side effect on multiple click.


est271:
Screenshot of Fullscreen

fullscreen

llaske:
Good job.

But you should locate the Unfullscreen button from the right size instead of using the left size (guess `right: 5%` instead of `left: 95%`).
Else the unfullscreen button could be truncated on small screens:

image

est271:
@llaske good catch. will fix soon

llaske:
Nice. Thanks.

Issue #411, opened 2019-11-09
Add a fullscreen button in Gears activity


llaske:
The idea is to add a fullscreen button in the toolbar to increase the size for activities contents on small devices.

The complete process is describe in #399

llaske:
Fixed in https://github.com/llaske/sugarizer/commit/913a878ccdd3c9e5da81266128c58db83e8eaafa

PR #449, opened 2019-11-18
Solved issue #411

sarthak-g:
Added fullscreen and unfullscreen button in Gears Activity as mentioned in issue #411 

gear1

gear2


llaske:
That's the idea.
But there is an error in the console when clicking on fullscreen/unfullscreen button:

```
activity.js:126 Uncaught ReferenceError: resizeHandler is not defined
    at HTMLButtonElement.<anonymous> (activity.js:126)
```

sarthak-g:
I've updated the code and made a commit.

llaske:
There is another issue: in fullscreen mode, the drawing is not aligned with the cursor. There is vertical shift to the size of the toolbar.

image

sarthak-g:
If vertical shift will not be done then blank space will display as below:
VerticalShiftNotDone
@canvasOffsetY in gearsketch_main.coffee gets value only when the page is displayed for first time (with toolbar), so when fullscreen is done then size of the canvas is not updated.
 What I think is we've to call @updateCanvasSize() when fullscreen as well as unfullscreen button is pressed but for doing so we need jquery.

llaske:
Of course the vertical shift should be done but it should be take into account when mouse position is computed.
You're right, I guess the issue is related to the canvasOffsetY initialized [here](https://github.com/llaske/sugarizer/blob/dev/activities/Gears.activity/lib/gearsketch/gearsketch_main.js#L92). This value should be updated each time there is a click on fullscreen/unfullscreen button. 
Calling updateCanvasSize method of the gearSketch object, from the activity, could be a solution. I don't think you need jQuery since the gearSketch object is available from the activity.

sarthak-g:
Fullscreen is working as expected.

activities/Gears.activity/js/activity.js
```diff
@@ -117,6 +117,22 @@ define(["sugar-web/activity/activity","sugar-web/graphics/radiobuttonsgroup","ge
             gearSketch.playDemo();
         });
 
+        // Full screen
+		document.getElementById("fullscreen-button").addEventListener('click', function() {
```
llaske: Use space instead of TAB as identation

activities/Gears.activity/js/activity.js
```diff
@@ -117,6 +117,22 @@ define(["sugar-web/activity/activity","sugar-web/graphics/radiobuttonsgroup","ge
...
+			document.getElementById("main-toolbar").style.opacity = 0;
```
llaske: Use space instead of TAB as identation

activities/Gears.activity/js/activity.js
```diff
@@ -117,6 +117,22 @@ define(["sugar-web/activity/activity","sugar-web/graphics/radiobuttonsgroup","ge
...
+            document.getElementById("canvas").style.top = "0px";
+            gearSketch.canvasOffsetY = gearSketch.canvas.getBoundingClientRect().top;
+            gearSketch.updateCanvasSize();
+            document.getElementById("unfullscreen-button").style.visibility = "visible";
+		});
+		document.getElementById("unfullscreen-button").addEventListener('click', function() {
```
llaske: Use space instead of TAB as identation

activities/Gears.activity/js/activity.js
```diff
@@ -117,6 +117,22 @@ define(["sugar-web/activity/activity","sugar-web/graphics/radiobuttonsgroup","ge
...
+			document.getElementById("main-toolbar").style.opacity = 1;
```
llaske: Use space instead of TAB as identation

sarthak-g:
Indentation issue resolved.

llaske:
Sorry, there is still a TAB instead of white spaces before lines `document.getElementById("fullscreen-button")`,  `});` and  `document.getElementById("unfullscreen-button")`

See the screen capture of Atom editor below:

image

sarthak-g:
It is fine now I guess.
indent show spaces

llaske:
Cool. That's fine now.
Thanks.

Issue #412, opened 2019-11-09
Add a fullscreen button in Clock activity


llaske:
The idea is to add a fullscreen button in the toolbar to increase the size for activity content on small devices.

The complete process is describe in #399

llaske:
Fixed in https://github.com/llaske/sugarizer/commit/08295319934d8c48c8c249cd68240af66903bf56

PR #456, opened 2019-11-25
Solved issue #412

sarthak-g:
Added fullscreen option in clock activity as mentioned in #412 
clockactivity1
clockactivity2


llaske:
Nice start, the toolbar is hidden/shown correctly but the window is not resized.
You must resize the window to increase the board size.

llaske:
Fixed by https://github.com/llaske/sugarizer/commit/08295319934d8c48c8c249cd68240af66903bf56

Issue #415, opened 2019-11-09
Add a fullscreen button in XO Editor activity


llaske:
The idea is to add a fullscreen button in the toolbar to increase the size for activity content on small devices.

The complete process is describe in #399

llaske:
Fixed in https://github.com/llaske/sugarizer/commit/c41dd12806765f50ebcfabe55defa7a194b1e7a4

PR #457, opened 2019-11-25
Solved issue #415

sarthak-g:
Added fullscreen option to XO Editor as mentioned in #415 
xo1
xo2


llaske:
The toolbar is hidden/shown correctly but the window is not resized.
You must resize the window to increase the board size so you will avoid to have a white gap at the bottom of the window.

sarthak-g:
It's fine now I guess.
xoresize1
xoresize22

llaske:
It don't works. It just flash between fullscreen and unfullscreen. See below.

Peek 26-11-2019 21-38

sarthak-g:
As canvas height is changed so I've created new stage and corresponding new editor.
ezgif com-video-to-gif

llaske:
Why not just calling [`ResizeCanvas()`](https://github.com/llaske/sugarizer/blob/dev/activities/XOEditor.activity/js/activity.js#L39) method?

sarthak-g:
When calling resizeCanvas() method, it just flash between fullscreen and unfullscreen.

llaske:
The problem is that if you recreate a new stage/editor, you lost all change to colors position.
See below:

sarthak-g:
Yeah, I felt same.
Is there any better way you can suggest?

llaske:
Hmmm. The current implementation of resize is not ideal. Calling `location.reload()` is not at all a good idea 😞  BTW there is a work around to fix your current issue: call `e.stop()` before creating a new editor, it will force the board to be saved into the datastore so the `location.reload()` will not lost all dots position.
It seems to work on my side. Should be more tested thought.

sarthak-g:
Issue fixed

llaske:
Good job. Thanks.

Issue #401, opened 2019-11-09
Add a fullscreen button in Memorize activity


llaske:
The idea is to add a fullscreen button in the toolbar to increase the size for activity content on small devices.

The complete process is describe in https://github.com/llaske/sugarizer/issues/399

llaske:
Fixed by https://github.com/llaske/sugarizer/pull/460

PR #460, opened 2019-12-02
add fullscreen button to memorize activity

pidddgy:
#401 

There's two buttons, zoom in and zoom out. When the user clicks the zoom in button, we adjust the `zoom` property of the game grid such that it fills up the screen (leaving a buffer), hides the toolbar, and shows the zoom out button. zoom out button shows the zoom in button, shows the toolbar, and sets the game grid back to its original size

https://youtu.be/3B4U3iX5y24


FreddieN:
Great start! You have accomplished the desired goal of making the top navigation disappear and the items on the page do partially enlarge. 

The task was to improve the experience for mobile devices and I've had a look in chrome developer tools at your solution and it works however the content overflows. Is there any way you could adjust the sizing of the blocks to use the page space more efficiently? Currently there is quite a lot of blank space on the user page on large displays as well.

Let me know if you need any guidance...

Video of mobile overflow: https://youtu.be/F_-Jk6uoUp0

FreddieN:
Also another note: in terms of your GitHub issue it might be worth explaining what process you underwent to achieve your goal ie what code needed changing for contributors in the future if they ever need to look back. Your commit messages as well could be a bit more descriptive :)

pidddgy:
hopefully better?

i've also added a description in my PR that briefly explains what I did

FreddieN:
Yes as you say, I think reducing the buffer (margin) size around the blocks to make them appear bigger with less space between them would be a better idea but well done on managing to make it fit!

Thanks,
Freddie

llaske:
It's not good to me. In fullscreen mode there is a margin at the top of the board so the game is not playable without scrolling. See screen capture.



Plus, I suggest you to uncomment lines https://github.com/llaske/sugarizer/blob/dev/activities/Memorize.activity/js/memorize-app.js#L917-L925 to better handling resize.

pidddgy:
the bug that llaske mentioned was due to me not setting the toolbar display to none, instead it was the visibility. I also made the margin on the blocks 3 px instead of 5 px to make the margin around the blocks smaller as freddie said. is this better?

main.js
```diff
@@ -15,7 +15,7 @@ var dialog = electron.dialog;
 
 var mainWindow = null;
 
-var debug = false;
+var debug = true;
```
llaske: Guess you've done it to debug but should be removed from the PR.

pidddgy: think i already pushed a commit to fix this https://github.com/llaske/sugarizer/pull/460/commits/ab5e5e5fc08ce39bb362ca4cb11e1a996759e26f

llaske: Oh sorry, you're right

Issue #384, opened 2019-09-28
Image insertion don't work in shared mode in Paint activity


llaske:
Image insertion don't work in shared mode.
Image move and image resize is not recorded and pen trace is display instead in other activities.
See result of image insertion below:

![image](https://user-images.githubusercontent.com/981841/65822120-3a461800-e23f-11e9-9a3a-eaa15518b504.png)

llaske:
Fixed by https://github.com/llaske/sugarizer/commit/9eb3f4bde6a9973695323271b35ee1ccce5cf58d

PR #461, opened 2019-12-03
Fix Issue #384

AndreaGon:
Image insertion now works properly when shared. This should fix issue #384 

Fix384


FreddieN:
However you may notice that there is still the pen trace left by the user clicking but this is then covered up by the image so I do not think this has a major effect however it would be a bonus if you were able to fix this issue.

llaske:
Very good jov @AndreaGon .

>  there is still the pen trace left I do not think this has a major effect 

You're right @FreddieN, I think we could ignore it.

PR #462, opened 2019-12-04
Fix Issue #398

### LINKED ISSUE #398: Close the presence palette when the activity is shared
When an user want to share an activity, he click on the presence palette, then on the neighborhood icon. 

image

In some activity, when the icon is clicked, the palette automatically close. For example in Maze activity. On some other, the palette remain open which is not pertinent.

The idea is to generalize this feature by calling `popDown` method like [here](https://github.com/llaske/sugarizer/blob/master/activities/MazeWeb.activity/js/activity.js#L886).

So the following activities should be fixed:

* [ ] Chat activity
* [ ] Memorize activity
* [ ] Paint activity

AndreaGon:
Added presencepalette.popdown() to close presence palette when activity is shared. Hopefully fixes #398.

- [x] Chat Activity

- [x] Memorize Activity

- [x] Paint Activity

### DIFF
--- a/activities/ChatPrototype.activity/js/activity.js
+++ b/activities/ChatPrototype.activity/js/activity.js
@@ -45,7 +45,9 @@ define(["sugar-web/activity/activity","webL10n","sugar-web/graphics/palette","su
 		// Connect to network
         var presenceObject;
 		function shareActivity() {
+      			presencepalette.popDown();
 			presenceObject = activity.getPresenceObject(function (error, presence) {
+
 				// Unable to join
 				if (error)  {
 					socketStatus.innerHTML = l10n_s.get('Error');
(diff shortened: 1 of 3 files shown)


activities/ChatPrototype.activity/js/activity.js
```diff
@@ -45,7 +45,9 @@ define(["sugar-web/activity/activity","webL10n","sugar-web/graphics/palette","su
 		// Connect to network
         var presenceObject;
 		function shareActivity() {
+      			presencepalette.popDown();
```
FreddieN: You have changed the indentation of this line compared to the file. Could you make sure that it is all in line?

activities/Memorize.activity/js/activity.js
```diff
@@ -81,6 +81,7 @@ function initPresence(activity, memorizeApp, presencepalette, callback) {
             presencePalette.setShared(true);
         } else {
             presencePalette.addEventListener('shared', function () {
+              presencePalette.popDown();
```
FreddieN: This needs to move over similar to above so it is in line with shareActivity.

AndreaGon:
@FreddieN Is this how it should be?

llaske:
Good job. Thanks.

PR #463, opened 2019-12-04
open dev tools shortcut when developer presses f12

pidddgy:
just added a submenu to `Menu` if debug is true that calls `mainWindow.webContents.openDevTools();`

### DIFF
--- a/main.js
+++ b/main.js
@@ -205,18 +205,18 @@ function createWindow () {
 						app.quit()
 					}
 				}]
-            };
-            menu.submenu[0].label = l10n.get("Quit");
-            
-            if (debug) {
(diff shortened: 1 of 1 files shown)


llaske:
When the `--debug` option is used, dev tools are open at startup: https://github.com/llaske/sugarizer/blob/dev/main.js#L217
Not sure to understand the interest to add a shortcut to launch it.
What is your idea?

pidddgy:
In the past sometimes I've wanted to close and reopen devtools since I'm working on a small display and it would be pretty frustrating to have to restart the whole app just to get devtools open again.

llaske:
Not pertinent.

Issue #416, opened 2019-11-09
Add a fullscreen button in BlockRain activity


llaske:
The idea is to add a fullscreen button in the toolbar to increase the size for activity content on small devices.

The complete process is describe in #399

llaske:
Fixed in https://github.com/llaske/sugarizer/commit/22f4d6bb0e49e07c687af8efa5a6c33136413601

PR #465, opened 2019-12-05
Fix Issue #416

AndreaGon:
Blockrain Activity can now be set to fullscreen mode.

Fix416

Small screen mode:

Mobile

I have to put the unfullscreen button at the bottom of the activity when screen size is small as it overlaps with the activity container. Any thoughts?


activities/Blockrain.activity/css/activity.css
```diff
@@ -1,204 +1,51 @@
...
+</style>
+</head>
+
+<body>
+<div id="main-toolbar" class="toolbar">
+  <button class="toolbutton" id="activity-button" title="My Activity"></button>
+	<button class="toolbutton" id="btn-next" title="Change theme"></button>
+  <button class="toolbutton pull-right" id="stop-button" title="Stop"></button>
+	<button class="toolbutton pull-right" id="fullscreen-button" title="Fullscreen"></button>
+</div>
```
FreddieN: I think there is a problem here with displaying the fullscreen button properly.

FreddieN:
See here: https://i.imgur.com/q1UKRv8.mp4

AndreaGon:
I found the problem: I accidentally put the HTML data inside the activity.css file 😆 Can you please check if it's working now?

FreddieN:
Thanks, I should have spotted that! It works perfectly now! :D

llaske:
There is an issue when doing fullscreen/unfullscreen. See capture below:

Peek 05-12-2019 21-46

I wonder also why this [line](https://github.com/llaske/sugarizer/blob/dev/activities/Blockrain.activity/js/blockrain.jquery.src.js#L235) is commented and if it's not a good idea to call the` updateSizes()` method when doing fullscreen/unfullscreen.

AndreaGon:
I think I've fixed the issue in this new commit.

@llaske to answer your questions:

1. The `game.updateSizes();` line is already commented when I've started working on the project. When I tried removing the comment (to see what it does), the generated background disappears and the blocks float when the screen is resized.

2. I did not call `updateSizes()` as it has the same effect as stated above when the screen is resized.

Picture for reference:
reason

llaske:
Perfectly works now. Thanks.

llaske:
You've done a PR on the master branch instead of the dev branch :disappointed: 
I had to revert it manually. Could you recreate a PR with this fix on the dev branch.

AndreaGon:
Oh no, I'm so sorry :sad:. I will create a PR in dev branch.

This won't happen again.

AndreaGon:
Oh no, I'm so sorry. I will create a PR in dev branch.

This won't happen again.

PR #471, opened 2019-12-06
Added tutorial for Memorize activity (works within editor mode as well)

Uasked4dat:


llaske:
Good start.
Few remarks:
- You must not do a PR on the master branch, it should be on the dev branch
- Will be nice to add a popup to explain the the user should click on tile
- Will be nice to add popup in Edit mode for elements: Add, Update, Remove, Input field and Insert Image button

llaske:
Nice.

Few remarks:
- Depending of the type of the game, there is no always a set of "1" cards and a set of "2" cards. Select for example the sound game in the palette.

image

- It will be better to do steps to explain toolbar button before steps to explain the board
- The step "End of tutorial" is useless
- In Edit mode, it will be nice to add a step to explain the button to go back to the Play mode.

llaske:
That's better.

By the way, could you refresh your branch because on my side the fullscreen button (coming from another PR) appears twice when I merge your branch.

image

Please update also the [es] part of the locale.ini to integrate new strings. You could just copy the [en] values.

Issue #402, opened 2019-11-09
Add a fullscreen button in Abacus activity


llaske:
The idea is to add a fullscreen button in the toolbar to increase the size for activity content on small devices.

The complete process is describe in https://github.com/llaske/sugarizer/issues/399

NathanDimmer:
Working on it!

llaske:
Fixed in https://github.com/llaske/sugarizer/pull/478

PR #478, opened 2019-12-07
Fullscreen abacus

NathanDimmer:
Add a fullscreen button to the Abacus activity.

Fixes issue #402 

@FreddieN @eohomegrownapps @utkarsh-raj @llaske @ashish0910


llaske:
It's a good start.

Two remarks:

- Only lines added/updated should be changed in your files. Here it seems that your editor (or JSLint) has changed more lines than needed.

- You call resizeCanvas() method and it's a good thing but the resizeCanvas() method consider that the toolbar is always present even in fullscreen mode. The number 55 in [this line](https://github.com/llaske/sugarizer/blob/dev/activities/Abacus.activity/js/activity.js#L48) is the size of the toolbar. So currently, even in fullscreen mode, the size of the board is still the same.

NathanDimmer:
Thank you! I disabled JSLint and reset the file, and added the fix to resizeCanvas(). Is there anything else that needs to be done?

llaske:
Good job. Thanks.

Issue #418, opened 2019-11-09
Add a fullscreen button in LastOneLose activity


llaske:
The idea is to add a fullscreen button in the toolbar to increase the size for activity content on small devices.

The complete process is describe in #399

sarthak-g:
Yeah, I also feel the same.Can I work on this?

pidddgy:
is this really necessary? space already seems to be used pretty well.

<img width="678" alt="image" src="https://user-images.githubusercontent.com/30125762/70396641-ceb7be00-19d8-11ea-87eb-cb62826e3d61.png">

llaske:
See my comment [here](https://github.com/llaske/sugarizer/pull/479#issuecomment-562862247 ), you will see a case where just hiding the toolbar without resize give a bad result.

llaske:
Fixed in https://github.com/llaske/sugarizer/commit/5e2a5035fa63ec7f3861f723d14517c135794208

PR #479, opened 2019-12-07
Fixed Sugarizer issue #418

jontymagan:
Added fullscreen and unfullscreen button in Last One Lose activity.
Added icons
Updated activity.css
Updated activity.js
Updated index.html


activities/LastOneLoses.activity/js/activity.js
```diff
@@ -483,5 +483,15 @@ define(["sugar-web/activity/activity", "sugar-web/graphics/radiobuttonsgroup", "
 			}
 		},
 	});
-
+		// Full screen
```
llaske: Not the good level of indentation here.

llaske:
That's good but the items should be resized to fit the board, else the new space is unused.
See for example in the following video: when the toolbar disappear, we could expect that items become longer.

Peek 07-12-2019 16-43

PR #480, opened 2019-12-07
Add tutorial to Video Viewer

sdziuda:
Added and modified all the files to let the tutorial start and be localized


llaske:
Nice.

Few remarks:
- Could you add a popup on the video item to explain to the user that he should click to launch the video?
- Your localization don't work because you didn't initialize the language with the Sugarizer language (see [here](https://github.com/llaske/sugarizer/blob/dev/docs/tutorial_step5.md#initialize-localization))
- You could add [fr], [es] and [pt] in the locale.ini, just use the same value than then one in [en].

llaske:
Very cool.
Could you also add a popup on the input field to explain that it allow you to filter by name?

sdziuda:
Thank you!
Of course, I'll add this popup in a moment.

activities/VideoViewer.activity/lib/tutorial.js
```diff
@@ -0,0 +1,86 @@
...
+                content: l10n.get("TutoExportvideoContent")
+            },
+            {
+                element: "#search",
+                title: l10n.get("TutoSearchTitle"),
+                content: l10n.get("TutoSearchContent")
+            },
+            {
+                element: "#app_item",
+                title: l10n.get("TutoVideoideoTitle"),
```
llaske: Look like this string is not defined in the locale.ini. Is a typo?

sdziuda: Oh, yeah, you are right. It's a typo. I'll correct that in a second.

activities/VideoViewer.activity/lib/tutorial.js
```diff
@@ -0,0 +1,86 @@
+define(["webL10n"], function (l10n) {
+	var tutorial = {};
+
+	tutorial.start = function() {
+		var steps = [
```
llaske: Your file mixed indentation with TAB and identation with white spaces. Could you use one sort of indentation only?

llaske:
Could you adapt media queries to avoid to override the search field when the window is resized.

Peek 08-12-2019 16-13

Issue #417, opened 2019-11-09
Add a fullscreen button in Constellation activity


llaske:
The idea is to add a fullscreen button in the toolbar to increase the size for activity content on small devices.

The complete process is describe in #399

llaske:
Fixed in https://github.com/llaske/sugarizer/commit/a02e2dad87e8f578d4939d3ffd1be7cdc5221dea

PR #481, opened 2019-12-07
Fixed Sugarizer issue #417

jontymagan:
Added a fullscreen button in Constellation activity
Added icons
Updated activity.css and activity.js
Updated index.html


activities/Constellation.activity/css/activity.css
```diff
@@ -27,11 +27,29 @@
...
+}
+
+#unfullscreen-button {
+  background-image: url(../icons/view-return.svg);
+  border: 0;
+  border-radius: 0px;
+  margin: 1px;
+  width: 45px;
+  height: 45px;
+  top: 0;
```
llaske: May be you could add 5px to avoid to be just on the top.

activities/Constellation.activity/css/activity.css
```diff
@@ -27,11 +27,29 @@
 }
 
 #starmap{
-  height: 100%;
+  height: 110%;
```
llaske: Not a good idea because it will add a scrollbar

llaske:
Good job. Thanks.

PR #485, opened 2019-12-09
Added Editor mode

Uasked4dat:
Working edit mode.


llaske:
Look like all files in the sugar-web directory were duplicated in your PR.

Plus, because Exerciser is an activity in ReactJS where JS files are generated, please do your PR on the [original repository](https://github.com/llaske/ExerciserReact) so we could see sources files updated.

llaske:
Fixed in https://github.com/llaske/sugarizer/commit/9602b657db3e8a3795155a4fe78c0c4d9cea495c

Issue #405, opened 2019-11-09
Board size in Flip activity is not optimized


llaske:
The board size in Flip activity don't use the dimension of the game.
So for a 7x7 game the window fit to the screen:

![image](https://user-images.githubusercontent.com/981841/68526232-eb79ac80-02d9-11ea-8481-adaaa5c71321.png)

But for a 4x4 game there is a large gap at the bottom of the grid:

![image](https://user-images.githubusercontent.com/981841/68526250-1237e300-02da-11ea-9967-229ede9a5dea.png)

llaske:
Fixed in https://github.com/llaske/sugarizer/commit/d98a415f4d21a11b911ac438fa57ffe8a0637a25

PR #487, opened 2019-12-10
Optimize Flip Activity #405

AndreaGon:
Flip Activity is now optimized. The grid size is updated every time the user levels up, or manually sets the level.

4X4:
flip

7X7:
flip2

15X15:
smallest

Let me know what you think!


llaske:
Good work but I guess something is missing.
When I relaunch the activity, the size is not recompute. So the board has not the right size.

AndreaGon:
The grid size should now be recomputed when the user opens a saved instance of the activity. Also, the activity does not seem to save and restore the data from datastore when it is closed (user turns, current arrangement of dots, etc.). This commit should hopefully fix both issue.

llaske:
Hmmm. You're right. Look like in some case it was not saved. 
Very good job.

PR #496, opened 2019-12-14
Add tutorial for Constellation activity

sdziuda:
Added the tutorial for Constallation activity


llaske:
Look like there is a missing style, the header has not the right color.

image

Should be a darker gray and text color should be write.
Are you sure you've copied all the CSS styles?

sdziuda:
I think it should be fine now. There was probably one line left in code before the title class in css that was making this issue happen.

llaske:
There is still an issue the fullscreen/unfullscreen button is now invisible

sdziuda:
Ok, I've resolved this issue.

PR #497, opened 2019-12-15
superscript and subscript features implemented

Uasked4dat:


llaske:
Good job.

image

Two small things:
- Could you move the new buttons to the right of the insert image button?
- Could you reduce size of the icons, it should have a margin to the up/bottom of the icons (see other letters icons)?

Uasked4dat:
Issues should be fixed, thanks.

PR #499, opened 2019-12-15
Add tutorial for XOEditor activity

sdziuda:
Added the tutorial for XO Editor activity


llaske:
Good job.

Three remarks:
- Could you put the help button to the left of the fullscreen button?
- Could you fix the multiple indentation mode in the tutorial.js?
- Could you rearrange the order in tutorial: the description of the board should be done before the description of the button, because it's the more important here.

Note also than it's possible in the board to move colors to order them, for example by preference or just to play with it

PR #501, opened 2019-12-16
Fix Issue #418

AndreaGon:
Last One Lose Activity should now be able to enter fullscreen mode:

lol

I also moved the playbutton to the left as it overlaps with the activity on screen resize.


llaske:
Sounds good but look like some files are missing in your package: what about SVG for fullscreen/unfullscreen icons?

llaske:
Good job @AndreaGon

Issue #410, opened 2019-11-09
Add a fullscreen button in Physics activity


llaske:
The idea is to add a fullscreen button in the toolbar to increase the size for activities contents on small devices.

The complete process is describe in #399

llaske:
Fixed by https://github.com/llaske/sugarizer/pull/526

PR #502, opened 2019-12-16
Fix issue #410 (add fullscreen button in Physics)

y-dejong:
The PhysicsJS activity can now enter fullscreen mode.


llaske:
Nice start but there is two issues:

- The Stop button should always be to the right of the screen
- When clicking to unfullscreen button, a new item is created (see below)

activities/PhysicsJS.activity/css/activity.css
```diff
@@ -77,6 +77,24 @@ canvas {
   z-index: 1;
 }
 
+#fullscreen-button {
+	background-image: url(../icons/view-fullscreen.svg);
```
llaske: Try to be consistent on indentation you add in the file

activities/PhysicsJS.activity/js/activity.js
```diff
@@ -460,8 +477,11 @@ define(["sugar-web/activity/activity"], function (activity) {
 			var createdStart = null;
 			world.on({
 				'interact:poke': function( pos ){
-					// create body at a static place
-					if (currentType != -1 && pos.y > toolbarHeight) {
+
+					// do not create body if clicking on unfullscreen button
+					if (pos.y < 45 && pos.x < 45) {
```
llaske: Not a good idea, I suggest instead to call preventDefault on the click event of the Unfullscreen button. It should solve the issue.

llaske:
Fixed by https://github.com/llaske/sugarizer/pull/526

Issue #408, opened 2019-11-09
Add a fullscreen button in Calculate activity


llaske:
The idea is to add a fullscreen button in the toolbar to increase the size for activities contents on small devices.

The complete process is describe in #399

llaske:
Fixed in https://github.com/llaske/sugarizer/commit/1d4d0e44aa4a4ce97f22f9e2e1eae4b6ee6051b2

PR #503, opened 2019-12-17
Fix Issue #408

jontymagan:
Added Icons and updated activity.js


llaske:
That's a good start but you have to resize the board else no more space is available. Here you're just hiding the toolbar.

jontymagan:
> That's a good start but you have to resize the board else no more space is available. Here you're just hiding the toolbar.

Okay.
Do I have to increase the size buttons as well because doing so results in overlapping of buttons. Moreover some buttons get hidden.

llaske:
Adding a `sizeincrease` and `sizedecrease` function is not the good way to do that.
My idea was more to rely on current resize method in the activity called when the window size is changed. See https://github.com/llaske/sugarizer/blob/dev/activities/Calculate.activity/js/activity.js#L45

jontymagan:
> Adding a `sizeincrease` and `sizedecrease` function is not the good way to do that.
> My idea was more to rely on current resize method in the activity called when the window size is changed. See https://github.com/llaske/sugarizer/blob/dev/activities/Calculate.activity/js/activity.js#L45

So, do I have to call just the onresize function..

jontymagan:
When the window resizes means when the width of the window is changed the board resizes but when the length is changed their is no change.
I guess as hiding toolbar on fullscreen changes the length of the window and not the width.... the use of CalculateApp.onResize(); function is not working.

llaske:
But there is still a margin to the size of the toolbar in fullscreen mode. This margin is a lost of space, specifically for small screens.

llaske:
That's nice like that.
By the way the screen is not responsive so unfortunately it's not possible to see the result zone in small screens. 

image

But it's not related to the issue #408, I will fill another issue on that.

PR #504, opened 2019-12-17
Add tutorial for LastOneLoses activity

sdziuda:
Added the tutorial for Last One Loses activity


llaske:
Could you solve conflict due to recent integration of fullscreen button?

PR #505, opened 2019-12-17
Add tutorial for Write activity

sdziuda:
Added the tutorial for Write activity


llaske:
Nice work.
Could you add also description for new features subscript/superscript, see https://github.com/llaske/sugarizer/pull/497

PR #506, opened 2019-12-18
Add Tutorial for TamTam Micro Activity

AndreaGon:
TamTam Micro Activity should now have a tutorial:

tamtam


quozl:
Well done, I like it.

activities/TamTamMicro.activity/css/activity.css
```diff
@@ -20,6 +24,135 @@
   z-index: 5;
 }
 
+/*Tutorial CSS*/
+.tutorial-title {
+	background-color: #808080 !important;
```
llaske: Could you avoid mixing different type of indentation? Here indentation should be white spaces.

AndreaGon:
This should fix the indentation issues

llaske:
There is an issue on localization, whatever the language settings in Sugarizer, it's always the language which is used. Look like webl10n library is not initialized in the activity like it should be https://github.com/llaske/sugarizer/blob/dev/docs/tutorial_step5.md#initialize-localization

AndreaGon:
Hmmm, that's weird, as base on the gif above, I've set the Sugarizer language to French and it worked.

I will look into this.

AndreaGon:
@llaske It turns out that I did not add the webL10n initialization from my local work environment into my commits. This new commit should hopefully fix the issue.

Let me know what you think!

llaske:
Nice. It works now. Thanks.

PR #507, opened 2019-12-18
Add tutorial for Abacus activity

sdziuda:
Added the tutorial for Abacus activity


llaske:
Nice work. But the localization is not taken into account: whatever the Sugarizer language, the tutorial use the language for browser. I suggest to force it like in: https://github.com/llaske/sugarizer/blob/dev/activities/LastOneLoses.activity/js/activity.js#L22

PR #509, opened 2019-12-20
Added Fullscreen and Unfullscreen Buttons in Game of Life Activity

jontymagan:
Added Icons and updated activity.js


activities/GameOfLife.activity/js/activity.js
```diff
@@ -217,6 +217,26 @@ function main(Board, State, patterns, color, shadeColor, l10n, dataStore, genSpe
     });
   })
 
+  function resize() {
```
llaske: I don't think you need to reinvent a resize function. There is an existing one [here](https://github.com/llaske/sugarizer/blob/master/activities/GameOfLife.activity/js/activity.js#L208) that you could probably call when clicking on fullscreen/unfullscreen buttons.

jontymagan: I guess the resize function is not effective (the inbuilt one)(handleResize). I removed it and then tested. It is responsible for responsiveness and the handleResize function has a fixed width and length of the Canvas and it changes the size of the canvas .But we have to change the size of the main div to make the activity responsive on fullscreen which is possible only if we change the size of the main division which I did in my own function....
So I guess having a function will be good....

jontymagan: What are your thoughts?

llaske: The handleResize works, try to resize the window, you will see that the method is called each time and the board is resized. So you have to try to call it in the same way.

PR #511, opened 2019-12-20
Add tutorial for Flip activity

sdziuda:
Added the tutorial for Flip activity


llaske:
Nice work. Could you add a popup to explain what's the turn number?
Could you use white space instead of TAB to indent CSS?

Issue #517, opened 2019-12-21
Can't see the previous result on small screens for Calculate activity


llaske:
The Calculate activity is responsive but it's impossible to see the previous results when the screen is small. 

![Peek 21-12-2019 11-17](https://user-images.githubusercontent.com/981841/71306662-9904d300-23e3-11ea-9d9d-90c4edeb30df.gif)

It should be nice to allow user to scroll the screen.

llaske:
Fixed in https://github.com/llaske/sugarizer/issues/517

PR #522, opened 2019-12-22
Solved issue #517 (Can't see result in Calculate)

sarthak-g:
Change in overflow property solves issue #517  .


llaske:
The scrolling zone should not be on the whole screen: the toolbar should always be visible (except in fullscreen mode).

image

Could you do it only in the canvas part?

llaske:
Fixed in https://github.com/llaske/sugarizer/pull/560

Issue #404, opened 2019-11-09
Handle window resize in TankOp activity during play


llaske:
TankOp activity don't handle window resize event during play. So, if the game is started with a small window then resized, the board keep the same size.

![image](https://user-images.githubusercontent.com/981841/68526078-60e47d80-02d8-11ea-829a-f405f5953e19.png)

llaske:
Solved in https://github.com/llaske/sugarizer/pull/524

PR #524, opened 2019-12-23
Solved issue #404 (Handle resize in TankOp)

sarthak-g:
Window resize problem fixed as mentioned in issue #4
ezgif com-video-to-gif (2)

### DIFF
--- a/activities/TankOp.activity/play.js
+++ b/activities/TankOp.activity/play.js
@@ -80,7 +80,7 @@ enyo.kind({
 		} else {
 			this.zoom = 1;
 		}
-
+		console.log(constant.areaHeight);
 		this.$.gamebox.setStyle("max-height: "+(this.zoom*constant.areaHeight)+"px;");
 		this.canvas = this.$.gamebox.createComponent({kind: "Canvas", id: "acanvas", name: "canvas", attributes: {width: constant.areaWidth, height: constant.areaHeight}});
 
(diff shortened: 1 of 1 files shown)


llaske:
Nice.

It's a good idea to do a `resize` method but may be you could call it from the `create` method to avoid duplicate code.

Could you use a smarter algorithm? i.e. imagin an algorithm to compute the zoom level instead of using a list of if statements.

sarthak-g:
I think algo below is good as compare to if statements:
wsize = document.body.clientWidth;
var ratio = Math.abs(wsize-100)/1000;  //statement 1
this.zoom = Math.max(ratio,0.4);  //statement 2
this.zoom = Math.min(this.zoom,1);  //statement 3
In statement 1, 100 is used because it is giving good results after trial and error from various screen sizes.
Statement 2 is to maintain zoom value greater than or equal to 0.4
Statement 3 is to maintain zoom value less than or equal to 1

sarthak-g:
Is this approach good? Can I change it in create method also?

llaske:
Hmmm. Not too bad but Could be better. Is there any way to be sure that the keyboard don't override the board?

sarthak-g:
I subtracted window's width with the width occupied by keyboard and calculate zoom on the basis of remaining width.

sarthak-g:
Is this approach good? or you can suggest some better way?

llaske:
It's very good!
May be you could now factorize this algorithm to do the same in the the create method?

sarthak-g:
Updated code in create method

PR #527, opened 2019-12-23
Add tutorial for GameOfLife activity

sdziuda:
Added the tutorial for Game of Life activity


llaske:
It's a good start.
Good idea to give a specific popup for: under population, next generation, over population, and reproduction. BTW I think that to be understandable, in this tutorial we have no choice to add images into popups to explain transitions. You could for example do GIF based on https://prostheticknowledge.tumblr.com/post/404756220/rules-of-conways-game-of-life

It's possible to add image in popup, see for example [here](https://github.com/llaske/sugarizer/blob/dev/lib/tutorial.js#L54).

Also, miss description on palettes to change speed and change size. May be your repo is not up to date?

sdziuda:
Thank you for advices! I'll apply changes as soon as I can.

sdziuda:
I've decided to add photos explaining the rules instead of gifs because I think they are way easier to understand.

PR #531, opened 2019-12-23
adding new set-time mode in clock Activity(fixing #387)

### LINKED ISSUE #387: Add a set time mode to the Clock activity
To learn how to read time, it's current to use a tool that let children set the time by themselves.
So the idea - suggested by a teacher in our [Saint-Ouen deployment](https://wiki.sugarlabs.org/go/Sugarizer_Saint-Ouen_deployment) - is to create a new "set time" mode in the Clock activity. In this mode, the time will not display the system time but just let user to set hands to a specific time.
Here's a mock up of what the Clock activity could look like in this new mode.

Capture d’écran 2019-10-08 à 17 11 54

Of course, the time set should persist in the Journal. So the next time the Clock activity is launched the time should be set to the time choose by the user.

prakashujjwal1010:
This PR
- adds a toolbar button to switch between normal and new set-time mode
- implements set-time mode
- adds a code to store the chosen time in journal or datastore
- adds localization
- modifies tutorial
- improves local.ini file
- fixes #387 
Screenshot from 2019-12-24 01-32-52
Screenshot from 2019-12-24 01-33-41

A smiley is drawn over the clock if the user sets the correct time i.e. same as given target time
Screenshot from 2019-12-24 01-33-23


llaske:
It's already very cool! Good job!

Few remarks:
- In my idea the set time mode is not a game to guess the right time, it just allow to move hand
- The set time mode icon should be selected (darkgray background) when active (like other modes)
- The digital clock should not be displayed in set time mode (it depends if it's active or not) but if it's displayed the value of numbers should change at the same time that the user move hands
- Could you confirm the new icon use a free/libre licence?

prakashujjwal1010:
- the icon used is licensed under cc 4.0 and i have added the attribution already. so we can use it, no problem.
- yes, gonna add code to make icon selected when active
- i thought it would be better if children can have a game like mode but ready to change it. Also currently what i have done is that the time(target time) gets automatically displayed and the two buttons(write time, write date gets locked) when the user switch to set-time mode because as i said it was intended to be a part of game.

should i modify the set-time mode and change it to non-game mode or change this mode's name and add set-time mode using another toolbar button ?

llaske:
I propose to modify the set-time mode to do a non-game mode like I suggest.
Then do another button to have a game mode. BTW I think than in game mode you should limit guess to hour<=12. Plus I must said that I didn't find how to win the game.

Nice that you've done icon attribution.

prakashujjwal1010:
well, in the game user will have to move the clock's hands and set the time in clock so that the clock show the same time as the target time(displayed just below the clock).

Screenshot from 2019-12-24 01-32-52

as yet clock is showing 07:16:03, so i have to move the clock's hands so that the clock will have time 10: 16:54(target time) 

Screenshot from 2019-12-24 01-33-23

in above when i set the time in clock correctly i.e now the clock is showing target time(10:16:54),
the smiley get drawn over the clock indicating that the user set the clock correctly or we can say user won the game :)

llaske:
It seems that it don't work on my side. May be I'm doing something wrong?

Peek 27-12-2019 23-00

prakashujjwal1010:
@llaske , if you move hour's hands  between 3:30 and 4:00 (not to 3:00, as target time is 3:48:04) , then you will get smiley.

however i have changed that mode and now i think it should not be confusing as before.

modifications done -
- now we have two modes set-time mode and set-time-game mode
- code for movement of hands changed  i.e. now movement of one hand depends on other hands and effect other hands movement (see the gif)
- modified locale.ini
- modified tutorial.js
- attribution of the new icon used
- improves activity.js

## the new set-time mode is : 
ezgif com-video-to-gif

## the new set-time-game mode is
ezgif com-video-to-gif(1)

llaske:
It's very nice. Good job.

BTW I think it's too complex for young children to link hour and minutes hands.
May be we could let hands independent and be tolerant during the game: if the time to set is 3h30 we could accept the hour hand between 3 and the middle of 3 and 4. So even if hands are not linked, the right position will be valid.

What do you think?

One more remark: use 12 instead of 0 for hours.

prakashujjwal1010:
well users don't have to actually link hour and minutes hands  for example if the time to set up is 5:42:24 then they can
- set seconds, minutes and then hour hand
- set hour hand ,then minutes, and seconds hands, (or)

# for first method (to set 7:48:49)
we first set seconds to 49 hand then minutes hand  to 48 then while moving hour hands we have some restriction or we have only 12 stops for hour hand , one stop between every pair of two numbers
ezgif com-video-to-gif (2)

# for second method
if i set my hour hand to 2 then if i set my minutes hand to 28 , my hour hand automatically moves to the right position and then we can set seconds hand to 4
see gif below for ( 2:28:04)
ezgif com-video-to-gif (1)

so i have provided restrictions to the movement of hands( like limited stops/positions for hour's hand which will depend upon minutes, seconds, movement of hour hand by moving minutes hand) which have reduced the complexity.

also personally in my view, after playing around for a bit children will get it. and the intention of game was to make children indulge in activity so that they have to move hands for sometime to get the right time.

also
i've changed 00 to 12 :)

llaske:
Okay, make sense like that. Good job.

Few remarks:

- There is an issue in set mode when moving from 9:59 to 10:00. Sometimes the digit display 9:00 instead of 10:00 See video below:

Peek 01-01-2020 12-08

- In game mode, sometimes there is a question to set seconds to 60! See capture below:

Capture du 2020-01-01 12-10-08

- In game mode, I think we should reset the 3 hands to 0 and to avoid confusion, between setting and asking, we could add a question mark at the end of the time in digit. So the use could know that we're asking to set a time and we're not displaying the time.

prakashujjwal1010:
bugs are fixed.

llaske:
Very cool. The game is now perfect! :+1: 

BTW regarding set time is still complicated to obtain the right hour. See my attempt below to set clock to 12:00:00 and 8:15:00

Peek 05-01-2020 16-11

prakashujjwal1010:
@llaske ,i think its now fixed.
well several bugs were there because of using Math.floor() with calculation including Math.PI .but after using toFixed(), those bugs are fixed :)

llaske:
Very good! it perfectly works now!

One issue however it doesn't work in touch screen:

Peek 05-01-2020 22-51

It's probably because you define only click event. Could you support also the touch event? It's easier to support both. See for example [here](https://github.com/llaske/sugarizer/blob/dev/activities/Gridpaint.activity/js/events.js#L33).

prakashujjwal1010:
### Update
- touch events added

llaske:
Everything is perfect.
Excellent job. It will be a very useful feature, asked by teachers on the field.
Thanks a lot for your work and your investment.

llaske:
@prakashujjwal1010 , could you have a look on https://github.com/llaske/sugarizer/issues/581

prakashujjwal1010:
@llaske , i have made a PR #582 fixing the issue.
the reason for the bug was that for every call of clock.drawNiceBackground(), always new and same 'clock.svg' image was being loaded, rather than using the preloaded image.

llaske:
Nice. Thanks.

PR #533, opened 2019-12-24
Fixed Issue #517 (Can't see result in Calculate)

s-mag:


llaske:
The scrollbar should not be on the body. Only the canvas part could be scrollable.

image

s-mag:
Can you differentiate between the body and the canvas part...
I am not getting it.

s-mag:
Is it the bootstrap-container-fluid div??

llaske:
As you've understood, what I'm calling the canvas part is the div under toolbar: [this](https://github.com/llaske/sugarizer/blob/dev/activities/Calculate.activity/index.html#L29).

You don't fix the issue, the result part is scrollable but the screen is not scrollable to see the result part when the screen is too small:
 
image

s-mag:
Screen Shot 2019-12-26 at 00 41 38
Screen Shot 2019-12-26 at 00 41 29
But its coming on my PC...

s-mag:
> You don't fix the issue, the result part is scrollable but the screen is not scrollable to see the result part when the screen is too small:

I don't know why this problem is coming..

s-mag:
I guess there's an issue here in [this ](https://github.com/llaske/sugarizer/blob/dev/activities/Calculate.activity/css/bootstrap.min.css#L1424)code

s-mag:
The scrollbar is coming only till the width of 512px..

llaske:
I don't think there is everything related to size 512px.
There is no error in bootstrap.css, may be you could force style with !important

llaske:
Don't work.

It seems that the buttons part is fixed and the result part fit the leaving space under the buttons parts and scroll inside.

Look like that to fix this issue, the whole way of drawing the board should be review.

Do you suggest something?

s-mag:
I guess the old way should be implemented 
Wherin I added the scrollbar to body and removed the scrollbar for results...
I guess that would be a lot better

s-mag:
Moreover if we are working on a big screen we can easily see all the results

s-mag:
Is there any problem if we shift the scrollbar to the body and make the toolbar fixed...

s-mag:
In this https://github.com/llaske/sugarizer/commit/56c16f4d89e511997020c3725716651589318c11 commit I guess it was working fine..

llaske:
Of course, the problem is not for big screens but to allow the activity to be usable on small screens.
The commit you mention is not correct because as I said the whole body zone was scrollable. You've just forced the toolbar to be on top of other items.
The expected result is that if the canvas part can't be displayed it should be the only part scrollable. See for example what happen in Write activity with long text.

PR #512, opened 2019-12-21
Add Tutorial for Clock Activity

AndreaGon:
Clock Activity should now have a tutorial:

clock

PR #537, opened 2019-12-27
Add tutorial for Calculate activity

sdziuda:
Added the tutorial for Calculate activity


llaske:
That's nice.

Could use media queries to hide the help button on small screens? Currently the button override the body part. See below:

Capture du 2019-12-28 18-45-44

prakashujjwal1010:
@sdziuda , I think you should add translated strings in fr, es, pt sections in locale.ini. you can use google translate for it :)

sdziuda:
Thanks for an advice @prakashujjwal1010 :)
As you had suggeted, I added the translations in es, fr and pt from google translate.

AndreaGon:
I don't think it's necessary to add the translations. I've tried it before, but @llaske reverted it back to English. See his comment on my PR #512.

Although I still do translate it to fr, for testing purposes.

sdziuda:
Yeah, after I thought about it for a while, probably google translate is not the best translator (for sure it is not as good as human). But in the end it is @llaske choice about which language will be merged into dev branch.

llaske:
@AndreaGon is right, you don't need to translate because I think it's better to let a native user localize its part. BTW it's nice to test with another language from the one in the browser to ensure tutorial localization work.

PR #540, opened 2019-12-28
Fix issue 438 in Dev branch

Shak2000:
I think that now I created a PR on dev branch


llaske:
It seems that there is an issue on localization handling. It's always the browser language used for tutorial.

Shak2000:
The user settings only determines the initial language displayed. You can change the language by clicking on the screwdriver icon in the activity and then clicking on the arrows on the third row that appears.

Screen Shot 2019-12-29 at 2 51 17 PM

In the screenshot image, the language is set to French. Therefore, the title and the tutorial content are in French as well. This is different from what other activities use for language, but I wanted to be consistent with this specific activity.

FreddieN:
Ah it seems like the activity author has written their own implementation of language selection. @Shak2000 could you reimplement web localisation like stated in the tutorial for the whole activity? It should not be too tricky as it seems they have used l10n for the actual colouring titles. Otherwise the code works fine. 👍

FreddieN:
code

Relevant section of tutorial here: https://github.com/llaske/sugarizer/blob/master/docs/tutorial_step5.md

Shak2000:
Do you want me to remove the language selection in the activity itself, including the UI, or should I just change the tutorial to use the standard language selection method?

FreddieN:
Just remove the language selection component within the html and then update the l10n logic to use the model that all the other activities use like in the tutorial.

Shak2000:
I removed the language changing feature from the activity itself to keep the language setting consistent with the rest of the platform.

Issue #538, opened 2019-12-28
Horizontal scrollbar appears in the Write activity


s-mag:
<img width="611" alt="ss" src="https://user-images.githubusercontent.com/44548987/71561283-6d947f00-2a9a-11ea-93b7-87469480ff6a.png">

I guess making the scrollbar on x-axis hidden will solve the problem. 
@llaske What do you think??

s-mag:
This problem comes when size of the screen is changed to small...

s-mag:
@FreddieN What do you think about this issue??

FreddieN:
I cannot seem to replicate the issue @s-mag I've tried using both chrome and firefox. Do I just make the window small like this?

![demo](https://i.imgur.com/WQjXSBP.png)

s-mag:
> I cannot seem to replicate the issue @s-mag I've tried using both chrome and firefox. Do I just make the window small like this?
> 
> ![demo](https://camo.githubusercontent.com/aa82c0cf64404ee42f2a43602b6ac24df521af97/68747470733a2f2f692e696d6775722e636f6d2f57516a585342502e706e67)

Well I am using firefox an the issue comes if you resize to small screens...and when you decrease the width...

s-mag:
Sorry I made a mistake that’s scrollbar...not toolbar

FreddieN:
Oh I see... Yes that is a bug. I will make a task instance on GCI now.

![bug](https://i.imgur.com/yUC7k3m.png)

FreddieN:
![task](https://i.imgur.com/Xc3o8uV.png)

It will appear once approved by an Org Admin.

s-mag:
> ![task](https://camo.githubusercontent.com/675ac2a111f5d412e46dc91fe4c04352c847168d/68747470733a2f2f692e696d6775722e636f6d2f5863336f3875562e706e67)
> 
> It will appear once approved by an Org Admin.

Thanks 
BTW I have fixed it and made a PR ...
I have referenced it above...

llaske:
It's a real issue. I've published the related GCI task here: https://codein.withgoogle.com/dashboard/tasks/5827221755789312/

llaske:
Fixed in https://github.com/llaske/sugarizer/pull/541

PR #541, opened 2019-12-29
Fixed unwanted scrollbar #538

s-mag:


s-mag:
Fixed Issue #538

s-mag:
@llaske Is the issue coming on your system???

llaske:
Good job. Thanks. Claim the matching GCI task if you want.

PR #542, opened 2019-12-30
Added tutorial in Grid Paint Activity.

s-mag:


llaske:
It don't work. The screen become gray but the popup doesn't appears.

s-mag:
> It don't work. The screen become gray but the popup doesn't appears.

This is the problem emerging with me ... I did all the steps correctly but I am not getting why the popups doesn’t appear... They work if we click right arrow from the keyboard but the css is not showing up... I don’t know why... Can you suggest me a way to solve this issue...

llaske:
Not sure it's related but webL10n is not initialized in this activity.

s-mag:
> Not sure it's related but webL10n is not initialized in this activity.

Its not related to wbL10n.....

llaske:
It's related to webL10n, when I'm removing all call to `l10n.get` method in your file, the popup is displayed...

s-mag:
> It's related to webL10n, when I'm removing all call to `l10n.get` method in your file, the popup is displayed...

Oh yes.. Thanks. I will do the necessary changes now..

s-mag:
Added Popups and fixed the Issue.
Thanks.

llaske:
You must not hard code the text for the tutorial, you should make localization work!

sdziuda:
I've also encountered the problem with localization not being initialized at all. Anyway fixing it is really simple. You need to go back to [step 5 of the tutorial](https://github.com/llaske/sugarizer/blob/dev/docs/tutorial_step5.md) and add some code in activity.js and index.html
I hope I could help you a little :)

s-mag:
> I've also encountered the problem with localization not being initialized at all. Anyway fixing it is really simple. You need to go back to [step 5 of the tutorial](https://github.com/llaske/sugarizer/blob/dev/docs/tutorial_step5.md) and add some code in activity.js and index.html
> I hope I could help you a little :)

Thanks man...

s-mag:
> You must not hard code the text for the tutorial, you should make localization work!

Ok sir I will add localization

s-mag:
Added Localization but i have not translated for other languages..

llaske:
It's better.

Two remarks:

- The webL10n is not initialized so the tutorial language is always the language of the browser. You should initialized by using code mentioned [here](https://github.com/llaske/sugarizer/blob/dev/docs/tutorial_step5.md#initialize-localization).
- Will be nice to display a different tutorial in list mode and in detailed mode. You could differentiate it using the global variable `mode`. See [here](https://github.com/llaske/sugarizer/blob/dev/activities/Gridpaint.activity/js/gridpaint.js#L55).

s-mag:
> It's better.
> 
> Two remarks:
> 
> * The webL10n is not initialized so the tutorial language is always the language of the browser. You should initialized by using code mentioned [here](https://github.com/llaske/sugarizer/blob/dev/docs/tutorial_step5.md#initialize-localization).
> * Will be nice to display a different tutorial in list mode and in detailed mode. You could differentiate it using the global variable `mode`. See [here](https://github.com/llaske/sugarizer/blob/dev/activities/Gridpaint.activity/js/gridpaint.js#L55).

I will do the changes right away..

s-mag:
Added different tutorial for different modes and fixed Localization.

llaske:
Nice.

One more issue, a horizontal scrollbar appears when the tutorial is launched:

image

s-mag:
Changed Placement.. to center

s-mag:
@llaske Is this approach fine.

Issue #389, opened 2019-10-12
Add a synth keyboard mode in TamTam Micro activity


llaske:
The idea of this enhancement is to add a keyboard mode into the TamTam Micro activity to allow user to play like a keyboard with the sound he choose. It will transform the activity into a mini synth keyboard.

To do that there is three things to do:

- Replace the current audio implementation [here](https://github.com/llaske/sugarizer/blob/dev/activities/TamTamMicro.activity/audio.js) using HTML5.Audio by the [Web Audio API](https://developer.mozilla.org/en-US/docs/Web/API/Web_Audio_API/Simple_synth)
- Integrate a Piano keyboard in the activity. Could be [this one](https://github.com/FreddieN/sugarizer/tree/master/activities/Piano.activity).

![uOQpaJI](https://user-images.githubusercontent.com/981841/66698475-c4ca5500-ecde-11e9-8914-7b4386044e67.gif)

- Implement synth instructions: could be something like [this](https://developer.mozilla.org/en-US/docs/Web/API/Web_Audio_API/Simple_synth).

pidddgy:
@llaske would you want something where you can right click on a sound and a popover appears, with a button that goes to the piano view?

llaske:
Hmmm. Right click is not a good idea because Sugarizer could be use on a touch screen.
Instead I'm thinking to:

- Update the activity to allow to "select" a sound. The selected sound will be the most recently used.
- Add a button to switch to a "Piano mode" in the toolbar. In Piano mode the list of sounds will be replaced by the piano view.

llaske:
Fixed in https://github.com/llaske/sugarizer/pull/544

PR #544, opened 2020-01-01
add synth keyboard to tamtam activity(WIP)

pidddgy:
#389


llaske:
Look like the file Toggle%20Piano.png is missing in the repo.

Warning: do not change indentation in the file. Here you should use TAB instead of white spaces.

activities/TamTamMicro.activity/piano.js
```diff
@@ -0,0 +1,81 @@
+enyo.kind({
+	name: "TamTam.Piano",
+    components: [
```
llaske: Please don't mix indentation, should be TAB

activities/TamTamMicro.activity/index.html
```diff
@@ -11,10 +11,12 @@
 	href="lib/sugar-web/graphics/css/sugar-200dpi.css">
 <link rel="stylesheet" href="css/activity.css">
 <link href="lib/enyo/enyo.css" rel="stylesheet" type="text/css" />
+<script src="https://unpkg.com/tone@next/build/Tone.js"></script>
```
llaske: Do not reference library on the web. Everything should be copied locally to allow offline mode

activities/TamTamMicro.activity/app.js
```diff
@@ -47,33 +48,53 @@ enyo.kind({
...
+			for(var i = 0 ; i < len ; i++ ) {
+				this.$.collections.createComponent(
+					{ kind: "TamTam.Collection", name: TamTam.collections[i].name, selection: (i == this.collection), ontap: "changeCollection" },
+					{ owner: this }
+				).render();
+			}
+			
+			// TODO: get icon
+			this.$.collections.createComponent(
+				{ kind: "TamTam.Collection", name: "piano", selection: (i == this.collection), ontap: "changePianoMode" },
```
llaske: This line generate a message in the browser console:

`
    enyo.js:21 enyo.Component.addComponent():  Duplicate component name "piano" in owner "app" violates unique-name-under-owner rule, replacing existing component in the hash and continuing, but this is an error condition and should be fixed.`

Seems related to the fact that you can't use piano as `name` because another component (probably the piano item) has already the same id.

activities/TamTamMicro.activity/piano.js
pidddgy: 3e59733

activities/TamTamMicro.activity/index.html
pidddgy: 4f99ebb15387e3e726bf0df0da8199b0d339e31f

activities/TamTamMicro.activity/app.js
pidddgy: fb7d165

llaske:
Nice. You solved the comments in the review.
The idea is to put the piano button in the toolbar not at the same place that the collections.
The image in the piano view should be the image of the item played.

llaske:
Good.

Few remarks:

- It should be nice to surround differently the last instrument clicked (to see the selection)
- The piano button in the toolbar should have the class `active` when you're on the piano view
- Don't use black background for item, use the same color than on the collection
- When the first click is on the piano icon, the piano view is incorrect (see below)

image

- Remove all log messages in your code
- I can't merge your code because you've not the last dev branch (miss the fullscreen and tutorial button)

pidddgy:
if by "the piano view is incorrect" you mean that there's no icon at the top showing the currently selected instrument, that's because there is no currently selected instrument by default - i can make it default to something if you want

pidddgy:
re: When the first click is on the piano icon, the piano view is incorrect (see below) I've made it default to using the piano instrument for now.

activities/TamTamMicro.activity/app.js
```diff
@@ -47,33 +47,49 @@ enyo.kind({
...
+			this.$.collections.applyStyle("background-color", '#ffffff');
 
-		// Display items
-		var collection = TamTam.collections[this.collection];
-		var len = collection.content.length;
-		for(var i = 0 ; i < len ; i++ ) {
-			var item = this.$.items.createComponent(
-				{ kind: "TamTam.Item", name: collection.content[i] },
+			if(currentPianoMode !== null) {
+				this.$.items.createComponent(
```
llaske: Please use this.userColor.stroke for background of this item.
See [here](https://github.com/llaske/sugarizer/pull/544/files#diff-c2fb8208740e638314fe55c3feeed5b6R90) to see how.

activities/TamTamMicro.activity/images/database/piano-icon.svg
```diff
@@ -0,0 +1,55 @@
+<?xml version="1.0" standalone="no"?><!DOCTYPE svg PUBLIC "-//W3C//DTD SVG 1.1//EN" "http://www.w3.org/Graphics/SVG/1.1/DTD/svg11.dtd"
```
llaske: Please move this file into `TamTamMicro.activity/icons` directory instead.

activities/TamTamMicro.activity/index.html
```diff
@@ -17,10 +17,12 @@
 <script src="lib/bootstrap-tour-standalone.min.js"></script>
 <script>if (window.module) module = window.module;</script>
 <link href="lib/enyo/enyo.css" rel="stylesheet" type="text/css" />
+<script src="lib/Tone.js"></script>
```
llaske: Please move this file into the` if (typeof module)` and `if (window.module)` to avoid possible incompatibility with electron as explained [here](https://github.com/llaske/sugarizer/blob/dev/docs/tutorial_step9.md#integrate-bootstrap-tour-components)

activities/TamTamMicro.activity/item.js
```diff
@@ -17,24 +17,20 @@ enyo.kind({
 	
 	// Item setup
 	nameChanged: function() {
-		this.$.itemImage.setAttribute("src", "images/database/"+this.name+".png");
+        if(currentPianoMode === this.name) {
```
llaske: Please respect indentation here

activities/TamTamMicro.activity/item.js
```diff
@@ -17,24 +17,20 @@ enyo.kind({
...
+            this.$.itemImage.setAttribute("src", "images/database/"+this.name+"sel.png");
+        } else {
+            this.$.itemImage.setAttribute("src", "images/database/"+this.name+".png");
+        }
 	},
 	
 	// Play sound using the media
 	play: function() {
+        currentPianoMode = this.name;
```
llaske: Please respect indentation here

main.js
```diff
@@ -15,7 +15,7 @@ var dialog = electron.dialog;
 
 var mainWindow = null;
 
-var debug = false;
+var debug = true;
```
llaske: Please set again to false

activities/TamTamMicro.activity/js/activity.js
```diff
@@ -50,7 +50,15 @@ define(["sugar-web/activity/activity", "sugar-web/env", "tutorial", "webL10n"],
 		// Stop sound at end of game to sanitize media environment, specifically on Android
 		document.getElementById("stop-button").addEventListener('click', function (event) {
 			sound.pause();
-		});
+        });
```
llaske: Please respect indentation here

llaske:
Hmmm. The idea was not to close piano mode when there is a click on tutorial button.
The idea was to add a part of the tutorial in piano mode to explain the piano UI:

- The current instrument
- The keyboard
- The toolbar button to go back to normal mode

You could easily customize tutorial depending of the mode, see for example what's done in Gridpaint [here](https://github.com/llaske/sugarizer/blob/dev/activities/Gridpaint.activity/lib/tutorial.js#L40).

Also, it will be nice to allow use of keyboard to play piano. It's why number 1 to 7 are written on the keys. You could do easily by adding a keyboard listener like [here](https://github.com/llaske/sugarizer/blob/master/js/password.js#L51). May be we could detect if CTRL key is pressed to play the matching black key.

main.js
llaske: Look like you've set again debug to true...

pidddgy: a0db5c8 set it to false in here, my bad

llaske:
Tutorial is perfect now.

Regarding keyboard, something goes wrong. It look like each time a key is pressed the event keydown is received multiple times so the sound played is very bad.
May be you could prevent playing multiple times the same note until the keyup event is received?

activities/TamTamMicro.activity/piano.js
```diff
@@ -69,6 +98,16 @@ enyo.kind({
 		player.connect(pitchShift);
 		player.autostart = true;
 	},
+	
+	create: function() {
+		this.inherited(arguments);
+		this.collection = 0;
+		var that = this;
+		document.addEventListener('keydown', function(event) {
```
llaske: Be careful: each time the screen is created, a new event listener is registered. You should remove the event listener when leaving the piano mode else the method will be called one more time each time you reopen the screen.

activities/TamTamMicro.activity/piano.js
```diff
@@ -119,4 +120,13 @@ enyo.kind({
 	abort: function() {
 		this.endofsound();
 	},
+
+	destroy: function () {
+		var that = this;
+		window.clearInterval(that.job);
```
llaske: Not sure to understand what is this `that.job` ?

activities/TamTamMicro.activity/piano.js
```diff
@@ -119,4 +120,13 @@ enyo.kind({
...
+		that.inherited(arguments);
+		document.removeEventListener('keydown', function(event) {
```
llaske: You should call the removeEventListener with the same function that the one you've registered in addEventListener. Here you pass a function that do the same think but I'm not sure it the same.
So, I guess you must set the function in a variable to be sure to pass the same value.

activities/TamTamMicro.activity/piano.js
pidddgy: addressed in 
e281f29

pidddgy:
e216aa3 re: glitch with app redrawing, i wasn't able to reproduce the error but I think this will fix it, as .update() only redraws the parts of the component that was changed

activities/TamTamMicro.activity/piano.js
pidddgy: 11c503b

activities/TamTamMicro.activity/piano.js
pidddgy: e281f29 my bad, was following a code sample and left that in by accident lol

activities/TamTamMicro.activity/piano.js
pidddgy: 6dd6371 not sure if using .bind will cause the same problem, but it was broken before and .bind fixed it. But it's not an anonymous function, so maybe this will work?

pidddgy:
not entirely sure if this will work, but I've added a fix that destroys the items and collections all at once rather than pushing collections to an array, destroying all the collections, pushing all the items, and so on. Hopefully this will make updating faster. Want to test again @llaske ?

llaske:
Nice, the UI glitch is fixed.

BTW I'm afraid something wrong. When I switch several times between Piano mode and Instruments mode, then use keyboard, the sounds is played several times (I've added a log message [here](https://github.com/llaske/sugarizer/pull/544/files#diff-59dbdad8972f42f759eded15b3a1ccf6R96) to be sure). Like when the repeat was not processed. Could means that the removeEventListener don't work...

activities/Abecedarium.activity/play.js
```diff
@@ -301,8 +301,6 @@ enyo.kind({
 	// Check taped
 	checkTaped: function() {
 		this.forbidentry = true;
-		if (this.playing != null)
```
llaske: This change is not related to this PR. Please exclude this file.

llaske:
Nice. The sound issue is fixed.

BTW the UI glitch is still there. I've said that it disappears but it's not the case.
I've found the issue, it's related to [this line](https://github.com/llaske/sugarizer/pull/544/files#diff-ec78797add88abde31a1dde83a9c6baaR61). It force the redrawing of the whole screen at each click.

I guess you've done it to handle change selection? May be you could try to find a way to do that without redrawing the whole screen. Why not changing only the item state?

Also, look like the file `images/database/pianobutton.png` is no longer used.

llaske:
Nice. The UI Glitch is now fixed.

Still an issue. When clicking on the item image in piano mode there is an error in the console:

```
item.js:31 Uncaught TypeError: Cannot read property 'setAttribute' of undefined
    at e.TamTam.Item.deselect() [as deselect] (item.js:31)
    at enyo.instance.TamTam.App.updateItem() [as updateItem] (app.js:108)
    at enyo.instance.enyo.Component.dispatch() [as dispatch] (enyo.js:63)
    at enyo.instance.enyo.Component.dispatchEvent() [as dispatchEvent] (enyo.js:63)
    at enyo.instance.enyo.Control.dispatchEvent() [as dispatchEvent] (enyo.js:171)
    at enyo.instance.enyo.Component.dispatchBubble() [as dispatchBubble] (enyo.js:63)
    at enyo.Component.bubbleUp() [as bubbleUp] (enyo.js:63)
    at enyo.Component.dispatchBubble() [as dispatchBubble] (enyo.js:63)
    at enyo.Component.bubbleUp() [as bubbleUp] (enyo.js:63)
    at enyo.Component.dispatchBubble() [as dispatchBubble] (enyo.js:63)
TamTam.Item.deselect() @ item.js:31
TamTam.App.updateItem() @ app.js:108
enyo.Component.dispatch() @ enyo.js:63
enyo.Component.dispatchEvent() @ enyo.js:63
enyo.Control.dispatchEvent() @ enyo.js:171
enyo.Component.dispatchBubble() @ enyo.js:63
enyo.Component.bubbleUp() @ enyo.js:63
enyo.Component.dispatchBubble() @ enyo.js:63
enyo.Component.bubbleUp() @ enyo.js:63
enyo.Component.dispatchBubble() @ enyo.js:63
enyo.Component.bubbleUp() @ enyo.js:63
enyo.Component.dispatchBubble() @ enyo.js:63
enyo.Component.bubbleUp() @ enyo.js:63
enyo.Component.dispatchBubble() @ enyo.js:63
enyo.Component.bubble() @ enyo.js:63
n.<computed>.n.<computed> @ enyo.js:63
TamTam.Item.play() @ item.js:36
enyo.Component.dispatch() @ enyo.js:63
enyo.Component.dispatchEvent() @ enyo.js:63
enyo.Control.dispatchEvent() @ enyo.js:171
enyo.Component.dispatchBubble() @ enyo.js:63
enyo.Component.bubbleUp() @ enyo.js:63
enyo.Component.dispatchEvent() @ enyo.js:63
enyo.Control.dispatchEvent() @ enyo.js:171
enyo.Component.dispatchBubble() @ enyo.js:63
enyo.Component.bubble() @ enyo.js:63
dispatchBubble @ enyo.js:186
dispatch @ enyo.js:186
e.dispatch @ enyo.js:186
sendTap @ enyo.js:195
up @ enyo.js:195
mouseup @ enyo.js:195
(anonymous) @ enyo.js:195
dispatch @ enyo.js:186
e.dispatch @ enyo.js:186
Show 5 more frames
```

Previously, it played the original sound.

activities/TamTamMicro.activity/app.js
```diff
@@ -105,7 +105,15 @@ enyo.kind({
 	},
 	
 	updateItem: function(sender, event) {
-		this.openItems[event.toChange].deselect();
+		try {
```
llaske: Wooaw, using try/catch to avoid the issue is not a good practice 😮 
Will be better to test if the `event.toChange` and `this.openItems[event.toChange]` is not null before calling `deselect` method

llaske:
It perfectly works now. Thanks a lot for your work and your determination.
It will be a very useful feature on the field.

llaske:
@pidddgy, could you have a look on https://github.com/llaske/sugarizer/issues/564 ?

PR #545, opened 2020-01-01
Add tutorial for FoodChain activity

sdziuda:
Added the tutorial for FoodChain activity


llaske:
It's a good start but we need to differentiate tutorial for each game and add popup on each UI element.

sdziuda:
Do you mean something like what is done right [here](https://github.com/llaske/sugarizer/blob/master/lib/tutorial.js#L76)?
If yes, I get the whole idea of how to do it but I'm not sure yet how to know on which screen the user is currently on.

llaske:
Yes, it's exactly what I mean.
I think you could test the value of context, see [here](https://github.com/llaske/sugarizer/blob/dev/activities/FoodChain.activity/app.js#L155).

sdziuda:
Okay, thank you very much! I think I figured it out :)
If I have any more questions, I will write.

sdziuda:
Now the popups are different for each of the games in the activity.

llaske:
It's very good.

It seems however that once you've been to a game, the tutorial don't detect when you go back to the home page. Following is the result of clicking on the help button when coming back from the Play game:

image

I guess you could test the `Foodchain.context.game` value (instead of `FoodChain.context.object.id` ?) to avoid this issue.

activities/FoodChain.activity/lib/tutorial.js
```diff
@@ -0,0 +1,251 @@
...
+					}
+				}
+			},
+			onEnd: function() {
+				tutorial.elements = [];
+				tutorial.icons = null;
+			}
+		});
+		var currentView=FoodChain.context.object.id;
+		console.log(currentView);
```
llaske: Please, remove the debug log here

llaske:
Oops look like there is now an issue on the home page

image

sdziuda:
I think I know what happened, probably I gave the same name to Play button and PlayGame button text in locale.ini file, I'll go check it out.

sdziuda:
Yeah, I was right. I gave the same name to Play button and PlayGame button in locale.ini file.

PR #548, opened 2020-01-03
Add tutorial for Abecedarium activity

sdziuda:
Added the tutorial for Abecedarium activity

PR #551, opened 2020-01-03
Add tutorial for Abecedarium activity

sdziuda:
I've really messed up the previous PR (#548) so I've decided to do a new one. I hope now it works fine.


llaske:
Very good.

Just miss description of two exports buttons in the toolbar in Play mode.

image

sdziuda:
They are a little problematic to add popups for them. As you can see on the screenshot the popups appear even if they shouldn't.
image

I think that it is because this buttons exist but they are hidden [using style in index.html](https://github.com/llaske/sugarizer/blob/b2e55be2ce509156a5800eb300c04a34c105947d/activities/Abecedarium.activity/index.html#L44). I'm not sure how to resolve this problem.

sdziuda:
> They are a little problematic to add popups for them. As you can see on the screenshot the popups appear even if they shouldn't.
> image
> 
> I think that it is because this buttons exist but they are hidden [using style in index.html](https://github.com/llaske/sugarizer/blob/b2e55be2ce509156a5800eb300c04a34c105947d/activities/Abecedarium.activity/index.html#L44). I'm not sure how to resolve this problem.

Nevermind, I figured it out by now. I'll push commit in just a second.

Issue #549, opened 2020-01-03
Resize function gives error in TankOp Activity


s-mag:
I guess there is a bug in the code of play.js file.
It is due to the last commit with respect to the issue #404 .
Error comes in an infinite loop.
<img width="957" alt="Error" src="https://user-images.githubusercontent.com/44548987/71746333-9a79d500-2e92-11ea-81d6-9a341cba8695.png">

The error arises after we change modes of the game and the game stops working.
![ezgif com-video-to-gif (1)](https://user-images.githubusercontent.com/44548987/71746609-59ce8b80-2e93-11ea-80f6-04ee7ff3038a.gif)

@llaske Can you replicate this Issue?

s-mag:
Due this issue I was not able to resize screen while fixing issue #547 
I am working on it...

llaske:
Hmmm. You're right.

s-mag:
@llaske should I recreate resize function in play.js

sarthak-g:
@s-mag thanks for raising this issue.
Fixed this issue in pull request #552

s-mag:
@sarthak-g there is a new error coming
see here..

<img width="957" alt="Error2" src="https://user-images.githubusercontent.com/44548987/71763234-a8b20a80-2eff-11ea-93a5-a6f524cf959b.png">

Kindly fix this error as well...

s-mag:
I have fixed this issue in #553

llaske:
Fixed in https://github.com/llaske/sugarizer/pull/552

PR #552, opened 2020-01-04
solved issue #549(resize function error)

sarthak-g:
Issue #549 fixed.
Issue is due to changing id of div everytime we play game(without reloading).


s-mag:
Oh thanks mate. I will check now...

s-mag:
@sarthak-g there is a new error coming
see here..
Error2
Kindly fix this error as well...

s-mag:
> @sarthak-g there is a new error coming
> see here..
> Error2
> Kindly fix this error as well...

This error arises if you call the resize function....

sarthak-g:
@s-mag  error arises because you are trying to call resize method when the game is not started.
Note: resize method is only for resizing canvas(battlefield) when game is started.

s-mag:
> @s-mag error arises because you are trying to call resize method when the game is not started.
> Note: resize method is only for resizing canvas(battlefield) when game is started.

Oh I see nvm
Thanks....

s-mag:
But the old issue has not been resolved if you change modes 5-6 times the old error comes again in infinite loop.....

s-mag:
No need to work on this issue I have fixed it..
Thanks

sarthak-g:
> 
> 
> @sarthak-g No need to work on this issue I have fixed it..
> Thanks

Your function doesn't resize at all.

sarthak-g:
> 
> 
> @sarthak-g But the old issue has not been resolved if you change modes 5-6 times the old error comes again in infinite loop..... without calling the resize function.

I have tried more than 5-6 times with changing modes everytime, it is working fine.

s-mag:
The code worked fine...
But I altered it a little bit because I am adding fullscreen buttons and calling the resize function which was causing an error.

sarthak-g:
You can't remove that part of code

> 
> 
> The code worked fine...
> But I altered it a little bit because I am adding fullscreen buttons and calling the resize function which was causing an error.


Your alteration doesn't resize properly.The part you removed is to make sure that keyboard does not overlap with canvas.

s-mag:
Ohk. Then I should create a zoom function for the fullscreen issue.
Thanks.

llaske:
It works. Good job @sarthak-g and @s-mag.

PR #553, opened 2020-01-04
Fixed Issue #404 & #549

s-mag:
Fixed Issue #404 & #549


sarthak-g:
Your changes doesn't resize at all. See below:
tankdemogig

s-mag:
Oh looks like I forgot the play mode..
Will fix it..

s-mag:
Altered the resize function a little bit and removed the code which was causing errors.

s-mag:
Some errors were coming when I added fullscreen buttons and called resize function.
So I altered the code a little bit.

Issue #342, opened 2019-04-12
Grid Paint Activity screen not working properly on Firefox


AndreaGon:
Grid activity does not completely fill the screen on Firefox, and it shows a black background to the right:

![image](https://user-images.githubusercontent.com/44361130/56013826-3839dc80-5d25-11e9-8b71-77e5a8869b74.png)

![image](https://user-images.githubusercontent.com/44361130/56013851-56074180-5d25-11e9-94a4-a56aedd48566.png)

It works properly on Chrome.

Scar26:
Working on it

llaske:
Fixed in https://github.com/llaske/sugarizer/commit/99ed20d71c4f18635efacf0d61b2da383fc28b76

Issue #546, opened 2020-01-01
Add a home button in Grid Paint activity


s-mag:
A home button should be added on toolbar so that a palette or drawing board can be changed instead of the small round button in the corner...

I guess this should avoid confusion and will improve UI.

![Grid Paint](https://user-images.githubusercontent.com/44548987/71645006-674f0e80-2cf7-11ea-8bef-97ff83d59641.gif)
 
<img width="960" alt="gp" src="https://user-images.githubusercontent.com/44548987/71645547-1b549780-2d00-11ea-92d8-df2e10e5eb63.png">

Currently tested in Firefox...
What do you think @llaske @FreddieN ?

llaske:
I'm agree that the meaning of the small green circle is not clear.
BTW I'm not sure adding a button in the toolbar is a good idea. May be we can just transform the circle in a more significant icon. Do you have any suggestion?

s-mag:
> I'm agree that the meaning of the small green circle is not clear.
> BTW I'm not sure adding a button in the toolbar is a good idea. May be we can just transform the circle in a more significant icon. Do you have any suggestion?

We can just change the icon to a home or kinda of a back button....

llaske:
Let's try a back button

s-mag:
> Let's try a back button

So should I choose a icon myself under free to use license and make a PR

llaske:
Try to the use the < icon (see in Calligra activity on dev branch)

s-mag:
> Try to the use the < icon (see in Calligra activity on dev branch)

I guess adding a button in toolbar would be better as the round button is drawn [here](https://github.com/llaske/sugarizer/blob/master/activities/Gridpaint.activity/js/gridpaint.js#L219).
What are your thoughts on this..
Or we can add the back button in upper left corner of the board...
It will be easily accessible.

llaske:
Let's try to put the back button in upper left corner

s-mag:
> Let's try to put the back button in upper left corner

Moreover there is a issue that if we click around the button it works....
![ezgif com-video-to-gif](https://user-images.githubusercontent.com/44548987/71664328-9ceb0a80-2d7e-11ea-81c6-e0a218ad0dc5.gif)

So it would  be nice to remove it.

llaske:
Fixed in https://github.com/llaske/sugarizer/pull/565

Issue #403, opened 2019-11-09
Gridpaint don't handle window resize


llaske:
Gridpaint activity don't handle window resize event. So, if the activity is open with a small window then resized, the board keep the same size.

![image](https://user-images.githubusercontent.com/981841/68525909-adc75480-02d6-11ea-977c-e1f2b01d164c.png)

llaske:
Fixed in https://github.com/llaske/sugarizer/commit/99ed20d71c4f18635efacf0d61b2da383fc28b76

PR #555, opened 2020-01-04
Fixed Issue #342 (GridPaint screen don't work properly on Firefox)

s-mag:
Now Grid Paint Activity screen will work properly on Firefox...
gp


s-mag:
@llaske Is this fine?

llaske:
It's a good start but scrollbar should not appear when the screen is resized.

image

And of course, the idea will be to fix #403 as well because both this issues forbid Gridpaint to be responsive.

s-mag:
Will do the changes and I will also make activity responsive...
Thanks

s-mag:
> It's a good start but scrollbar should not appear when the screen is resized.
> 
> And of course, the idea will be to fix #403 as well because both this issues forbid Gridpaint to be responsive.

Fixed Scrollbar...
Any suggestions how to make it responsive..
Tried many things but failed..

AndreaGon:
@s-mag Hi! You're doing great! As for responsiveness, you can try using media queries to detect when screen size is small.

s-mag:
> @s-mag Hi! You're doing great! As for responsiveness, you can try using media queries to detect when screen size is small.

@AndreaGon I used media queries but it seems that it only resizes the border of the frame not the contents of it...

s-mag:
The latest commit fixes issue #546.It adds a back button on top of color palette.

s-mag:
@llaske This looks fine and enhances the UI...

s-mag:
@llaske
Will it be fine if this PR fixes the other issues and not the responsiveness.
I will fix the responsiveness in other PR.
Can we do this???

llaske:
Yes you can do it in another PR because it's related to another issue: #403 

BTW there is an issue in your fix, a scrollbar now appear in Chrome and the background is not the same.

Here's a capture of your fix:

image

Here's a capture before your fix:

image

s-mag:
Oh looks like I changed the background colour. Will change that right away...

s-mag:
@llaske Latest commit fixes background color..

s-mag:
So this PR fixes two issues #342 & #546

llaske:
The back icon is nice.

BTW there is now an issue in Chrome. There is a shift between the location clicked and the triangle filled.

s-mag:
Will fix right away..

s-mag:
@llaske Fixed the issue mentioned..

llaske:
Hmmm. Look like there is also the same issue on Firefox now.

Peek 07-01-2020 22-35

s-mag:
@llaske Its working fine on my side...
Can you check it again??

s-mag:
Have a look at this...Working fine..
ezgif com-video-to-gif (2)

s-mag:
@llaske Have a look here. Its working fine. PS. Ignore my slow computer

Chrome
ezgif com-video-to-gif (3)

Firefox
ezgif com-video-to-gif (4)

s-mag:
@llaske Have a look here. Its working fine. PS. Ignore my slow computer

Chrome
ezgif com-video-to-gif (5)

Firefox
ezgif com-video-to-gif (4)

s-mag:
@llaske Can you review this PR as well.The issue is not coming on my PC.

llaske:
I've tried on another computer and I've got the issue on Firefox:

Peek 11-01-2020 16-15

Did you clear your browser cache?

s-mag:
@llaske I got the problem..
See here if the screen size is full it works fine
ezgif com-video-to-gif (7)

But if the screen size is made small then the problem arises.
ezgif com-video-to-gif (8)

Can you suggest a way to solve this issue. I tried many things but failed to do so...

llaske:
Everything happens in the `events.js` file and more specifically in the `localy`, `localx` methods [here](https://github.com/llaske/sugarizer/blob/dev/activities/Gridpaint.activity/js/events.js#L82).
You should have a detailed look on it and try to understand how it works and how calculation is done.

s-mag:
@llaske Is this approach fine

llaske:
Hmmm. I'm not in favor to add hard coded values in the source code.
In my opinion there is already too many hard coded values in the current source code of GridPaint activity! Ultimately I would love to have no (or at least less) hard coded values.
So it will be better to try to compute these values depending of the size of the screen. Specifically because the idea is later to fix https://github.com/llaske/sugarizer/issues/403 too.

Plus your current fix don't work for every window size. For example with a width of 755px or 1175px.

s-mag:
@llaske Can I know how did you calculate these values [here](https://github.com/llaske/sugarizer/blob/dev/activities/Gridpaint.activity/js/gridpaint.js#L126) so that I can calculate value for firefox...

PR #556, opened 2020-01-05
Added Fullscreen and Unfullscreen buttons in Tankop Activity

s-mag:
@llaske Made a new PR


llaske:
It's nice but the resize should be call when fullscreen is called else the board is not fully used.

s-mag:
> It's nice but the resize should be call when fullscreen is called else the board is not fully used.

@llaske that gives a error when called..
see here
Error2

llaske:
Should be probably called only when on the play screen.

s-mag:
> Should be probably called only when on the play screen.

If I call it in play mode it does not do anything..
Can you suggest any another way

s-mag:
Added back resize function

s-mag:
@llaske 
Will it be fine if this PR fixes the other issues and not the responsiveness.
I will fix the responsiveness in other PR.
Can we do this???

activities/TankOp.activity/js/activity.js
```diff
@@ -38,30 +38,19 @@ define(["sugar-web/activity/activity"], function (activity) {
...
-						} 
-					}
-
-					function zoomout(){
-						var x=document.getElementsByClassName("body");
-						for (var i = 0; i < x.length; i++) {
-							x[i].style.bottom="0px";
-						} 
-					}
+					play= new TankOp.Play();
```
llaske: Not a good idea to create a new Play object. Play object is the play screen so it should be created only when the play screen is displayed.

activities/TankOp.activity/js/activity.js
```diff
@@ -38,30 +38,19 @@ define(["sugar-web/activity/activity"], function (activity) {
...
 					// Full screen
 					document.getElementById("fullscreen-button").addEventListener('click', function() {
 						document.getElementById("main-toolbar").style.visibility = "hidden";
 						document.getElementById("unfullscreen-button").style.visibility = "visible";
-						document.getElementById("unfullscreen-button").style.top = "40px";
-						zoomin();
+						if (app.init())
```
llaske: Not a good idea to call init, it should be called once at the activity initialization.

s-mag:
@llaske can you suggest any another way to make it work...

activities/TankOp.activity/js/activity.js
s-mag: Then how should I check the play mode

llaske:
Look like the `TankOp.Play` object is already stored in a global variable `play` [here](https://github.com/llaske/sugarizer/blob/dev/activities/TankOp.activity/play.js#L59).
May be you could try to use it.

s-mag:
@llaske Fixed the errors and made changes...
Can you help me out to resize the screen as calling it does not resize it and at the same time no error is passed which resolves the changes requested..

llaske:
It's nice but there is still an error. 

```
play.js:173 Uncaught TypeError: Cannot read property 'id' of null
    at e.TankOp.Play.resize() [as resize] (play.js:173)
    at HTMLButtonElement.<anonymous> (activity.js:65)
TankOp.Play.resize() @ play.js:173
(anonymous) @ activity.js:65
```

Steps to reproduce:
- Click on fullscreen
- Click on start
- Click on home
- Click on unfullscreen

s-mag:
> It's nice but there is still an error.
> 
> ```
> play.js:173 Uncaught TypeError: Cannot read property 'id' of null
>     at e.TankOp.Play.resize() [as resize] (play.js:173)
>     at HTMLButtonElement.<anonymous> (activity.js:65)
> TankOp.Play.resize() @ play.js:173
> (anonymous) @ activity.js:65
> ```
> 
> Steps to reproduce:
> 
> * Click on fullscreen
> * Click on start
> * Click on home
> * Click on unfullscreen

Will fix the error. Is there anything else to do apart from this ??

s-mag:
@llaske Fixed these errors...

s-mag:
@llaske Is anything to be done for this issue??

llaske:
Just fix the changes requested and it will be okay.

s-mag:
I guess I am not getting what to change. You requested changes but the changes are not shown...

PR #557, opened 2020-01-05
Refactor code of presence popups in Memorize activity

sdziuda:
Currently I've replaced the code of presencepalette.js with the code from the same file from ActivityTemplate. I've also moved some of the deleted code into activity.js. I know that not everything is working fine. I will need some help with that because I have no clue what's working improperly. 

At this moment Join and Leave popups work fine. But the person who joins the host can't see the board.

### DIFF
--- a/activities/Memorize.activity/css/libnotify.css
+++ b/activities/Memorize.activity/css/libnotify.css
@@ -1,0 +1,104 @@
+html,
+body {
+  min-height: 100%;
+}
+.humane,
+.humane-libnotify {
+  position: fixed;
+  -moz-transition: all 0.3s ease-out;
+  -webkit-transition: all 0.3s ease-out;
+  -ms-transition: all 0.3s ease-out;
+  -o-transition: all 0.3s ease-out;
+  transition: all 0.3s ease-out;
+  z-index: 100000;
+  filter: progid:DXImageTransform.Microsoft.Alpha(Opacity=100);
+}
+.humane,
+.humane-libnotify {
+  font-family: Ubuntu, Arial, sans-serif;
+  text-align: center;
+  font-size: 15px;
+  bottom: 10px;
+  right: 10px;
+  opacity: 0;
+  width: 150px;
+  color: #fff;
+  padding: 10px;
+  background: #000;
+  -moz-transform: translateY(-40px);
+  -webkit-transform: translateY(-40px);
+  -ms-transform: translateY(-40px);
+  -o-transform: translateY(-40px);
+  transform: translateY(-40px);
+}
+.humane p,
+.humane-libnotify p,
+.humane ul,
+.humane-libnotify ul {
+  margin: 0;
+  padding: 0;
+}
+.humane ul,
+.humane-libnotify ul {
+  list-style: none;
+}
+.humane.humane-libnotify-info,
+.humane-libnotify.humane-libnotify-info {
+  background-image: url('data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAAEAAABQCAYAAADYxx/bAAAABmJLR0QA/wD/AP+gvaeTAAAAMUlEQVQYlWNgYDB6ysTAwMDAxMDACCcYUFkMDEwMDEwMBNVhkxg65jGhmke6M6hgHgBSdgHnpZwADwAAAABJRU5ErkJggg==');
+  background: -webkit-gradient(linear, left top, left bottom, color-stop(0, rgba(0,0,50,0.9)), color-stop(1, rgba(0,0,100,0.9))) no-repeat;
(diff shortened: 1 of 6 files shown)


sdziuda:
I've made the code be very similiar to the one you can find in MazeWeb activity. Anyway the problem with person not seeing the board is still live and I really have no clue where it comes from.

llaske:
I think you forget to call to `memorizeApp.onUsersListChanged` method.
It was called [here](https://github.com/llaske/sugarizer/pull/557/files#diff-a9819884ebd549090c650f4d206cc8b2L120) and it's not called anywhere in your new code.

sdziuda:
I've added this method but it still doesn't change anything. Therefore I think that it's not the problem with `memorizeApp.onUsersListChanged` not being called. 
I think that the problem occurs because the player who joins doesn't get the message with the content, what is really weird because as you can see on the screenshot it is [sent](https://github.com/llaske/sugarizer/blob/84d8392ac29bdfaa3e4f2738835af7c8baf2f0aa/activities/Memorize.activity/js/activity.js#L130) but not[ delivered](https://github.com/llaske/sugarizer/blob/84d8392ac29bdfaa3e4f2738835af7c8baf2f0aa/activities/Memorize.activity/js/activity.js#L192).
image

llaske:
Got it.
You send correctly the message but `sharedId` [here](https://github.com/llaske/sugarizer/pull/557/files#diff-a9819884ebd549090c650f4d206cc8b2R128) is `null` for the host. 
On the host, the `sharedId` is not received in the environment, it's build [here](https://github.com/llaske/sugarizer/pull/557/files#diff-a9819884ebd549090c650f4d206cc8b2R175) and you could get it by calling `presence.getSharedInfo().id`.

sdziuda:
Thank you for noticing this!
Now the messages are sent and delivered properly. Nevertheless there is a problem with `lzstring.js` file. The error like on the screenshot occurs.

image

Maybe it's because of the [content](https://github.com/llaske/sugarizer/blob/4313515dcac987f1a9180b60f7598d2e5254734c/activities/Memorize.activity/js/activity.js#L127) of the message?

llaske:
Look like you've sent [here](https://github.com/llaske/sugarizer/pull/557/files#diff-a9819884ebd549090c650f4d206cc8b2R130) a message non compressed.
You hate to compress the message using the `MemorizeApp.sendMessage` method [here](https://github.com/llaske/sugarizer/blob/4313515dcac987f1a9180b60f7598d2e5254734c/activities/Memorize.activity/js/memorize-app.js#L718).

sdziuda:
You were right, after adding the compression sending the message works fine. Now the non-host players can see the board. So we've got a little success!
But the next problem is that the `displayUsersAndScores` function doesn't work. From what I've observed it's probably because of the `MemorizeApp.game.players.length` is returning 0 right [here](https://github.com/llaske/sugarizer/blob/f8715c83b43b26463e791a82160074272b6ab2fa/activities/Memorize.activity/js/memorize-app.js#L734). Because of the same reason the non-host player can't see which card has been clicked, I think that code responsible for this is [here](https://github.com/llaske/sugarizer/blob/f8715c83b43b26463e791a82160074272b6ab2fa/activities/Memorize.activity/js/memorize-app.js#L710).
I thought that maybe it's because of the `onUsersListChanged` function not working properly [here](https://github.com/llaske/sugarizer/blob/f8715c83b43b26463e791a82160074272b6ab2fa/activities/Memorize.activity/js/activity.js#L145) so I've tried moving it back to the `presencepalette.js` file and it didn't give any other output. Therefore I think it's not a problem.

llaske:
I guess that most of problems come from the fact that you should yourself the `sendMessage` we mentioned earlier. 
Theoretically, you must not have to call it yourself. Everything (including handling players length) happens directly in the `MemorizeApp.onUsersListChanged` method.
Are you sure this method is still called?

sdziuda:
Yes, it is called, but maybe because it's just a function in `activity.js` instead of being a prototype from `presencepalette.js` it isn't working properly. From what I've noticed the `memorizeApp.onUsersListChanged()` isn't even accessed [here](https://github.com/llaske/sugarizer/blob/f8715c83b43b26463e791a82160074272b6ab2fa/activities/Memorize.activity/js/activity.js#L146) while running the code online.

llaske:
You're wrong, I've added a log message [here](https://github.com/llaske/sugarizer/blob/f8715c83b43b26463e791a82160074272b6ab2fa/activities/Memorize.activity/js/memorize-app.js#L819) and it's never called thought it was called in the initial source code.
I suggest you to remove your send call and try to understand how the message is never called.

llaske:
Fixed in https://github.com/llaske/sugarizer/pull/776

PR #558, opened 2020-01-06
adding toolbar button to display trails of cells deaths

prakashujjwal1010:
This PR
- adds toolbar button to display trails of cells deaths
- adds the state of button to datastore
- modified tutorial.js and locale.ini for new toolbar button 
- adds attribution for new icon used

Screenshot from 2020-01-07 00-28-17

Screenshot from 2020-01-07 00-28-20


llaske:
Very cool.

Two UI remarks:

- I think the death color could be brighter, may be a lighter gray or why not do a very light version of the stroke color
- I'm not a big fan of your icon: it's too big (icons should have margin between the border and the content) and it's not really explicit. May be you could find something around the concept of "footsteps"?

prakashujjwal1010:
- changed icon (also added the attribution)
Screenshot from 2020-01-07 23-38-08

Screenshot from 2020-01-07 23-32-30
Screenshot from 2020-01-07 23-36-49

i have not use lighter version of stroke color as if stoke color is like yellow or any other bright color, then we can have a problem where both are alive and dead cells are of same color.

llaske:
Good job. Thanks.

PR #560, opened 2020-01-09
Fixed Issue #517 (Can't see result in Calculate)

s-mag:
@llaske Fixed Issue #517. Deleted previous PR


llaske:
Hmmm. It don't fix the issue. It's not possible to scroll the screen to see the result zone.

image

s-mag:
@llaske It's working on my side....
ezgif com-video-to-gif (6)

s-mag:
It's strange that it is working here on my pc and not on your side

s-mag:
@llaske The scrollbar aappears if you have >3 results

llaske:
Ok. I've reproduce it.
That's the idea but the scrollbar should appear as soon as a part of the screen is not visible, not only when >3 results.
Plus, all the zone is not always scrollable. See the following sample:

Peek 11-01-2020 09-54

s-mag:
@llaske Fixed the scrollable area.

llaske:
Good.

Still some issues:

Could we avoid the scrollbar zone to the left when there is no need to scroll?

image

Could we avoid to have the scrollbar when it's not need?

Peek 11-01-2020 16-01

When a scrollbar is here, we can't scroll until the last result:

Peek 11-01-2020 16-04

s-mag:
@llaske Made the changes
1)Avoided the scrollbar when not required
2)The issue of not scrolling to the last result occurs when screen size is small.In this case Fullscreen button comes into account and clicking on that shows the last result
Capture

llaske:
Hmmm. Okay for 1) but not okay for 2).
We can't force user to click on fullscreen mode. Could you find a way to avoid this?

s-mag:
I have fixed the issue where the last result was not showing...

PR #565, opened 2020-01-16
Added Back Button in Gridpaint activity.

s-mag:
Fixed Issue #546 
@llaske I am making different PR for each Issue.
Closed previous PR.

Untitled


llaske:
Nice.

Could you update the tutorial too? It still mention the "green circle"

image

I suggest to add a popup to point the new back icon.

s-mag:
Oh I forgot that!! Will do that

s-mag:
@llaske I have updated localization for the new back icon. I cannot add popup for it because it is also drawn in the same way as the other buttons [here](https://github.com/llaske/sugarizer/pull/565/files#diff-f00bf3e76d44a27c767deacc7c4cbc1dR221).

llaske:
Nice. Thanks.

PR #567, opened 2020-01-18
Export ODT Improvements

ashish0910:
Since New Font families were added to the activity . This pull request enhances export to odt feature by adding these Font families to exported odt file as well .


llaske:
Good. Thanks.

PR #568, opened 2020-01-19
Fixed Issue #342 (GridPaint screen don't work properly on Firefox)

s-mag:


activities/Gridpaint.activity/js/gridpaint.js
```diff
@@ -119,21 +119,43 @@ function handleButton(y){
...
-	ctx.scale(scale, scale);
-	for(var i=0;i<shapes.length;i++){	
-		var low = hexdigit(i&0xf);
-		var mid = hexdigit((i>>4)&0xf);
-		var high = hexdigit((i>>8)&0xf);
-		ctx.fillStyle = '#'+high+'F'+mid+'F'+low+'F';
-		shapePath(i, ctx, true);
-	  ctx.fill();		
+var useragent = navigator.userAgent.toLowerCase();
+if (useragent.indexOf('chrome') == -1) {
```
llaske: Hmmm. Not a big fan of this idea to differentiate the whole process.

s-mag: Can I know how did you calculate the value for translate (140,20) So that I can calculate common.

llaske: Unfortunately I don't know. Was there from the beginning.

llaske:
Fixed in https://github.com/llaske/sugarizer/commit/99ed20d71c4f18635efacf0d61b2da383fc28b76

PR #570, opened 2020-01-20
fixes #556 responsiveness of TankOp activity

b18050:
Please check @llaske .


llaske:
It could be improved. Text still continue to overlap.
May be the background image size could be reduced too.

image

You must not send a PR on the master branch. Please use dev branch.

PR #571, opened 2020-01-21
fixes #556 responsiveness of TankOp activity

b18050:

PR #573, opened 2020-01-22
fixes #556

anshuldutt21:
Used media queries to make the activity more responsive.


llaske:
It don't solve the issue.

image

llaske:
Check https://github.com/llaske/sugarizer/pull/571
It has no sense to be two developers to work on the same issue at the same time.

anshuldutt21:
@llaske Actually I thought as given in the contributing guidelines that we shouldnt ask before working on the issue and so I didnt know someone else was also working on it. Sorry for that.  Regarding the Pull Request, Can you please explain as to what should  be done since I have tried to avoid overlapping of text as much as possible with the html code provided except for the mission completed text.

PR #577, opened 2020-01-25
fixes #559 ( Fototoon don't handle window resize )

### LINKED ISSUE #559: Fototoon don't handle window resize
Fototoon activity don't handle window resize event. So, if the activity is open with a small window then resized, the board keep the same size.

ft

anshuldutt21:
fixes #559

### DIFF
--- a/activities/Fototoon.activity/js/activity.js
+++ b/activities/Fototoon.activity/js/activity.js
@@ -39,8 +39,6 @@ define(["sugar-web/activity/activity","sugar-web/datastore","sugar-web/env","tex
         // remove 5 more to be sure no scrollbars are visible
         mainCanvas.height = window.innerHeight - sugarCellSize - 5;
         mainCanvas.width = mainCanvas.height * 4 / 3;
-
-        console.log(mainCanvas.width);
         mainCanvas.style.left = ((window.innerWidth - mainCanvas.width) / 2) + "px";
 
         var previousButton = document.getElementById("previous-button");
@@ -68,7 +66,7 @@ define(["sugar-web/activity/activity","sugar-web/datastore","sugar-web/env","tex
         toonModel.init();
         toonModel.attachPageCounterViewer(pageCounter);
         toonModel.attachPrevNextButtons(previousButton, nextButton);
-        
+
         var editMode = true;
 
         var addGlobeButton = document.getElementById("add-globe");
--- a/activities/Fototoon.activity/js/activity.js
+++ b/activities/Fototoon.activity/js/activity.js
@@ -68,6 +68,7 @@ define(["sugar-web/activity/activity","sugar-web/datastore","sugar-web/env","tex
         toonModel.init();
         toonModel.attachPageCounterViewer(pageCounter);
         toonModel.attachPrevNextButtons(previousButton, nextButton);
+        
         var editMode = true;
 
         var addGlobeButton = document.getElementById("add-globe");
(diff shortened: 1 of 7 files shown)


s-mag:
@anshuldutt21 It does not work as expected. The size should decrease on decreasing width and title box should also resize in the same proportion. See [here.](https://gfycat.com/coldcoarsearcticfox)

activities/Fototoon.activity/js/activity.js
```diff
@@ -39,6 +39,8 @@ define(["sugar-web/activity/activity","sugar-web/datastore","sugar-web/env","tex
         // remove 5 more to be sure no scrollbars are visible
         mainCanvas.height = window.innerHeight - sugarCellSize - 5;
         mainCanvas.width = mainCanvas.height * 4 / 3;
+
+        console.log(mainCanvas.width);
```
llaske: Dot not let a log message

activities/Fototoon.activity/js/activity.js
```diff
@@ -66,7 +68,6 @@ define(["sugar-web/activity/activity","sugar-web/datastore","sugar-web/env","tex
         toonModel.init();
         toonModel.attachPageCounterViewer(pageCounter);
         toonModel.attachPrevNextButtons(previousButton, nextButton);
-
```
llaske: No need to change that.

activities/Fototoon.activity/js/toon.js
```diff
@@ -822,7 +822,7 @@ define(["easel","sugar-web/datastore","sugar-web/env","webL10n","humane"], funct
             this._textView.lineWidth = this._globe._width * 2;
             this._textView.x = this._globe._x;
             this._textView.y = this._globe._y -
-                this._textView.getMeasuredHeight() / 2;
```
llaske: You don't need to change this file.

anshuldutt21:
Hi @llaske can you please review it once again. I have changed the proportion of title with respect to the canvas. The canvas size changes on resizing but the title size doesnt unless refreshed even when I call the init method. I think we can add the canvas resize feature only.

s-mag:
@anshuldutt Please remove log messages

s-mag:
@anshuldutt21 you should not use init in resize function. This resets the canvas or the title. See here:-

ezgif com-video-to-gif (11)

anshuldutt21:
@s-mag I cant remove the init function or else a blank canvas is shown. If I change the width and height of canvas I would have to pass it to the new Toon model. Can you suggest any method I could use to remove resetting.

s-mag:
@anshuldutt21 I am not sure but using media queries and scaling the canvas might help...

s-mag:
@anshuldutt21 Nice but it does not solve the issue completely... The title box should resize.

anshuldutt21:
@s-mag I have changed the width and height of the title box according to the canvas, but it is still not resizing. I think there is some other issue in it which works on reloading but not on resizing.

anshuldutt21:
@llaske Can you please review the PR

llaske:
It's a good start but I think it will be nice if when the window is resized the proportion (relative position) is keep.

image

anshuldutt21:
@llaske if I use the init method for resizing the title box , all the initial data corresponding to it will be lost. So I think the best way for now is to leave it like that because otherwise it will keep on reloading and all the data will be lost. Also, We could create another issue for the title resizing thing.

llaske:
@anshuldutt21  I'm not agree. I don't understand why resize necessarily need that all data will be lost. Currently, images are resize to fit the screen and no data are lost.

llaske:
Close due to lack of activity. @anshuldutt21 feel free to reopen it if need.

Issue #563, opened 2020-01-13
Font size of Title reduces in fototoon activity


sarthak-g:
Font size of Title( only after applying all three color, bold, italic) reduces in fototoon activity on adding objects from datastore. See below:
![ezgif com-video-to-gif(3)](https://user-images.githubusercontent.com/42293606/72277732-77041680-3658-11ea-9ec8-fc3d9b4f4119.gif)

s-mag:
Moreover the buttons does not work afterwards...
See here
![ezgif com-video-to-gif (9)](https://user-images.githubusercontent.com/44548987/72332532-3223c280-36e0-11ea-8bdb-2aabbb8139b5.gif)

sarthak-g:
> 
> 
> Moreover the buttons does not work afterwards...
Yeah, I forgot to mention that

llaske:
Interesting. Very strange.

anshuldutt21:
Hi , I would like to work on this issue.

llaske:
Fixed in https://github.com/llaske/sugarizer/pull/587

PR #578, opened 2020-01-26
Fixed #575 (Improvement of text palette)

### LINKED ISSUE #575: Improve design of Text Palette in Fototoon activity
The Bold and Italic buttons should get highlighted when selected so that the user knows that the text is selected bold or italics. Same can be done for the color codes. It can solved by doing minor changes in the toon.js file.

ezgif com-video-to-gif

sarthak-g:
fixed issue #575


sarthak-g:
I encounter 2 bug in this pr.
I will commit soon after upgrading.

llaske:
Hmmm. It look like it don't work when both Bold and Italic is selected.
See the video below:

Peek 26-01-2020 16-53

Could be related to https://github.com/llaske/sugarizer/issues/563

sarthak-g:
Yeah, it's because of #563 
In video below, I printed whether box selected has text in italic or normal on first line and similarly bold or normal on second line.
When we come back to first page then only bold property is present.
ezgif com-video-to-gif(4)

llaske:
Nice. Thanks.

PR #586, opened 2020-02-06
Made improvement in sugarizer spiral.

### LINKED ISSUE #579: Improve Sugarizer spiral in home view
I think The activity icons on the main page are a little bit packed between 2 different radiuses. They need to be seperated a little bit to give the spiral feels which can be accomplished by improving the css of the main panel. Also the journal icon must be a little above and should look different than the usual activities.
Screenshot from 2020-01-26 17-07-45

Swayamsvk:
- fixes #579 

- Sugariser spiral improved.
Screenshot from 2020-02-06 20-45-14


llaske:
I don't see where is the improvement here: it take more place and don't give more visibility to icons.

Swayamsvk:
Actually I have increased the space between two radiuses. Should I increase the space between the two radiuses So that they look more distinct?

llaske:
You should try to reproduce the way Sugar work. Just increase spacing will not change everything.

Please, write correctly the name of the project.

Swayamsvk:
Sorry @llaske for the spelling mistake. I will try to make the changes according to you. Thank you.

Swayamsvk:
@llaske actually I can't understand how the spacing between different activities is being manipulated can you please help me?

Swayamsvk:
@llaske actually I am having some problems in manipulating the space between different activities can you help me by at least specifying where it is being manipulated I can do the rest. Thank you

llaske:
@Swayamsvk everything is in the draw method [here](https://github.com/llaske/sugarizer/blob/dev/js/homeview.js#L263). Icon size and padding then distribution of icons in the spiral are computed in this method.

Swayamsvk:
Screenshot from 2020-03-11 00-19-42

Swayamsvk:
@llaske I think I have fixed this issue please review this PR.

llaske:
Hmmm. Shortly you've just:
* Reduce the icon size to 45 instead of 55
* Increase the space factor to 1.41 instead of 1.21
* Add 25 to the radius 

I'm not in favor of reducing the icon size because it should be large enough to be clickable by hand.
Plus it don't solve the issue that the spiral is not smooth.

The better solution should be to compute a real [Archimedean spiral](https://en.wikipedia.org/wiki/Archimedean_spiral).

Swayamsvk:
Ok, I will work on it.

Swayamsvk:
@llaske So, basically I have to increase the spacing between the activity icons and make it a proper archimedean spiral so that the icons look more detailed and organised

llaske:
@Swayamsvk yes that's the idea.

Swayamsvk:
Screenshot from 2020-03-13 03-41-09

Swayamsvk:
How about this @llaske ?

llaske:
It's not an archimedean spiral: look the height gap between Record activity icon and Abecedarium icon.
You can't fix this issue without updating the current algorithm used to compute icon position and replace it by the archimedean spiral formulae.

llaske:
Fixed in https://github.com/llaske/sugarizer/commit/e24a3d80fb23dbb06fb1e8fd8a25e9c929eaa139

PR #591, opened 2020-02-10
adding onNext in last step of tutorial of color my world activity

prakashujjwal1010:
### before:
ezgif com-video-to-gif (7)

### after:
ezgif com-video-to-gif (8)


llaske:
In some cases, the Next button could be the same than the End button because there is some other steps depending of the visibility of some UI elements. See for example on the Sugarizer first screen:

image

So the fix will force us to handle specifically onNext depending of the context (is it the true last or is the supposed last step).
I think it's not worth the effort for something that is not really a bug.

prakashujjwal1010:
Okay, I see.

PR #607, opened 2020-02-14
Add fullscreen/unfullscreen mode for Stopwatch Activity

### LINKED ISSUE #599: Add a fullscreen button in the Stopwatch activity
Sugarizer activities don't look good on devices with small screen - for example smartphones.
On these devices, the toolbar could sometimes take 1/4 of the screen.

To increase the size for activities contents, the idea is to add a fullscreen button in the toolbar  like in the QRCode activity. When clicked, the fullscreen button hide the toolbar and display only an unfullscreen button. When clicked, the unfullscreen button show again the toolbar.

This features is implemented here: https://github.com/llaske/sugarizer/blob/dev/activities/QRCode.activity/js/activity.js#L144-L160

The objective of this issue is to implement the same feature in the StopWatch activity.

Thinks to check:

* [ ] The fullscreen button should be on the right of the toolbar, near the Stop button.
* [ ] The fullscreen button should be always visible, including in small screen size. Could require to adapt media queries in CSS.
* [ ] The size of the board should be increase in fullscreen mode. So it means that the resize event should be call directly or indirectly when clicking on fullscreen/unfullscreen button.
* [ ] The unfullscreen button should be visible only when the screen is in fullscreen mode.
* [ ] The unfullscreen button should be attach to the right of the screen. It's better if it's not at the same location than the Stop button to avoid side effect on multiple click.

AndreaGon:
This should fix #599 

stopwatch


activities/Stopwatch.activity/js/activity.js
```diff
@@ -243,5 +243,23 @@ define(["sugar-web/activity/activity","mustache", "sugar-web/env"], function (ac
         activity.getDatastoreObject().save();
 
     });
+    
+    // Switch to full screen when the full screen button is pressed
+  	document.getElementById("fullscreen-button").addEventListener('click', function() {
```
llaske: Indentation issue: you've mixed TAB and SPACE. In this file it should be SPACE only.

llaske:
Good job. Thanks.

PR #608, opened 2020-02-14
Added a fullscreen button in GridPaint activity

### LINKED ISSUE #601: Add a fullscreen button in the GridPaint activity
Sugarizer activities don't look good on devices with small screen - for example smartphones.
On these devices, the toolbar could sometimes take 1/4 of the screen.

To increase the size for activities contents, the idea is to add a fullscreen button in the toolbar  like in the QRCode activity. When clicked, the fullscreen button hide the toolbar and display only an unfullscreen button. When clicked, the unfullscreen button show again the toolbar.

This features is implemented here: https://github.com/llaske/sugarizer/blob/dev/activities/QRCode.activity/js/activity.js#L144-L160

The objective of this issue is to implement the same feature in the GridPaint activity.

Thinks to check:

* [ ] The fullscreen button should be on the right of the toolbar, near the Stop button.
* [ ] The fullscreen button should be always visible, including in small screen size. Could require to adapt media queries in CSS.
* [ ] The size of the board should be increase in fullscreen mode. So it means that the resize event should be call directly or indirectly when clicking on fullscreen/unfullscreen button.
* [ ] The unfullscreen button should be visible only when the screen is in fullscreen mode.
* [ ] The unfullscreen button should be attach to the right of the screen. It's better if it's not at the same location than the Stop button to avoid side effect on multiple click.

Swayamsvk:
fixed #601 
grid3 mkv


llaske:
It's a good start but you should also resize the board like when the window is resized.
Else the new space is not used.

Swayamsvk:
Ok I will work on it.

llaske:
Please do not mix several fix on the same PR.
Do a PR for each different fix: here one for Gridpaint and another one for typos.

Swayamsvk:
Sorry,my mistake.

Swayamsvk:
Is it necessary to delete these unnecessary commits or I should continue with them?

llaske:
I suggest you to close this PR and reopen distinct PR for each issue

Issue #600, opened 2020-02-11
Add a fullscreen button to the Video Viewer activity


llaske:
Sugarizer activities don't look good on devices with small screen - for example smartphones.
On these devices, the toolbar could sometimes take 1/4 of the screen.

To increase the size for activities contents, the idea is to add a fullscreen button in the toolbar  like in the QRCode activity. When clicked, the fullscreen button hide the toolbar and display only an unfullscreen button. When clicked, the unfullscreen button show again the toolbar.

This features is implemented here: https://github.com/llaske/sugarizer/blob/dev/activities/QRCode.activity/js/activity.js#L144-L160

The objective of this issue is to implement the same feature in the Video Viewer activity.

Thinks to check:

* [ ] The fullscreen button should be on the right of the toolbar, near the Stop button.
* [ ] The fullscreen button should be always visible, including in small screen size. Could require to adapt media queries in CSS.
* [ ] The size of the board should be increase in fullscreen mode. So it means that the resize event should be call directly or indirectly when clicking on fullscreen/unfullscreen button.
* [ ] The unfullscreen button should be visible only when the screen is in fullscreen mode.
* [ ] The unfullscreen button should be attach to the right of the screen. It's better if it's not at the same location than the Stop button to avoid side effect on multiple click.

ksraj123:
Fixed in #629

PR #611, opened 2020-02-15
Added a fullscreen button to the Video Viewer activity

s-mag:
Fixed Issue #600 

@llaske I have used the resize function but it does not increases the size of the content. It increases the size of the canvas. Should I create my function to increase the content size or its fine in this way??

- [x] The fullscreen button should be on the right of the toolbar, near the Stop button.
- [x]  The fullscreen button should be always visible, including in small screen size. Could require to adapt media queries in CSS.
- [x]  The size of the board should be increase in fullscreen mode. So it means that the resize event should be call directly or indirectly when clicking on fullscreen/unfullscreen button.
- [x]  The unfullscreen button should be visible only when the screen is in fullscreen mode.
- [x]  The unfullscreen button should be attach to the right of the screen. It's better if it's not at the same location than the Stop button to avoid side effect on multiple click.


llaske:
Good job.

The problem comes from the `computeSize` method and more specifically from [here](https://github.com/llaske/sugarizer/blob/dev/activities/VideoViewer.activity/app.js#L81).
The `*2` in this line is to reserve both the height for the toolbar and the height of the footer bar. 

So when the height of the toolbar is 0 (in fullscreen mode), the height of the footer toolbar is considered to be 0 too. It's not correct. I guess you could fix it by considering than the height of the footer toolbar is always 55px like mentioned [here](https://github.com/llaske/sugarizer/blob/dev/activities/VideoViewer.activity/css/activity.css#L60).

s-mag:
@llaske I guess the height of the footer is always 55 px its just that on calling app.resize method the height of the canvas changes. I guess making the content of the activity which includes the list of videos. There size should increase to use the new space not the empty body which just shifts the footer down. What do u think??

llaske:
Yes, you should still call resize but you should fix the size of the footer toolbar

s-mag:
Anyone interested can fix the related issue. My exams are starting and I will be back by 21st March.

ksraj123:
@llaske I have opened a new PR https://github.com/llaske/sugarizer/pull/629 implementing the suggestions provided as the original author @s-mag is not available.

PR #619, opened 2020-02-19
Add fullscreen/unfullscreen mode to FoodChain Activity

AndreaGon:
FoodChain activity should now enter fullscreen/unfullscreen mode, as said in #598


llaske:
Good job. Thanks.

PR #625, opened 2020-02-21
Reset functionality fixed in Stopwatch activity

### LINKED ISSUE #624: Time laps in stopwatch activity should be reset when reset button is clicked
The time-lapse feature in the stopwatch activity should save the time-lapse in a  continuous manner without any interruptions. But currently, as the reset button exists it creates a problem. The reset button is responsible to reset the time in the middle of a time-lapse
Screenshot from 2020-02-21 18-30-13
 and it may create a problem if it is being used to record time-lapse in an actual event.

abhishektanwar:
Fixes #624 

- Fixed reset of running time and splits(mark) upon clicking reset button
- Reset button cannot be clicked until the stopwatch is in pause state like [here](https://www.timeanddate.com/stopwatch/)

stopwatch_fix_1

- Improved UI/UX of stopwatch activity by adding hover effect on Play,Stop,Reset,Remove,Add Stopwatch buttons  
stopwatch_fix_2


llaske:
Nice. Thanks.

PR #636, opened 2020-02-22
tutorial with localization added to stopwatch activity

### LINKED ISSUE #596: Add a tutorial to the Stopwatch activity
Add a tutorial to the Stopwatch activity following tutorial 9 in Sugarizer development tutorial: https://github.com/llaske/sugarizer/blob/dev/docs/tutorial_step9.md

The tutorial:

* [ ] Should display a welcome message to explain the activity (could reuse the help message describing the activity in Sugarizer activities List view)
* [ ] Should explain usage of each button in the toolbar
* [ ] Should explain usage of each element in the UI
* [ ] Should be localized
* [ ] Should respect the Sugarizer UI, as explained in tutorial

ksraj123:
fixes #596 

- [x] Should display a welcome message to explain the activity (could reuse the help message describing the activity in Sugarizer activities List view)
- [x] Should explain usage of each button in the toolbar
- [x] Should explain usage of each element in the UI
- [x] Should be localized
- [x] Should respect the Sugarizer UI, as explained in tutorial

stopwatch tutorial


Swayamsvk:
I don't think this fix supports other languages.

ksraj123:
@Swayamsvk working on it current

activities/Stopwatch.activity/lib/sugar-web/activity/activity.js
```diff
@@ -98,10 +98,13 @@ define(["webL10n",
         shortcut.add("Ctrl", "Q", this.close);
 
         env.getEnvironment(function (error, environment) {
-            var l10n ={"en":"{{name}} Activity","fr":"Activité {{name}}","es":"Actividad {{name}}","pt":"{{name}} Atividade","de":"Aktivität {{name}}"};
+            var localization ={"en":"{{name}} Activity","fr":"Activité {{name}}","es":"Actividad {{name}}","pt":"{{name}} Atividade","de":"Aktivität {{name}}"};
```
llaske: You're not supposed to change this file. It's a standard library.

activities/Stopwatch.activity/lib/tutorial.js
```diff
@@ -0,0 +1,127 @@
...
+				title: l10n.get("TutoExplainTitle"),
+				content: l10n.get("TutoExplainContent")
+			},
+			{
+				element: "#fullscreen-button",
+				placement: "bottom",
+				title: l10n.get("TutoFullScreenButtonTitle"),
+				content: l10n.get("TutoFullScreenButtonContent")
+			},
+            {
```
llaske: Avoid mixing indentation style in the same file.

ksraj123: Indention Issue Fixed.

llaske:
That's fine but something is wrong regarding the font/style used by popup. See for example this comparison with another tutorial:

image

Plus, could you reorder items in steps array? The idea is too put at first the most important. So I guess it could be: start button, reinit, time laps, counter, remove button, add button, ...

ksraj123:
Issues fixed based on suggestions

- Font issue in tutorial fixed. Unlike other activities, this one was using bootstrap already and `bootstrap.min.css` and `bootstrap-tour-standalone.min.css` had conflicting properties for some classes. I guess this was causing the issue.
- Reordered tutorials popups.
- Pop up for Add Stopwatch button made to appear on top if it has little space below it. The popup overlaps with the button when all stopwatches are removed if it is fixed to top so conditional logic is needed.

stoptut

abhishektanwar:
@ksraj123 I believe the styling problem still persist in the popup .

ksraj123:
@abhishektanwar I think bootstrap was overriding the line-height of popup title as well, fixed that. Apart from that they look the same to my eyes.

compare

llaske:
Hmm I suggest to try to use bootstrap-tour.min.css instead of bootstrap-tour-standalone.min.css because the later already include bootstrap.min.css. It could the reason of the conflict.

Could you also solve conflict related to PR merged since your last commit?

ksraj123:
`boostrap-tour-standalone.css` and `bootstrap-tour-standalone.js` replaced with `bootstrap-tour.css` and `bootstrap-tour.js` respectively of the same version of bootstrap tour i.e. `v10.2`. `bootstrap-tour-standalone` used in this project and others which have same styling included `v3.1.0` of bootstrap but this project was already using `v3.3.4`. Now since `bootstrap-tour.css` does not have its own appropriate copy of bootstrap issues become apparent again.

llaske:
Nice. The UI issue is now fixed.

BTW the localization doesn't work: the Sugarizer language settings is not used.
It's probably related to the fact that it's not initialized in the activity. See [here](https://github.com/llaske/sugarizer/blob/dev/docs/tutorial_step5.md#initialize-localization) for more.

ksraj123:
@llaske Suggested change made, Localisation initialised.
localization added to stop watch

Issue #592, opened 2020-02-10
Develop a new chess activity


llaske:
> This issue is a challenge for students to the [GSoC 2020](https://github.com/sugarlabs/GSoC/blob/master/Ideas-2020.md). Every student interested to be candidate to GSoC 2020 Sugarizer projects must complete this task.

Develop **a new chess activity** for Sugarizer.

![image](https://user-images.githubusercontent.com/981841/74162843-b9058580-4c21-11ea-89ee-972268e78f79.png)

Detailed features expected:
- [ ] The activity should be playable against computer or against another user on the network (using presence).
- [ ] The activity should display the list of move during play.
- [ ] The UI of the activity should be simple, self-explanatory and respect Sugar standards.
- [ ] The activity should be correctly integrated into Sugarizer and use buddy colors when possible.
- [ ] The activity should be integrated with Journal (save its context in the Journal).
- [ ] The activity should use localization.
- [ ] The activity should include a tutorial to explain the UI and an interactive tutorial to explain how each pieces move on a chessboard.
- [ ] The activity should be responsive (work on any screen size) and should include a fullscreen button.
- [ ] The activity should work on any browser (Chrome, Firefox and Safari) and in Electron.
- [ ] The activity should not depend of an Internet connection (use local content only).
- [ ] The activity should work with mouse, touch and possibly keyboard.
- [ ] Every content/library used in the activity should rely on a Free Libre Open Source license.

Technically speaking, the activity could be written using pure JavaScript or using Vue.js inline (as in the [Calligra activity](https://github.com/llaske/sugarizer/tree/dev/activities/Calligra.activity)). JavaScript ES6 syntax should be avoided.

The activity should be based on [Chessboard JS](https://chessboardjs.com/) for the UI  and  [P4wn](https://github.com/douglasbagnall/p4wn) for the engine.

sarthak-g:
I've two doubts:
1. "pawn promotes to" and "computer level" feature should be there or we can skip them using their default value?
2. Computer playing for both sides should be allowed?

ashish0910:
1. Pawn should promote on reaching the end like in normal chess game . 
2. No I don't think we need computer vs computer . There should be computer vs Player or Player vs Player .

Hemansh31:
can we use libraries other than P4wn for the engine?

llaske:
[...]

PR #644, opened 2020-02-24
Chessboard Task for GSoC 2020

Hemansh31:
Initial commit for the Chessboard Task Challenge #592 for GSoC 2020


ashish0910:
@Hemansh31 Looks Promising 
Few Remarks :
1. On playing smart mode the game hangs for few seconds maybe because of large calculations ( does not happen in any other mode though ) . 
2.May be we can show list of moves .

Hemansh31:
@ashish0910 First of all Thanks for the remarks.
Secondly, when you say show the list of moves what is it that you expect because I have already shown all possible moves of a piece on mouse hover, is that not what you mean? Also yeah the smart mode is supposed to be the slowest as it is brute force backtracking without any alpha beta pruning.

llaske:
Your PR contains content not related to this PR (Chat activity).
Please do one branch and one PR for each issue.

Hemansh31:
@llaske I added that file by mistake but I removed it. Thanks for the remarks

ashish0910:
> @ashish0910 First of all Thanks for the remarks.
> Secondly, when you say show the list of moves what is it that you expect because I have already shown all possible moves of a piece on mouse hover, is that not what you mean? Also yeah the smart mode is supposed to be the slowest as it is brute force backtracking without any alpha beta pruning.

By List of moves I mean maintaining a log of moves played by user

llaske:
Good start. I specifically like your tutorial.

My remarks:

* The activity is not responsive
* Will be better to use a palette to select computer level
* You could use buddy color preference for background color
* Will be better to display a popup when an user join/leave the game (see in Memorize or Paint activity)
* Will be better to display buddy icon colors for opponent (see in Memory or LOL activity)
* The network icon should to the left, just after the activity palette
* The activity palette is empty when open, it should display the activity name by default
* Will be better to use an icon with a king instead of a pawn for the activity
* The message at the end of the game should not be a simple Javascript popup

llaske:
Unfinished

Issue #646, opened 2020-02-25
Journal functions not visible on mobile


dhruvmisra:
After selecting an activity, the Erase activity function is not visible on mobile devices including iPhone 6/7/8 Plus. (Also the 'Copy to your device' feature in smaller devices)
This affects major functionalities as the user might not be able to delete unnecessary activity instances.

iPhone 6/7/8 Plus:
![Annotation 2020-02-26 001405](https://user-images.githubusercontent.com/40134655/75278130-5b7f5480-582f-11ea-92ce-9d8e9593e319.jpg)

Pixel 2 XL:
![Annotation 2020-02-26 001613](https://user-images.githubusercontent.com/40134655/75278420-d7799c80-582f-11ea-9c5f-3ef7c213a051.jpg)

dhruvmisra:
I have solved this issue in my pull request #647.

llaske:
You're not in normal condition: https://github.com/llaske/sugarizer/issues/422

Issue #422, opened 2019-11-13
Gear Activity is not responsive


sarthak-g:
![Screenshot from 2019-11-13 22-48-34](https://user-images.githubusercontent.com/42293606/68787973-7eaf3b00-0668-11ea-93ba-94049e108ab4.png)

Gear Activity is not responsive on smart phone. Neither complete image nor "stop" button is visible.

llaske:
Thanks for this feedback.

Here's few insights regarding Sugarizer responsive policy:

- The preferred way to run Sugarizer on mobile devices is via the iOS and Android app.
- Sugarizer UI is adapted to screen larger than longer (landscape), this rule is forced on iOS and Android apps. In browser in portrait mode (like in your screen capture), nothing is required from activities today.
- Sugarizer is not adapted today to smartphone, it's not only related to toolbar but also to the size of graphical UI items in activities (too small to be handle with fingers). Sugarizer activities running on smartphone should do the best to run: minimal behavior is to be able to be launched and stopped. Nothing more is required. And of course, due to 1., this requirement is in landscape mode. A current evolution is to add a fullscreen button in activities. See open issues.

sarthak-g:
Thanks for sharing insights regarding Sugarizer responsive policy.

PR #647, opened 2020-02-25
[Issue #646 fix] Made Journal functions visible on mobile

dhruvmisra:
The Erase function will now be visible on mobile devices. The changes are localized to CSS classes related to journal and toolbar.

iPhone 6/7/8 Plus:
Annotation 2020-02-26 002409

Additionally, I fixed the overlap of search bar with the Favorites view button on the Home view on small devices.

Before:
before

After:
after


llaske:
You're not in normal condition. See https://github.com/llaske/sugarizer/issues/422

llaske:
Not an issue.

Issue #649, opened 2020-02-25
Erratic sizing of the Calculator


dhruvmisra:
The calculator displays irregular responsiveness for different screen sizes, filling only half the screen on mobiles. Makes the calculator harder to operate and see results.

![before](https://user-images.githubusercontent.com/40134655/75288083-97231a00-5841-11ea-9f70-00fee97764e5.gif)

llaske:
You're not in normal condition. See https://github.com/llaske/sugarizer/issues/422

s-mag:
@dhruvmisra Sugarizer activities are not meant or designed to work in small screens. See #422 
There is no need to correct all these activities as no one will use them in such a small screen size.These activities work perfectly in normal conditions...

PR #650, opened 2020-02-25
[Issue #649 fix] Fixed the erratic behavior of Calculator

dhruvmisra:
The issue has been fixed. I have also made the history scrolling independent of the calculator, providing perhaps a better user experience.

after


s-mag:
@dhruvmisra This will not work when the height is less... You can do one thing- Just make the toolbar responsive which will solve the issue...

activities/Stopwatch.activity/css/activity.css
```diff
@@ -144,4 +144,4 @@ button.remove {
 
 .buttons-group button.clear-marks-button {
   background-image: url(../icons/edit-clear.svg);
-}
```
s-mag: No need to make a change where not required.

activities/Stopwatch.activity/js/activity.js
```diff
@@ -263,4 +263,4 @@ define(["sugar-web/activity/activity","mustache", "sugar-web/env"], function (ac
 
     });
 
-});
```
s-mag: No need to make a change where not required.

dhruvmisra:
@s-mag Thank you for the review. The actual issue I was tackling here was the irregular resizing of the calculator container (Issue #649). I have created a PR which makes the toolbar responsive for Home views. (PR #647) Kindly review the same and I shall implement it for other activities too.

s-mag:
Sugarizer activities are not meant or designed to work in small screens. See #422
There is no need to correct all these activities as no one will use them in such a small screen size.These activities work perfectly in normal conditions...

activities/Stopwatch.activity/css/activity.css
dhruvmisra: There was only a removal of new line character. Fixed it.

activities/Stopwatch.activity/js/activity.js
dhruvmisra: There was only a removal of new line character. Fixed it.

llaske:
Not an issue.

PR #659, opened 2020-03-04
Chess Activity for GSoC 2020 [#592]

dhruvmisra:
This is my solution for #592 

- [x] The activity should be playable against computer or against another user on the network (using presence).
- [x] The activity should display the list of move during play.
- [x] The UI of the activity should be simple, self-explanatory and respect Sugar standards.
- [x] The activity should be correctly integrated into Sugarizer and use buddy colors when possible.
- [x] The activity should be integrated with Journal (save its context in the Journal).
- [x] The activity should use localization.
- [x] The activity should include a tutorial to explain the UI and an interactive tutorial to explain how each pieces move on a chessboard.
- [x] The activity should be responsive (work on any screen size) and should include a fullscreen button.
- [x] The activity should work on any browser (Chrome, Firefox and Safari) and in Electron.
- [x] The activity should not depend of an Internet connection (use local content only).
- [x] The activity should work with mouse, touch.
- [x] Every content/library used in the activity should rely on a Free Libre Open Source license.

UI: Chessboard.js
Engine: p4wn.js

Please review @llaske and @ashish0910. Would be happy to work on any changes you require.


llaske:
Very good. The UI is very Sugar friendly.
Few remarks:
* Will be nice to add columns and lines name
* Will be nice to use buddy background color for window background instead of white
* The rules could be mixed with the help
* Will be better to avoid text, you could replace Opponent information line by icon (buddy icon with the right color, robot, pawn white or black)
* Same for XXX to move, could just be <pawn color>

dhruvmisra:
Thanks a lot @ashish0910  and @llaske for your reviews.
I have resolved the conflicts in PR.
Would surely work on these improvements next!

dhruvmisra:
Made the suggested improvements:
- [x] Added column and line names
- [x] Buddy color in background
- [x] Rules start right after Help
- [x] Replaced text with icons at most places.

Please review @llaske and @ashish0910.

llaske:
Very cool. 

Few remarks:
- There is no way to switch color at start
- The game status box is too close of the chessboard, adding a margin will be better
- The current difficulty level should be selected in the palette so you could remove it from the game status box (the less text is displayed the better is)
- Instead of "You"/"Opponent" in the game status box, use the players name
- By convention the presence palette should be the first one to the right of the activity palette
- The rules of game is displayed at the end of tutorial and it's good. So the specific button is the toolbar to learn rules could be removed

llaske:
Thanks for your nice implementation of the Chess activity.
It show your skill on Sugarizer activity development.
BTW we've decided that the winner of the challenge is https://github.com/llaske/sugarizer/pull/702
Feel free to propose improvements or to send PR on the activity.
Thanks again for your work.

PR #660, opened 2020-03-05
Added Search functionality in Chat

dhruvmisra:
Added search button to the Chat Activity making it easier to look for any message in a long conversation.

chatSearch

Please review @llaske and @ashish0910

### DIFF
--- a/activities/ChatPrototype.activity/css/activity.css
+++ b/activities/ChatPrototype.activity/css/activity.css
@@ -11,9 +11,24 @@ background-image:url(../icons/image-upload.svg);
 visibility: hidden;
 }
 
+#main-toolbar #find-button {
+background-image:url(../icons/find.svg);
+visibility: hidden;
+}
+
+#find {
+  height: 30px;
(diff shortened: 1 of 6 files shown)


llaske:
Thanks for this proposal but I don't think it's useful. No one asked for it and it's not include in the original Chat activity neither.
It could be a valid feature for a real chat activity but because Sugarizer target children, it's better to limit features to the minimum to avoid a too complex UI.

quozl:
I agree that the feature is unnecessary for children.  I wish I'd had your attention on https://github.com/sugarlabs/chat/issues/18 and https://github.com/sugarlabs/chat/pull/19 so we could have had this discussion earlier.  It was part of the original feature set proposed, but was not implemented until last year in Sugar.

dhruvmisra:
I agree @llaske. I wanted to bring this version of the activity at par with the python version by implementing the search. Thanks @quozl for the heads up on earlier discussions regarding this.

Issue #668, opened 2020-03-06
Issues in game of life activity


ksraj123:
Here are some issues found in game of life activity
- Issues with Tutorial
    - Tutorial backdrop disappears for some steps, the steps are supposed to point at the game board
![bug game of life activity backdrop dissappears in tutorial](https://user-images.githubusercontent.com/39027928/76117277-84fb6580-6011-11ea-9125-efb73e2b5fa3.gif)

- Issues with responsiveness of the activity, on mobile resolutions - 

    - The help button disappears
    - The canvas element overflows over the Generation count element.
![help button hides and the board comes over score](https://user-images.githubusercontent.com/39027928/76117931-d0624380-6012-11ea-8b3d-3b957c4a2228.png)

I am working on the issues will open associated PRs soon.

ksraj123:
@llaske Issue with tutorial fixed in #669
![Peek 2020-03-07 02-32](https://user-images.githubusercontent.com/39027928/76122867-26d47f80-601d-11ea-9bfb-cc5b5dbfd6a0.gif)

llaske:
Please fill an issue with several problems even if it's related to the same activity.

I don't see an issue regarding the tutorial. It's not a problem that the step doesn't point to the board. It's perfectly clear that the popup is related to the board.

I don't see an issue regarding disappearance of the help button on small screens. On small screen some compromises should be done. The disappearance of the help button is a good one.

I'm agree regarding the overflow of generation number. It could be nice to propose something better.

PR #669, opened 2020-03-06
Fixing issues in tutorial of game of life activity

ksraj123:
fixes issues with tutorial of game of life activity mentioned in #668

### DIFF
--- a/activities/GameOfLife.activity/lib/tutorial.js
+++ b/activities/GameOfLife.activity/lib/tutorial.js
@@ -67,31 +67,31 @@ define(["webL10n"], function (l10n) {
 				content: l10n.get("TutoExplainContent")
 			},
 			{
-				element: "#canvas",
+				element: "canvas",
 				placement: "top",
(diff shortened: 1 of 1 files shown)


ksraj123:
Issue with backdrop fixed but the popups with images do no point at the right position. They should point that the canvas element but they overflow over it.
Peek 2020-03-07 02-14

ksraj123:
I think the reason for the overflow issue the popup over canvas element was that the properties of the div containing the image were being changed with javascript in the `onShown` method which is executed after the step is shown on the screen. I guess the coordinates for the popup to point to are calculated by bootstrap-tour before the step is shown depending upon its layout and the target, dynamically changing layout after popup has rendered already caused the issue.

The issue is fixed in https://github.com/llaske/sugarizer/pull/669/commits/62afa4e257e152d224e20c5da948fdf6a5a8259a
Peek 2020-03-07 02-32

llaske:
Not an issue.

PR #674, opened 2020-03-07
[Issue #673 fix] Fixed opening links in Markdown

### LINKED ISSUE #673: Markdown link opens in frame which crashes the output
If a link is clicked in output, it tries to open in the same frame, which crashes the output for rest of the activity duration.

markdownLinkIssue

I'll open a PR fixing this.

dhruvmisra:
Fixes #673. The links will now open in a new tab.

markdownLinkFix


llaske:
Good. Thanks.

PR #682, opened 2020-03-09
Challenge GSoc 2020

Swayamsvk:
Completed #592 for GSOC 2020.
Please review my PR @llaske  and @ashish0910 . Please let me know If if any modifications are required. I will fix them soon.


Swayamsvk:
Actually, I am interested in the idea "Sugarizer knowledge activity pack" and I have started writing the proposal for the same. @llaske I want to discuss the proposal with you. How should I make contact. Please let me know.

llaske:
It's a good start.

My remarks:

- [ ] The toolbar is resized when the window is resized
- [ ] The activity should be integrated with Journal (save its context in the Journal).
- [ ] The activity should use localization.
- [ ] The activity should include an interactive tutorial to explain how each pieces move on a chessboard.
- [ ] The activity should work with touch and possibly keyboard
- [ ] Will be better to use a palette to select computer level
- [ ] You could use buddy color preference for background color
- [ ] Will be better to display a popup when an user join/leave the game (see in Memorize or Paint activity)
- [ ] Will be better to display buddy icon colors for opponent (see in Memory or LOL activity)
- [ ] Will be better to use an icon with a king instead of a pawn for the activity
- [ ] The start button should be in the toolbar
- [ ] Miss a button to switch player

Swayamsvk:
Thanks @llaske for your review I will make the changes soon.

llaske:
Unfinished

Issue #205, opened 2018-04-18
Port the Fraction Bounce Sugar activity


llaske:
Port the [Fraction Bounce](http://activities.sugarlabs.org/en/sugar/addon/4488) Sugar activity (Python) to Sugarizer (JavaScript).
Follow the Sugarizer tutorial "[Create your own activity](https://github.com/llaske/sugarizer/blob/dev/docs/tutorial.md)"  to create a new Fraction Bounce activity.

Guidelines:

- The activity should be correctly integrated into Sugarizer (the activity icon should adapt its color when launched from the Sugarizer home view) 
- The activity should be responsive (work on any screen size)
- The activity should work with the keyboard and with the mouse (to support touch screen)
- The activity should use localization library lib/webL10n.js 
- The activity should work on any browser (Chrome, Firefox, Safari and EDGE)
- The activity should not use Internet connection (link to local content only)
- The activity should implement all existing features of the original activity
- The activity should retrieve graphics elements (icons, images, ...) of the original activity

ashish0910:
@llaske can I work on this ?

llaske:
@ashish0910 sure.

b18050:
@llaske Can I also work on this?

llaske:
No one has to ask to work on an issue. See https://github.com/llaske/sugarizer/blob/dev/CONTRIBUTING.md

Swayamsvk:
@llaske actually I am comfortable with the making of activities in sugarizer but I know a little bit of python. Should I try it?

llaske:
@Swayamsvk you're welcome. Porting don't mean to translate Python/Gtk code to JavaScript/HTML5 code. You can just use the current activity as inspiration but write a totally different source code.

llaske:
Fixed in https://github.com/llaske/sugarizer/pull/683

PR #683, opened 2020-03-09
[Activity Port] Fraction Bounce

dhruvmisra:
This is an initial build of Fraction Bounce game for issue #205 implementing the single player mode. Adding this to get an initial review on the progress made so far.

fractionBounce1

Single player mode is completely playable through keyboard with working options such as changing ball & background images, percentage mode, etc. The smileys are logged similar to the python activity. Custom fractions can also be added and will be saved in the journal.

**Yet to implement:**
- [x] ~~Sectors mode for the ball~~
- [x] ~~Toolbar help text~~
- [x] ~~Adding custom ball/background from journal~~
- [x] ~~Add Touch Controls~~
- [x] ~~Tutorial for the activity~~

This is an initial build hence it might contain some bugs. Please review the progress @llaske and @ashish0910 and let me know the changes if any.


llaske:
Very good start. Congrats!

dhruvmisra:
I have implemented most of the functionalities of the activity. I have also made the activity touch-friendly, i.e., if the user touches to the left/right of the ball, the ball will move in the respective direction. 
The workflow of multiplayer isn't completely clear to me, @llaske can you please clarify exactly how the multiplayer is going to work? 
Please review the current build @llaske and @ashish0910.

llaske:
Very cool.

I don't think a multiplayer version make sense for this activity.

Few remarks:

- [ ] The game should be playable with the mouse too
- [ ] The current mode (ball, fraction, percent) should stay selected in the toolbar (use `active` property in toolbar item)
- [ ] There is no visual indication that the answer is correct: I suggest to replace red line by a green line when the answer is correct
- [ ] The user should have a way to pause the game (may be by clicking again on the ball)
- [ ] The context of the game is not saved in the Journal
- [ ] The localization file refer to chess string :-)
- [ ] In percent mode the 1 to the right of the screen should be replaced by 100%
- [ ] The Add fraction don't seem to work
- [ ] When the game is played for a long time it hang the browser (look like there is an infinite loop somewhere)
- [ ] I suggest to stop the game after 10 fractions
- [ ] There is a scrollbar in the Journal popup

image

- [ ] When a custom background is set, I suggest to fit the image to the screen
- [ ] Remove debug messages in the code
- [ ] Miss a tutorial

dhruvmisra:
In my latest set of commits:
- [x] The game should be playable with the mouse too
- [x] The current mode (ball, fraction, percent) should stay selected in the toolbar (use active property in toolbar item)
- [x] There is no visual indication that the answer is correct: I suggest to replace red line by a green line when the answer is correct
- [x] The user should have a way to pause the game *[Can be paused by clicking the ball again.]*
- [x] The context of the game is not saved in the Journal *[Custom fractions will be saved in the journal]*
- [x] The localization file refer to chess string :-) *[An honest mistake. xD Fixed it.]*
- [x] In percent mode the 1 to the right of the screen should be replaced by 100%
- [x] The Add fraction don't seem to work *[Fixed a bug using parseInt]*
> - [ ] When the game is played for a long time it hang the browser (look like there is an infinite loop somewhere)
> - [ ] I suggest to stop the game after 10 fractions

I haven't been able to replicate this issue. How long did you play the game for?
Maybe we can make other improvements rather than stopping the game like reducing delay between bounces the longer you play as mentioned in the wiki notes. (https://wiki.sugarlabs.org/go/Activities/Fraction_Bounce)

- [x] There is a scrollbar in the Journal popup
> - [ ] When a custom background is set, I suggest to fit the image to the screen

The current behavior is to cover the screen with the image. If I fit the entire image, there will be blank spaces on the sides if the image's aspect ratio doesn't match the screen.

- [x] Remove debug messages in the code
- [x] Miss a tutorial

llaske:
Good job. It's cool to be able to play to this nice Sugar activity on Sugarizer.

My remarks:
* Pausing the game by clicking on the ball again is not a good idea. The user could accidentally click on the ball instead of to the left and to the right and it's frustrating to see the game paused. I suggest to include a play/pause button in the toolbar (like in the Physics activity)
* The 1 at the bottom right of screen should be replaced by 100% as soon as the Percent mode is selected
* Do not display decimal on percent value: 33,33% = 33%
* Mark unselectable text/content for score else it tend to be selected when clicking on the board

image

* Good idea to increase complexity the longer you play, let's go
* I'm not comfortable with an infinite game. I suggest the game stop after 10 plays. Any attempt to start the game will reinit the score board
* Changing ball by an image from the Journal has a strange behavior before start

image

* Regarding background image, may be you could repeat the image if it don't fit the screen?
* I think the speed of the ball should be adapted to the height of the screen else the game is very complex on small screens
* I suggest to rename the activity Fraction instead of Fraction Bounce (just the title, you could keep the same directory and id)

dhruvmisra:
In the last series of commits:
- [x] Added play/pause button in toolbar similar to Physics activity.
- [x] The 1 at the bottom right of screen should be replaced by 100% as soon as the Percent mode is selected
- [x] Do not display decimal on percent value: 33,33% = 33%
- [x] Mark unselectable text/content for score else it tend to be selected when clicking on the board
- [x] Reducing delay of launch by 100ms after each attempt to increase complexity
- [x] The game now stops after 10 attempts. Added a help message in the toolbar

image

- [x] Fixed strange behavior after image upload before starting the game
- [x] Repeating the custom image if it doesn't fit.
- [x] Added a formula to adapt speed as per height of the screen. Might not be perfect but should work with most screen sizes.
- [x] Changed the name to Fraction

llaske:
Very good job. Will try how it works on mobile device.

llaske:
I've done some testing on tablets, it's pretty usable but, to move the ball the user should let its finger on the screen (instead of trying to touch several times).

BTW I've two more remarks.

* There is a bug when the game is restarted the green/red indication no longer appears (see below):

Peek 04-04-2020 11-07

* In the initial Sugar game, a sound is played when the ball touch the ground. See [here](https://github.com/sugarlabs/fractionbounce/blob/master/bounce.py#L632) and [here](https://github.com/sugarlabs/fractionbounce/blob/master/bounce.py#L635). You could copy this sound from the TamTam micro activity, it's `crash.mp3` (if fail) and `bottle.mp3` (if success).

dhruvmisra:
Made the required changes @llaske.
- [x] Fixed ball not moving on multiple taps
- [x] Green/red bars will now appear as soon as the ball touches the slope
- [x] Added success/failure sounds

llaske:
Nice.

The original version also include support for accelerometer to move the ball. See [here](https://github.com/sugarlabs/fractionbounce/blob/master/bounce.py#L446).
Do you think you could add it too? You could take inspiration on Physics activity that rely on Cordova accelerometer support. It should not be so complex. See [here](https://github.com/llaske/sugarizer/blob/dev/activities/PhysicsJS.activity/js/activity.js#L75). 

One more thing: the tutorial should be updated to mention the new Play/Pause button.

dhruvmisra:
I have added the accelerometer support to the game, taking inspiration from Physics activity. Also updated the tutorial for play/pause buttons and accelerometer. Looking forward to your review @llaske.

llaske:
It's perfect.
Thanks a lot to port this mythic Sugar activity in Sugarizer.

PR #685, opened 2020-03-10
Update tutorial_step0.md

DevNM21:
The package is now around 560MB.


llaske:
Hmmm. I've just try to download it and it's a 316Mb package unzipped.
It's this size that should be considered because the warning is about the download time.
So you could round it to 320Mb.

DevNM21:
Oh okay. That is understood

Issue #686, opened 2020-03-11
Add a Simon mode game in TamTam activity


llaske:
The idea is to add a new game mode into TamTam micro activity to play to the famous [Simon game](https://en.wikipedia.org/wiki/Simon_(game)).

<img width="491" alt="image" src="https://user-images.githubusercontent.com/981841/76427499-aad79000-63ac-11ea-885a-1020c33e018c.png">

Like for the Piano mode, this new Simon mode will use the current selected instrument to play classic Simon notes.
Both notes and color should be inspired by the 4 items in the original game : E-note (blue, lower right), C#-note (yellow, lower left), A-note (red, upper right), E-note (green, upper left, an octave lower than blue).
The UI should be simple and attractive and as for other activities:

- [ ] Should be correctly integrated into Sugarizer and use buddy colors when possible.
- [ ] Should be integrated with Journal (save its context in the Journal).
- [ ] Should use localization.
- [ ] Should include a tutorial to explain the UI.
- [ ] Should be responsive (work on any screen size) and in fullscreen mode.
- [ ] Should work on any browser (Chrome, Firefox and Safari) and in Electron.
- [ ] Should not depend of an Internet connection (use local content only).
- [ ] Should work with mouse, touch and possibly keyboard.
- [ ] Every content/library used should rely on a Free Libre Open Source license.

sarthak-g:
working on this

ksraj123:
working on this. There is another issue #679 related to this. I think the same problem will appear in the new game mode as well unless the underlying problem is solved first.

llaske:
@ksraj123 I think the issue #679 is related to a Tone.js issue. So yes, the same problem should be there too. BTW if it bother you during development, a work around is to develop with a running Sugarizer Server or with Electron.

ksraj123:
@llaske Please review PR #687 - basic functional simon mode implemented.

llaske:
Fixed in https://github.com/llaske/sugarizer/pull/687

Issue #679, opened 2020-03-08
TamTam Micro Activity Piano Mode Not working in Chrome in file:///


Hemansh31:
![TTM](https://user-images.githubusercontent.com/48997514/76158466-0ea75200-613c-11ea-976c-9032b086dcee.gif)

llaske:
Can reproduce it only when launched locally from the browser using file:///

llaske:
Raised an error in Tone.js library [here](https://github.com/Tonejs/Tone.js/issues/628)

llaske:
Fixed in https://github.com/llaske/sugarizer/commit/28e634cce8d0f8e7a63c667c27e3690c27ca54ee

PR #687, opened 2020-03-11
[Issue #686] Simon Mode added to TamTamMicro activity

ksraj123:
A basic functional Simon mode added to TamTam Micro activity for issue #686
This is an initial version so some features are yet to be implemented and there may be some bugs or possible improvements.
@llaske Opening this to get a review on if I am heading in the right direction.

Tam Tam

Tasks - 
- [x]  Should be correctly integrated into Sugarizer and use buddy colors when possible.
- [x]  Should be integrated with Journal (save its context in the Journal).
- [x]  Should use localization.
- [x]  Should include a tutorial to explain the UI.
- [x]  Should be responsive (work on any screen size) and in fullscreen mode.
- [x]  Should work on any browser (Chrome, Firefox and Safari) and in Electron.
- [x]  Should not depend of an Internet connection (use local content only).
- [x]  Should work with mouse, touch and possibly keyboard.
- [x]  Every content/library used should rely on a Free Libre Open Source license.

Known Issues - 
TamTam Activity does not function as intended when launched locally in file:///  - issue #679 - Issue apparent in Simon mode too.


ksraj123:
This activity did not support the journal. So support for the journal has been added to both piano mode and the new simon mode.

Before - 
Tam Tam did not support the journal

Now - 
Tam Tam now supporting journal

ksraj123:
Styling improvements to simon mode.

Screenshot from 2020-03-12 06-39-18

ksraj123:
@llaske Would love to have your feedback and suggestions on this initial build, specially on the following points -

- Is any functionality expected out of the keyboard like in the piano mode?
- Is the icon in the toolbar alight?
- The green button should have E-note but an Octave lower than blue. This would correspond to which value in the `pitchMap` in `piano.js` file? Right now both the buttons are mapped to same E-note.
- Styling of piano mode is kept consistent with piano mode. Any suggested changes to styling?

llaske:
Nice start but it's so sad that it's not possible to play the game today...

Few remarks:
- Yes, the game should be playable using touch, click and keyboard
- Your first the colored icon was very cool but not compatible with Sugar UI rules (use only black&white icons in the toolbar). But the new icon is ugly. May be you could arrange another icon with gray colors
- Because of this new mode, I think we should add a new icon in the toolbar for the "instrument" mode. It's not intuitive to have to click on the current mode to unselect it and go back to instrument mode
- Regarding the green note, yes it should play the note one octave lower. Not sure it's related to pitchmap, have a look on the Tone.js documentation
- Your current style is good but the Simon game could take more space on the screen. Plus the color of pressed button should be different than the color for unpressed button (more dark color)
- You should prepare a place to display the score
- You should display a button to start/stop/restart the game, may be in the middle of the board

ksraj123:
@llaske Basic gameplay added to simon mode. Yet to implement many features and there may be some bugs right now, just wanted to know if the current gameplay is on the lines of what is expected. Requesting feedback and suggestions.

Sorry for the debug messages will remove them in next commit.

Peek 2020-03-13 07-41

ksraj123:
@llaske Requesting feedback on gameplay and changes made in the current build since last review. Some features are yet to be improved and implemented.

- [ ] Functional Gameplay added.

- [ ] Added an instruments mode.

- [ ] Changed icon for simon mode in toolbar.

- [ ] Score and level added.

- [ ] Added Start and reset button.

llaske:
Good job. I like the general UI of the game.

My remarks:

- [ ] The Simon icon is nice but the instrument icon could be improve. It should more look like to an instrument.
- [ ] I don't like the idea to force the user to restart the game each time. I think the next level should start automatically. Plus the color sequence for the next level should start with the previous one, just adding a new color. It was the way of work in the original game :-)
- [ ] I'm not a big fan of the Wrong/Correct animation. The Wrong/Correct message is fine but do not change the board color in red/green. It's disturbing.
- [ ] The first level should start with one color
- [ ] When there is wrong answer, the level should go back to 1
- [ ] All messages (Score, Level, Start, Wrong, Correct, ...) should be localized
- [ ] The restart button should appear only after a wrong sequence
- [ ] When stopping the activity on the Simon mode, then restart the activity there is a mix between Instrument and Simon mode in the toolbar
- [ ] Regarding keypad, I think you can't consider that there is a no numpad so arrow keys is better

ksraj123:
Suggested changes made to the gameplay,

- [ ] Next level starts with the same sequence adding 1 more colour to it
- [ ] on wrong answer level goes back to 1
- [ ] next level starts automatically no need to click start button every time now
- [ ] removed the wrong and right board animation
- [ ] changed first level to start with 1 colour

ksraj123:
@llaske ??

llaske:
Nice.

My remarks:
- Still an issue on selected item in toolbar: when closing the activity in Piano or Simon mode, both  Instrument and Piano/Simon icons are selected at restart.
- By default, if no instrument has been selected, the same default instrument than Piano mode should be use (Piano)
- When the user click directly on a color when starting Simon mode, there is a "Wrong" message. Nothing should happen until the start button is clicked
- The game should not restart automatically at end
- The restart button seems to do nothing. It's probably usefulness and could be removed
- The score is hidden in fullscreen mode
- When changing mode during Play an error is displayed in the console: 

```
simon.js:185 Uncaught TypeError: Cannot read property 'addRemoveClass' of undefined
    at e.<anonymous> (simon.js:185)
    at Array.forEach (<anonymous>)
    at e.TamTam.Simon.addRemoveAll() [as addRemoveAll] (simon.js:184)
    at e.<anonymous> (simon.js:159)
```

llaske:
@ksraj123 do you still work on it?

ksraj123:
@llaske Sorry for the delay, Most of the issues mentioned in the previous review are fixed, However I am not sure what you meant by Game should not restart automatically at end, Can you please explain that point?

ksraj123:
@llaske Changes Suggested in the last review are made - 

- Selection of both instrument mode and simon/piano mode icon on restart fixed
- Piano made default instrument of Simon mode i.e. same as in Piano mode
- User click on colour before game start issue fixed, and many other similar bugs fixed
- Added a play again button for the game to restart on a wrong answer
- Restart button removed
- Score no longer hidden in fullscreen mode
- console error while changing mode fixed

Here is what the current build looks like 
now

llaske:
Cools. Thanks for update.

My remarks:
* At each level the game should become more complex, the idea is to reduce slightly the delay between each note on each new level. So the melody is played quicker. It's how the original game worked
* There is still an issue when the activity is reopened. The toolbar is updated but no the screen.

image


* When a wrong keyboard is used, there is an error in the console:

```
simon.js:94 Uncaught TypeError: Cannot read property 'addRemoveClass' of undefined
    at e.TamTam.Simon.clickColor() [as clickColor] (simon.js:94)
    at e.TamTam.Simon.handlePlayNote() [as handlePlayNote] (simon.js:43)
    at e.TamTam.Simon.handlePlayNoteListener() [as handlePlayNoteListener] (simon.js:110)
```

* Could you reduce the font size for the START / PLAY AGAIN message, else it will be too short when localized
* Could you avoid mixing TAB and SPACE indentation style in you source code

llaske:
@ksraj123  Could you fix the leaving issues so we could merge the PR.

ksraj123:
@llaske Most of the issues raised in the last review are fixed.
- Game gets faster by 50ms on every level, the game may crash or give an error after level 30 but practically it should not be a problem
- Console error when wrong keyboard button clicked is fixed
- Font size for START / PLAY AGAIN reduced.

I was not able to reproduce the screen not updated but toolbar does issue, can you please confirm is the issue is still and what steps you took?

issue not reproducible

llaske:
Very nice.

Unfortunately the toolbar issue is still there. See below:

Peek 03-05-2020 10-46

I've tested both in Chrome and in Firefox.
I've put some trace to understand: the issue is related to the fact that the `app.draw` method could be called before the end of `loadAsText` call in `activity.js`.
It's what always happens on my side. Sounds like you've got more luck on your side :-)
To fix that, may be you could force redraw at end of `loadAsText` call if the current mode is different from the current one?

ksraj123:
@llaske Made suggested change. Please review.

llaske:
Good job. It works now.
I'm going to test on mobile devices.

llaske:
I've merged it.
Thanks a lot for this great work.

PR #696, opened 2020-03-12
Add a feature to change difficulty of falabracman activity

### LINKED ISSUE #695: Add a feature to change difficulty of falabracman activity
- add a feature to change difficulty of falabracman activity like easy, medium, hard.
- changing difficulty may includes changing no of obstacles and changing player's speed
- setting screen may get updated for this.

prakashujjwal1010:
- fixed #695 
- added three buttons in setting screen for slecting easy, medium and hard difficulty
- updated locale.ini
- updated tutorial.js
- difficulty is also stored in journal
- improved responsiveness of setting screen

Screenshot from 2020-03-13 03-03-30

alternative: instead of the buttons in setting screen a toolbar can have a palette for selecting difficulty


ashish0910:
Looks good . Just one thing , the current difficult can be highlighted. For example after setting difficulty as medium and play game and then again come back to set difficulty then one should see medium as highlighted . This will help user to compare difficulty levels .

prakashujjwal1010:
Yep, you are right. I'll add that

llaske:
Not an issue.

PR #702, opened 2020-03-13
[GSOC 2020-TASK] chess.activity

prakashujjwal1010:
- New Chess.activity
- implemented using 
   - Vue.js, 
   - Chessboard.js, 
   - p4wn.js
- The activity should be playable against computer or against another user on the network (using presence)
- The activity should display the list of move during play.
- The activity should be integrated with Journal (save its context in the Journal).
- The UI of the activity should be simple, self-explanatory and respect Sugar standards.
- The activity should be correctly integrated into Sugarizer and use buddy colors for background etc
- The activity should use localization.
- tutorial added
- The activity should be responsive (work on any screen size) and should include a fullscreen button.
- The activity should work on any browser (Chrome, Firefox and Safari) and in Electron.
- The activity should not depend of an Internet connection (use local content only).
- The activity work with mouse, touch.
- Every content/library used in the activity should rely on a Free Libre Open Source license.
- timer or chess clock added


prakashujjwal1010:
Screenshot from 2020-03-14 00-41-56


Screenshot from 2020-03-14 02-24-21

ashish0910:
Just one thing -

In multiplayer mode when the number of users reaches one you should start playing against computer ( In current implementation when there are 2 users and one leaves then the game gets stuck )

Screenshot 2020-03-14 at 11 10 36 AM

prakashujjwal1010:
thanks @ashish0910 for reviewing. 
will add the requested functionality :)

llaske:
I think your Chess implementation is the best one that we've received.
Congrats!

I've got few remarks to improve it:
- Change the icon of the activity, for example [this one](https://github.com/llaske/sugarizer/blob/902ed1b7689257deb6314234366e945bf2b94eac/activities/Chess.activity/activity/activity-icon.svg)
- Change the green box color, may be just a grey one will be good
- Improve the tutorial to explain how to play, I do like the one in https://github.com/llaske/sugarizer/pull/659
- Rename the title of the HTML page (currently "My Activity")
- Reduce size of icons in the list of move
- Move the new game icon to the right of the neighborhood icon
- Move the undo icon to the right of the new game icon
- Change icon to switch change color, for example [this one](https://github.com/llaske/sugarizer/blob/58cd95926743d479ad92513f69c7c9bc67d46b97/activities/Chess.activity/icons/switch-player.svg) (need to be adapted however)
- It's not clear which one is the current player, may be you could just add a symbol before the buddy icon to show it
- Could you scroll the list of move at each play
- Not a big fan of your popup message at end, I think you don't need it
- Remove debug messages
- Solve merge conflict

prakashujjwal1010:
thanks :).
okay, will start the work and add the improvements.

prakashujjwal1010:
- [x] icons changed
- [x] box color changed
- [x] tutorial changed, added images for pieces,
- [x] title changed
- [x] icons size reduced in moves list
- [x] toolbar buttons positions changed
- [x] indication added for current player 
- [x] scroll list is movable
- [x] popup removed
- [x] debug messages removed
- [x] solved merge conflict
- [x] now host can toggle color in shared mode also
- [x] now In multiplayer mode when the number of users reaches one user start playing against computer 
- [x] bugs fixed

llaske:
Very good job. Thanks.

Few things left:
- The list of move is now scrollable but the idea is to automatically scroll it at each play so the last plays are always visible
- When the button fullscreen is clicked the toolbar disappear but the canvas size is not resized so the new space is not used
- In touch, at each play there is a set of error in the console:

         jquery.min.js:3 [Intervention] Unable to preventDefault inside passive event listener due to target being treated as passive. See https://www.chromestatus.com/features/5093566007214080

llaske:
Hi @prakashujjwal1010 , could you solve the remaining issues?

prakashujjwal1010:
issues solved
- [x] The list of move is now scrollable but the idea is to automatically scroll it at each play so the last plays are always visible
- [x] When the button fullscreen is clicked the toolbar disappear but the canvas size is not resized so the new space is not used

regarding the error,
I think the error is in chessboard.js as the same error is also available on their website ,[here](https://chessboardjs.com/examples#5001)
@llaske any advice on this error.

quozl:
@prakashujjwal1010, yes, you can be asked to solve issues in upstream projects and dependencies.  It really helps to engage with the upstream project about it because they have the expertise in their code, and only if they are unresponsive then you fork.

llaske:
Ok I've seen you've declared an issue on chessboard.js [here](https://github.com/oakmac/chessboardjs/issues/187). Nice. 

Other issues are fixed. Thanks.

llaske:
Thanks a lot for your work on this activity. It's a valuable addition to Sugarizer.

PR #708, opened 2020-03-17
Measure Activity for Sugarizer

abhishektanwar:
Developed the Measure Activity in sugar for Sugarizer.
measure_1

measure_2


abhishektanwar:
#709

quozl:
57b113c commit message is a bit brief.  You could say what you used, like p5 sound.

I see you added `exec = sugar-activity-web`.  Will you be porting this to Sugar or does it work already?

abhishektanwar:
@quozl @llaske  I am developing this activity for Sugarizer. Sugar has the measure activity and I took the inspiration from https://help.sugarlabs.org/measure.html .
I have used the p5.js ,p5.sound and p5.dom libraries in this activity.
There is more work that needs to be done like plotting waveform for different octaves,play a sound and show how waveform is varying at different amplitudes.

llaske:
Good start. Nice.

You should also adapt the UI to Sugarizer. Specifically, settings should be in a palette, not directly on the string. Plus the buttons in the toolbar should be selected/unselected instead of hidden.

abhishektanwar:
@llaske Thanks for the feedback.
By settings do you mean the sliders at the bottom of the page ?
Also can the x-axis scale be shown on the graph or do I need to place it somewhere else?If I have to place it somewhere else,where should it be placed?
Working on suggestions.

llaske:
Yes I mean the sliders.
Regarding the x-axis scale, may be you could place it at the bottom of the graph like in Measure activity.

abhishektanwar:
Working on the suggestions.

abhishektanwar:
@llaske I am finding difficulties in moving the sliders to the toolbar palette. As the current sliders have been made using p5.js and these sliders are directly linked to the waveforms if I am making any changes to these, everything breaks. Could you provide me some documentation for sliders in sugarizer . I have tried to implement the working of sliders from different activities but it didn't work.

llaske:
@abhishektanwar sliders in Sugarizer is just HTML input with the type `range`. You could find a very basic sample [here](http://sugarlabs.github.io/sugar-web-samples/inputrange.html). It's used for example in Speak or Game of Life activities.
Regarding the current sliders related to p5.js code, may be you could just hide them.

abhishektanwar:
@llaske thanks for the help it worked.
I have fixed all the requested changes. Please review.
Activity is complete now.

activities/Measure.activity/lib/speechpalette.js
```diff
@@ -0,0 +1,48 @@
...
+
+        var containerElem = document.createElement('div');
+        containerElem.innerHTML = template;
+        this.setContent([containerElem]);
+
+        this.pitchScale = containerElem.querySelector('#pitchvalue');
+        this.rateScale = containerElem.querySelector('#ratevalue');
+
+        this.pitchScale.onclick = function() {
+            console.log("he;;p");
```
ashish0910: Can you please remove these console messages .

activities/Measure.activity/lib/speechpalette.js
```diff
@@ -0,0 +1,48 @@
...
+            console.log(this.value);
+            // document.getElementById('pitch').innerHTML = this.value;
+        }
+        this.rateScale.onclick = function() {
+            // document.getElementById('rate').innerHTML = this.value*(300/100) + 10;
+            console.log("hell");
```
ashish0910: Please remove this console message .

abhishektanwar: fixed

activities/Measure.activity/lib/speechpalette.js
abhishektanwar: fixed

abhishektanwar:
@ashish0910 @llaske removed console logs. Please review.

llaske:
Mmmm. The activity does nothing on my side.
At start the screen is totally black:

image

When the graph button is clicked the activity seems to listen but nothing happen:

image

Plus will be nice to improve icons/slide bar proportions/alignement in palette:

image

abhishektanwar:
@llaske @ashish0910 Waveform is plotted from the values of sliders (amplitude and frequency). Initially, the values of amplitude and frequency slider is 0(zero). As the sliders are moved, the corresponding waveform is plotted.
- Fixed alignment of icons and sliders in palette.
Please review


measure_3a

llaske:
So do not start with a value of zero for amplitude and frequency and display the waveform at launch.
Note sure to understand why sliders values change when switching graph?
Plus, will be nice to save values (and current graph selection) in the datastore so the user could reopen the activity to recover its context.

abhishektanwar:
thanks for the feedback. working on it.

abhishektanwar:
@llaske @ashish0910 
- added datastore functionality
- fixed initial zero value of sliders
- on initializing an old instance of the activity , activity starts from previous slider settings and last waveform(time or frequency)
- saves current graph when activity is stopped

please review

llaske:
That's nice.

Few remarks:
- Somethings goes wrong with the palette, the icon become big when the palette is open
- Sound like Amplitude and Frequency settings are related to the graph. That's fine but all settings are not saved in datastore, only the one of the active graph
- When clicking on fullscreen/unfullscreen you should not only hide the toolbar, you should also resize the canvas
- Look like the activity description in the tutorial repeat two times the same thing "This activity draw a picture..."
- Try to force the opening of palette before displaying the slider helper in tutorial (could be done with event on steps)
- Title of graphs should be localized too
- Try to better indent your code, in `activity.js` some blocks have a very weird indentation
- Change the title of the HTML page - currently "My activity"

Also, is it possible to add two features from the Sugar version of the activity ?
- Freeze the graph (need to add a Pause/Play button in the toolbar)
- Save the current graph as PNG in the journal

abhishektanwar:
> * Sound like Amplitude and Frequency settings are related to the graph. That's fine but all settings are not saved in datastore, only the one of the active graph

I did not understand what you are trying to say. Currently, I am saving the active graph(to load the activity from the save graph when an old instance is initiated), the values of all the slider(on both time and frequency graph) are also saved to start the graph from previous slider settings. Could you please explain a bit? At all time when a slider value is changed in any of the graph time or frequency based , that value is stored locally in **data** object and when activity is stopped all the updated settings of last active graph and sliders is stored in datastore obejct.

> * Look like the activity description in the tutorial repeat two times the same thing "This activity draw a picture..."

I don't see this problem on my side. Maybe there is a confusion between the first and the eleventh step of tutorial . First dialog box of tutorial gives the description of what the activity is all about(This activity draw a picture) and eleventh dialog box gives a brief information about the graph ("This graph shows the waveform generated").
measure_4


 > * Try to force the opening of palette before displaying the slider helper in tutorial (could be done with event on steps)

This is a great suggestion. I tried to work it out, could you please help me with some more information or steps to be followed. Any activity that might have a similar kind of working would help.

> * Freeze the graph (need to add a Pause/Play button in the toolbar)

What I understand from your point of view is to freeze the graph and take the screenshot.
I am not sure about freezing this because the graph is the result of live input through a microphone, I believe we can add the functionality of taking the screenshot of the graph at any time and save it in the journal.

llaske:
> I did not understand what you are trying to say.

See in the video capture below. Amplitude and Frequence are different for each graph. Values are not retrieved for Sound time frequence graph when the activity is reopen.

Peek 12-04-2020 22-42


> I don't see this problem on my side

See screen capture below. I've underlined in red the repeated text.

Capture du 2020-04-12 22-44-16


> Any activity that might have a similar kind of working would help.

See the popup for the activity palette in SharedNotes activity [here](https://github.com/llaske/sugarizer/blob/dev/activities/SharedNotes.activity/lib/tutorial.js#L29).


> What I understand from your point of view is to freeze the graph and take the screenshot.

Yes, see buttons 5 and 6 [here](https://github.com/godiard/help-activity/blob/master/source/measure.rst#using).
Regarding 5, may be you could set the input to null during the freeze.
Regarding 6, it's possible to convert a Canvas into an image then store it in the datastore. See for example [here](https://github.com/llaske/sugarizer/blob/dev/activities/Moon.activity/js/moon-activity.js#L305) how it works in Moon activity.

abhishektanwar:
> Yes, see buttons 5 and 6 here.
Regarding 5, may be you could set the input to null during the freeze.
Regarding 6, it's possible to convert a Canvas into an image then store it in the datastore. See for example here how it works in Moon activity.

Regarding button 5, mic captures input live and plots a graph for it, I am afraid that a null input will plot a straight horizontal line on the graph(it will not be plotting sinusoidal wave for null input).
I think what you are expecting from freeze is that whatever is the shape of the wave currently only that should be shown on the canvas with no effect of input through the mic. Like the graph [here](https://github.com/godiard/help-activity/blob/master/source/measure.rst#using)

Rest of the issues I understand and working on them

llaske:
Hmmm. May be you could just stop drawing when the Pause button is clicked. If the canvas is not refreshed, it should work.

Another thing: could you remove ES6 syntax in your file? I've seen for example lot of use of `let` keyword. Unfortunately, ES6 syntax is not supported by all browsers.

abhishektanwar:
> * Somethings goes wrong with the palette, the icon become big when the palette is open

I tried to fix this however the issue is with this particular icon. So I have changed the icon, now 


> * Sound like Amplitude and Frequency settings are related to the graph. That's fine but all settings are not saved in datastore, only the one of the active graph

FIxed
> * When clicking on fullscreen/unfullscreen you should not only hide the toolbar, you should also resize the canvas

FIxed
> * Look like the activity description in the tutorial repeat two times the same thing "This activity draw a picture..."

FIxed
> * Try to force the opening of palette before displaying the slider helper in tutorial (could be done with event on steps)

FIxed
> * Title of graphs should be localized too

FIxed
> * Try to better indent your code, in `activity.js` some blocks have a very weird indentation

FIxed
> * Change the title of the HTML page - currently "My activity"

FIxed
> 
> Also, is it possible to add two features from the Sugar version of the activity ?
> 
> * Freeze the graph (need to add a Pause/Play button in the toolbar)
> * Save the current graph as PNG in the journal

I have incorporated the freeze functionality within the save functionality. The user does not have to freeze the graph/waveform and then take a screenshot/convert graph in an image. Users can simply click the save button and the image of the graph will be saved in the journal without the hassle of freeze/play , this makes it easier to operate the activity.

activities/Measure.activity/js/activity.js
```diff
@@ -0,0 +1,438 @@
...
+			if(!environment.objectId){
+				console.log("New INstance");
+				setRetreivedDatastore(data);
+				initiator();
+			}
+			else{
+				activity.getDatastoreObject().loadAsText(function(error, metadata, data) {
+					if (error==null && data!=null) {
+						pawns = JSON.parse(data);
+						console.log(pawns);
```
llaske, line 30: Remove debug message

activities/Measure.activity/js/activity.js
```diff
@@ -0,0 +1,438 @@
...
+						check_data=pawns;
+						trans();
+						data=check_data;
+						console.log(data);
```
llaske, line 34: Remove debug message

activities/Measure.activity/js/activity.js
```diff
@@ -0,0 +1,438 @@
...
+		var palettechecker=0;
+		env.getEnvironment(function(err, environment) {
+			currentenv = environment;
+			// Set current language to Sugarizer
+			var defaultLanguage = (typeof chrome != 'undefined' && chrome.app && chrome.app.runtime) ? chrome.i18n.getUILanguage() : navigator.language;
+			var language = environment.user ? environment.user.language : defaultLanguage;
+			webL10n.language.code = language;
+			
+			if(!environment.objectId){
+				console.log("New INstance");
```
llaske, line 22: Remove debug message

activities/Measure.activity/js/activity.js
```diff
@@ -0,0 +1,438 @@
...
+						data=check_data;
+						console.log(data);
+					}
+				});
+			}
+		});
+
+		function setRetreivedDatastore(data_val){
+			retreived_datastore=data_val;
+			console.log("retreived",retreived_datastore);
```
llaske, line 42: Remove debug message

activities/Measure.activity/js/activity.js
```diff
@@ -0,0 +1,438 @@
...
+		}
+
+		function trans(){
+			console.log(check_data);
```
llaske, line 46: Remove debug message

activities/Measure.activity/js/activity.js
```diff
@@ -0,0 +1,438 @@
...
+			setRetreivedDatastore(check_data);
+			initiator(retreived_datastore);
+		}
+		document.getElementById("save-image-button").addEventListener('click',screenshot);
+		function screenshot(){
+			var ss = document.getElementById("canvas");
+			html2canvas(ss).then(function(canvas){
+				var mimetype = 'image/jpeg';
+				var inputData=canvas.toDataURL("image/jpeg",1);
+				console.log(inputData);
```
llaske, line 56: Remove debug message

activities/Measure.activity/js/activity.js
```diff
@@ -0,0 +1,438 @@
...
+				var s = function(p){
+					p.mic1;
+					p.yslider;
+					p.xslider;
+					p.buttonx1;
+					p.buttony1;
+					p.buttonx2;
+					p.buttony2;
+					p.bg;
+					p.img;
```
llaske, line 202: Not sure to understand what is done here?

activities/Measure.activity/js/activity.js
```diff
@@ -0,0 +1,438 @@
...
+						});
+						p.timegraphtitle=webL10n.get("TimeGraph")
+						p.xaxis1=p.createP(p.timegraphtitle);
+						p.xaxis1.style('padding-left:40%');
+						p.xaxis1.style('color:white');
+					};
+
+					p.draw = function(){
+						p.background(p.bg);
+						let waveform = p.fft.waveform();
```
llaske, line 265: Avoid ES6 syntax

activities/Measure.activity/js/activity.js
```diff
@@ -0,0 +1,438 @@
...
+						p.noFill();
+						p.beginShape();
+						p.stroke(255);
+						for (let i = 0; i < waveform.length; i++){
+							let x = p.map(i*p.txp, 0, waveform.length, 0, p.width);
+							let y = p.map( waveform[i]*p.typ, -1, 1, 0, p.height);
```
llaske, line 271: Avoid ES6 syntax

activities/Measure.activity/js/activity.js
```diff
@@ -0,0 +1,438 @@
...
+				var t = function(p){
+					p.mic2;
+					p.cnv;
+					p.yslider;
+					p.xslider;
+					p.yaxis;
+					p.xaxis;
+					p.bg;
+					p.px;
+					p.py;
```
llaske, line 333: Not sure to understand what this code do?

activities/Measure.activity/js/activity.js
```diff
@@ -0,0 +1,438 @@
...
+							}
+						});
+						p.frequencygraphtitle=webL10n.get("FrequencyGraph")
+						p.xaxis=p.createP(p.frequencygraphtitle);
+						p.xaxis.style('padding-left:40%');
+						p.xaxis.style('color:white');
+					};
+					p.draw = function(){
+						p.background(p.bg);
+						let spectrum = p.fft.analyze();
```
llaske, line 394: Avoid ES6 syntax

activities/Measure.activity/js/activity.js
```diff
@@ -0,0 +1,438 @@
...
+						p.noStroke();
+						p.fill(255, 255, 255);
+						for (let i = 0; i< spectrum.length; i++){
+							let x = p.map(i*p.fxp, 0, spectrum.length, 0, p.width);
+							let h = -p.height + p.map(spectrum[i]*p.fyp, 0, 255, p.height, 0);
```
llaske, line 399: Avoid ES6 syntax

activities/Measure.activity/js/activity.js
```diff
@@ -0,0 +1,438 @@
...
+							p.rect(x, p.height, p.width / spectrum.length, h )
+						}
+						p.stroke(255);
+					};
+			
+					p.windowResized = function(){
+						p.resizeCanvas(window.innerWidth-100,window.innerHeight-200);
+						document.getElementById("fullscreen-button").addEventListener('click',function(){
+							console.log("po");
```
llaske, line 408: Remove debug message

activities/Measure.activity/js/activity.js
```diff
@@ -0,0 +1,438 @@
...
+							p.resizeCanvas(window.innerWidth-50,window.innerHeight-100);
+						});
+						document.getElementById("unfullscreen-button").addEventListener('click',function(){
+							console.log("pot");
```
llaske, line 412: Remove debug message

activities/Measure.activity/index.html
```diff
@@ -0,0 +1,58 @@
+<!DOCTYPE html>
+<html>
+
+<head>
+<meta charset="utf-8" />
+<title>Measure Activity</title>
+<script type="text/javascript" src="lib/p5.js" ></script>
```
llaske, line 7: Use p5.min.js instead and delete p5.js

activities/Measure.activity/index.html
```diff
@@ -0,0 +1,58 @@
...
+  <script type="text/javascript" src="lib/p5.sound.js"></script>
```
llaske, line 8: Use p5.sound.min.js and delete p5.sound.js

activities/Measure.activity/js/activity.js
abhishektanwar: Fixed

activities/Measure.activity/js/activity.js
abhishektanwar: Fixed

activities/Measure.activity/js/activity.js
abhishektanwar: Fixed

activities/Measure.activity/js/activity.js
abhishektanwar: Fixed

activities/Measure.activity/js/activity.js
abhishektanwar: Fixed

activities/Measure.activity/js/activity.js
abhishektanwar: Fixed

activities/Measure.activity/js/activity.js
abhishektanwar: It was for testing purposed . Fixed

activities/Measure.activity/js/activity.js
abhishektanwar: Fixed

activities/Measure.activity/js/activity.js
abhishektanwar: It was for testing purpose. Fixed

activities/Measure.activity/js/activity.js
abhishektanwar: Fixed.

activities/Measure.activity/js/activity.js
abhishektanwar: Fixed

activities/Measure.activity/index.html
abhishektanwar: Fixed

activities/Measure.activity/js/activity.js
abhishektanwar: fixed

activities/Measure.activity/js/activity.js
abhishektanwar: fixed

llaske:
Cool. Good job.

When the graph is paused it's not at the same place than the running graph.
Could you try to set it at the exact same place?

llaske:
Nice.

I've got an issue when I'm launching from file://
The graph is not displayed and there is an error in console:

image

I guess it could be the same issue than [here](https://github.com/llaske/sugarizer/issues/679) and could require to load the file as a blob object.

abhishektanwar:
It is working on my side @llaske 
image

Peek 2020-04-21 22-57
it worked with google-chrome-stable --allow-file-access-from-files also.

I am not sure why this is happening. Maybe if you could show me how you produced this error

llaske:
Of course you could open the file alone but there is an issue when it's open from Measure activity in file://

Peek 21-04-2020 21-20

abhishektanwar:
@llaske I have fixed the file:// error(this error was only in chrome ) but after fixing this error I encountered  a new error 
image
the link with the error in console : https://goo.gl/7K7WLu
This error occurs in chrome only and not firefox due to google's policy . I have tried to bypass so that as soon as the activity starts graph can be plotted by the input audio but chrome only take audio input after a user gesture . 
It works fine in firefox . what we can do is when whenever or wherever user clicks or touches ,graph will start

I have push fix for both the errors . please review

llaske:
Okay. I've already seen this error.
I suggest to start the activity paused, in this case only (if on Chome and if `file://`).

BTW your current fix - paste the entire image in base64 in the source file - is ugly: will be better to use and XmlHttpRequest and read the file into a blob like [here](https://github.com/Tonejs/Tone.js/issues/628).
Plus remove commented and unused lines.

abhishektanwar:
> I suggest to start the activity paused, in this case only (if on Chome and if `file://`).

I think to maintain the working of activity similar to for all the browsers , activity can be started with pause state (the button which is being used to pause the graph) and then the user can start the activity by clicking on play button or we can proceed as you suggested earlier by tweaking the pause button for chrome browser.

llaske:
I'm not agree. The activity should start automatically.
Paused at startup is just a work around.

abhishektanwar:
> I'm not agree. The activity should start automatically.
> Paused at startup is just a work around.

Okay then activity will start normally for all browsers . But for chrome ,activity will start with paused state . User will have to click play button. 
Please let me know if this is fine.

llaske:
> But for chrome ,activity will start with paused state 

For Chrome, only when run from `file://`

llaske:
The activity no longer work on `http://` :

Capture du 2020-04-24 21-31-07

abhishektanwar:
can you layout the steps you followed to run this activity from the beginning . It worked fine on my side @llaske

llaske:
Just cleaning the browser cache. Then launch a new instance of Measure.

abhishektanwar:
@llaske there was one error . Fixed it . Please review again.

llaske:
In `http` mode it works but the title "Sound time base" is displayed under graph:

Capture d’écran du 2020-04-27 18-06-20
Capture d’écran du 2020-04-27 18-05-56

In `file` mode the graph is not drawn correctly:

Capture d’écran du 2020-04-27 18-07-39

Try to do more testing before sending your PR.

abhishektanwar:
@llaske it is working fine on my side in file mode .
If you are using some other method to testing other than I am using please let me know . 
When in chrome and file mode , I am showing the empty graph ,user clicks on play button which also accounts for the policy of chrome that a user gesture is required for accessing microphone.

Peek 2020-04-28 00-40

llaske:
Do you clear your browser cache?

abhishektanwar:
@llaske Yes, I have cleared all the cached files in the chrome browser.
measure

llaske:
Sound like it happens randomly. When it happens, there is a message `file protocol error` in the console:

Capture d’écran 2020-04-28 à 21 15 58

Another issue, when switching graph in pause mode, a double graph appears:

erreree

abhishektanwar:
> Sound like it happens randomly. When it happens, there is a message `file protocol error` in the console:

What can we do with this?What do you recommend?
> Another issue, when switching graph in pause mode, a double graph appears:

Yes it can happen because according to previous discussions activity was designed to switch between time and frequency graphs only when the graph is in play mode (not in pause state).

llaske:
> What can we do with this?What do you recommend?
Try to explore in which condition this can happen and fix it.

> Yes it can happen 
Okay but it should be solve. We can't let it like that.

abhishektanwar:
> Sound like it happens randomly. When it happens, there is a message `file protocol error` in the console:

I have tweaked the activity for chrome a bit . Please review it again . Works fine on mine side(tested after clearing cache)
> Another issue, when switching graph in pause mode, a double graph appears:

Fixed

llaske:
Still few issues:

- The graph should start automatically in Chrome when not in `file` mode. My request was to pause the activity at start in Chrome AND in `file` mode
- When the activity is reopened, the double background issue still appears:

Peek 01-05-2020 22-08

- I've got several times the random error described before. Following is some screen capture when it happens. Did you have the issue sometimes on your side?

Capture d’écran du 2020-05-01 21-56-48
Capture d’écran du 2020-05-01 21-48-07

- Remove commented lines and `//change` comment

llaske:
Replaced by https://github.com/llaske/sugarizer/issues/709

PR #713, opened 2020-03-18
Implode Activity

abhishektanwar:
This is the basic implementation in which random blocks are created. Black blocks represent space.


llaske:
Fixed in https://github.com/llaske/sugarizer/pull/784

PR #721, opened 2020-03-21
Geolocation

### LINKED ISSUE #718: Geolocation functionality for clients
The addition of geolocation functionality to track the location of clients could be a good addition to sugarizer.

abhishektanwar:
fixes #718 
I am extracting the location of the client and consoling it out as of now. A column could be added to the sugarizer-server dashboard in the user's section.I think it could be a great addition to sugarizer.

I need some help in understanding how this can be implemented in the server. I tried to understand the flow of control in the server but could not.
Any help would be appreciated.
image

image


llaske:
Not an issue.

PR #726, opened 2020-03-21
[Issue #725 fix] Fixed acceptance of repeated letters in Falabracman

### LINKED ISSUE #725: Falabracman only accepts a specific letter in case of repeated letters in a word
When a word contains repeated letters, only a specific letter can be accepted and all others are treated as different letters. See here:

FalabracmanMultipleLetters

FalabracmanMultipleLetters2

Shouldn't all repeated letters also be accepted? What do you say @llaske?
I'm working on a fix for this.

dhruvmisra:
Fixes #725.

FalabracmanMultipleLettersFix

The GIF might be considered luck. I recommend testing yourself. :)


llaske:
Nice. More fun like that.

PR #730, opened 2020-03-22
chess activity added

nikitaghps:
Undo, swap, computer vs person, person vs person, updating into journal, multi- browser 
supported.
Frontend-  chessboardjs
Engine- p4wn
Technology- javascript, Html CSS
Capture


ashish0910:
Looks like the activity is not properly integrated with sugarizer . You can follow tutorial [here](https://github.com/llaske/sugarizer/blob/dev/docs/tutorial.md) for better understanding . 
Few other things :

- will be better to have drag and drop rather than current implementation
- will be better to show possible moves highlighted of selected character

nikitaghps:
Implemented drag/drop and character highlighting.

ashish0910:
@nikitaghps Please don't add Activity template to pull request .  
Regarding chess activity , Looks like you have added a mix of Pawn Activity and Chess Activity . Although I am not able to start the game because of console error .
You need to initialise your presence object .

Screenshot 2020-03-24 at 8 10 25 AM

nikitaghps:
I keep getting server internal code 500 when I try to login through my system. Any suggestions on how to fix this problem?

ashish0910:
@nikitaghps can you please share the steps to reproduce the problem ?

nikitaghps:
Sure, when I try to register/login through my system using xampp, I get the following message-


Capture


Although, running the chess activity alone, it looks like this-


Capture2

ashish0910:
Server code 500 is a server related problem . I would suggest you to follow steps mentioned [here](https://github.com/llaske/sugarizer/tree/dev#sugarizer-application)  to setup development environment rather than using xampp .

llaske:
Thanks for your proposal but it's too basic.

PR #717, opened 2020-03-19
created a chess activity

samcaspus:
created a chess activity against comp for Gsoc

PR #735, opened 2020-03-24
added files only related to chess activity to this branch

samcaspus:
As there was some conflict between the files I created another pull request against [#717 ](https://github.com/llaske/sugarizer/pull/717) . In this commit all the files are related to Chess.Activity alone. Hope that meets the requirement @ashish0910


ashish0910:
Good Start . Few Remarks : 
- Game should have multiplayer mode using presence 
- Game should have undo and reset button
- Different difficulty modes can be added
- Drag and drop implementation should be there rather than current implementation
- Use king Pawn icon as activity icon
- Moves history should be displayed
- Possible moves of currently selected character should be highlighted

llaske:
Thanks for your proposal but it's too basic.

PR #740, opened 2020-03-25
[Issue #739 fix] Keyboard overlap fix on Android

### LINKED ISSUE #739: Keyboard overlaps input fields on Android
A major known bug in the Android app is the overlap of virtual keyboard on input fields and text areas. Login screens use a hack (increasing bottom padding) to deal with this although causes entire UI to stay on top.
Other activities such as Get Things Done, Markdown, Writer, etc show this problem once the input focus moves below the keyboard.

I will propose a solution in my next PR.

dhruvmisra:
Fixes #739.
The issue targets the apps that run in fullscreen (like Sugarizer) where resizing of the UI is disallowed. I researched about this issue and found a Cordova plugin which detects the opening/closing of virtual keyboard on fullscreen and provides a workaround for the resizing issue.

*Plugin*: https://github.com/ionic-team/cordova-plugin-ionic-keyboard

This plugin should be added to the APK builder (make_android.sh file):
```
cordova plugin add cordova-plugin-ionic-keyboard@2.2.0
```
I have provided the config needed for resizing and also removed the padding on login screens in this PR.

**Results**

Login:

Markdown:

Get Things Done:

Writer:

Added the required plugin to the sugarizer-apkbuilder repository in llaske/sugarizer-apkbuilder#2. Please review @llaske.

### DIFF
--- a/config.xml
+++ b/config.xml
@@ -16,6 +16,7 @@
 	</author>
 
 	<preference name="Fullscreen" value="true" />
+	<preference name="resizeOnFullScreen" value="true" />
 
 	<preference name="Orientation" value="landscape" />
 
--- a/css/styles.css
+++ b/css/styles.css
@@ -1138,7 +1138,6 @@ body {
 
 .first-namebox {
 	height: 100%;
-	padding-bottom: 100%;
 }
 
 .first-nameline {
@@ -1159,7 +1158,6 @@ body {
 
 .first-serverbox {
 	height: 100%;
-	padding-bottom: 100%;
 }
 
 .first-serverline {
@@ -1253,7 +1251,6 @@ body {
 
 .first-passbox {
 	height: 100%;
-	padding-bottom: 100%;
 }
 
 .first-input {


llaske:
Very smart fix! Thanks a lot.

Two things however:
- Plugins should be added in the `config.xml` file
- Unfortunately it don't work on iOS, so I've limited to Android

dhruvmisra:
Thank you @llaske! I'm glad you liked the fix. I have added the the plugin to `config.xml` as a commit to this branch. I'm not sure if you can merge changes from this branch again so if need be, I'll add another PR to add the plugin to config.xml.
I'll research more on this and see what I can find to fix the iOS version as well. Happy to make Sugarizer better on all platforms.

llaske:
No need to do a new branch. I've fixed config.xml before merging your branch. See [here](https://github.com/llaske/sugarizer/blob/dev/config.xml#L63).

Regarding iOS, the plugin should work but it doesn't. I've tested all `KeyboardResize` and `KeyboardResizeMode` options (specific iOS options mentioned in the documentation) without success.
BTW it's not a blocking issue on iOS because iOS allow the user to scroll the screen by default. So the keyboard doesn't prevent the user to see the whole screen.

dhruvmisra:
I see, that's good. Would still look for a method to completely solve it on iOS as well. Thanks!

llaske:
@dhruvmisra I'm packaging the next Sugarizer version and I've got a weird behavior on the Gridpaint activity on Android (screen is flashing when starting to draw).
I suspect it could be related to this fix. Are you able to test it on your side on a real Android device?

dhruvmisra:
It seems to be working fine on my Android device.



Could this be a device specific issue?

llaske:
Hmmm. Thanks for testing. Did you test on the latest dev branch and with Sugarizer APK builder?
I'm going to do more testing on my side.

dhruvmisra:
Yes I tested it with the latest dev branch. I'll try it on another device too.

llaske:
I've tested on another tablet and it works.
I'm working to find the reason of this issue but it's not related to this fix: removing the cordova plugin don't fix the issue.
Thanks for your help.

llaske:
Issue opened here https://github.com/llaske/sugarizer/issues/847

PR #751, opened 2020-03-28
[Issue #750 fix] Fixed Pomodoro timer on break change

### LINKED ISSUE #750: Pomodoro timer resets on break change
If the timer is paused and the break value is changed, the timer preview resets to the beginning. Although on resuming, the timer starts from the previous paused value. 

pomodoroTimerIssue

The timer should follow either one of the two workflows on **break change**:
1. Reset completely and start from the beginning
OR
2. Persist the old value when paused and resume from the same value

Please suggest the required workflow @llaske. Adding a PR to implement the second workflow however if need be, I can change it to the first one.

dhruvmisra:
Fixes #750. Currently implemented the second workflow.

pomodoroTimerFix

### DIFF
--- a/activities/Pomodoro.activity/js/activity.js
+++ b/activities/Pomodoro.activity/js/activity.js
@@ -302,7 +302,6 @@ function main(Progress, Stopwatch, l10n, color, datastore) {
   this.handleBreakPlusClick = function () {
     if (!this.state.isButtonsDisable) {
       this.state.breakTimerLimit = this.state.breakTimerLimit + 1;
-      resetTimer()
       updateBreakPomodoro();
       saveInDataStore()
     }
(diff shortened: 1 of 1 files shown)


llaske:
That's not the expected fix. If the user want to add 1 minute break and there is only 56s left, we could expect that there is now 1mn56s left.
Plus, it look like than we the user increase work duration, it increase break duration instead.

dhruvmisra:
I think changing 56s to 1min56s is the expected behavior of increasing the work value. Currently it resets the timer. Should I retain the previous value in that too?
I'm not sure if I understand your query here @llaske. Increasing the work value doesn't change the break duration I believe.

llaske:
Here's the current behavior.

Peek 30-03-2020 22-25

The expected behavior is:
- Current time should be 1:57 when the user click on + to add 1mn break
- Current time and break time should not increase when the user click on + to add 1mn work

dhruvmisra:
Implemented the expected behavior in the latest commit. 
Added some null checks too, for example if the user decreases break time from 2 to 1 min with only 50s left, the new break duration will be updated from the next run.

pomodoroTimerFix3

Please review and let me know if I missed something @llaske.

llaske:
That's perfect. Thanks.

PR #763, opened 2020-03-31
New Activity : VEDIC MATHS

abhishektanwar:
Vedic Maths is an activity that focuses on improving the mathematical calculation skills of users. It helps the user to learn various mathematical calculation tricks that can be employed by the user to do faster calculations in an interactive way.


abhishektanwar:
@llaske please share your thoughts on this activity.

Working on: Full Screen functionality and responsiveness of activity.

llaske:
The target for Sugar/Sugarizer is children between 6 and 12. It seems too complex for this target.
If you're interested by a Math activity, I suggest you to the Mind math activity proposal [here](https://github.com/sugarlabs/GSoC/blob/master/Ideas-2020.md#sugarizer-game-activity-pack).

abhishektanwar:
@llaske we can reduce the level of difficulty of tricks that children can learn . If you like the concept of the activity

llaske:
No I think it's too close to the Mind Math activity.

PR #769, opened 2020-03-31
make responsive the toolbar in clock's activity

### LINKED ISSUE #768: toolbar in clock activity not responsive
but here it is not in the case of clock activity.
Screenshot from 2020-03-31 01-31-43

prakashujjwal1010:
- fixed #768 
Screenshot from 2020-03-31 19-58-02

### DIFF
--- a/activities/Clock.activity/css/activity.css
+++ b/activities/Clock.activity/css/activity.css
@@ -82,14 +82,13 @@ canvas {
   z-index: 5;
 }
 
-
 @media screen and (max-width: 677px) {
-	.toolbar {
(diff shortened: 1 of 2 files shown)


activities/Clock.activity/css/activity.css
```diff
@@ -82,6 +82,66 @@ canvas {
   z-index: 5;
 }
 
+
+@media screen and (max-width: 677px) {
+	.toolbar {
```
ashish0910: Please use Tab based Indentation here

llaske:
Not an issue.

PR #771, opened 2020-03-31
making toolbar responsive in game of life activity

### LINKED ISSUE #770: toolbar in game of life activity not responsive
but here it is not in the case of game of life activity.

Screenshot from 2020-03-31 02-54-02

prakashujjwal1010:
- fixed #770 
Screenshot from 2020-03-31 20-01-14


llaske:
Not an issue.

PR #773, opened 2020-03-31
making toolbar responsive in speak activity

### LINKED ISSUE #772: toolbar in speak activity not responsive
but here it is not in the case of speak activity.
Screenshot from 2020-03-31 20-07-16

prakashujjwal1010:
- fixed #772 
Screenshot from 2020-03-31 20-07-45


llaske:
Not an issue.

PR #775, opened 2020-03-31
making toolbar responsive in constellation activity

### LINKED ISSUE #774: toolbar in constellation activity not responsive
but here it is not in the case of constellation activity.
Screenshot from 2020-03-31 01-31-25

prakashujjwal1010:
- fixed #774 
Screenshot from 2020-03-31 20-14-28


llaske:
Not an issue.

PR #781, opened 2020-04-08
Fixed relative links in README.md

dhruvmisra:
Some links in the documentation led to 404 error pages on GitHub. These were:
1. All links in **Activities distribution** section
2. `js/dialog.js` in **Localization** section

Problem:
readmeLinksIssue

These were caused due to some wrong paths and use of relative-path syntax, which isn't required on the root directory.
Fixed the links in this PR.


llaske:
Yes, that's true. Thanks.

PR #783, opened 2020-04-14
Replaced Uglify-js config by Terser config for Grunt

dhruvmisra:
Replaced the grunt config for Uglify-js by Terser which can minify ES6+ syntax. It was discussed [here](https://github.com/llaske/sugarizer-apkbuilder/issues/3#issuecomment-612850952).

The new config was tested and the results were error-free (as far as I could check) with the `keep_fnames` option set. Other [compress options](https://github.com/terser/terser#compress-options) can also be tweaked to obtain slightly better results although the current configuration performs quite well.


dhruvmisra:
I have removed all activity paths and replaced them with a single dynamic reference. It will now cater to all activities It is also pretty future-proof as no management is required if a new activity is added.
The `dest` path has been removed as the minifications will now take place in the output folder.

llaske:
Very cool. Good job.

PR #786, opened 2020-04-21
[Issue #580 fix] Added username check on the first screen

### LINKED ISSUE #580: alert "user already exists" should come earlier
When logging for the first time , a person need to choose images and colors before he gets to know that the user exists , so he would have to do all the procedure once again. I think that the new user should get to know in the first page where he enters his name whether the user already exists or not.

dhruvmisra:
Fixes #580.

Added `checkUsername()` function which sends a call to signupURL with `beforeSignup` flag and handles the response.

Result:
userExistsFix

Please check the corresponding PR on the server too. (llaske/sugarizer-server#266)

### DIFF
--- a/js/firstscreen.js
+++ b/js/firstscreen.js
@@ -257,7 +257,7 @@ enyo.kind({
 			this.step++;
 			if (util.getClientType() == constant.appType && (this.createnew || !this.$.server.getValue())) { // No password for the app when create new or server is null
 				this.step++;
-			} else {
+			} else if(this.createnew) {
 				this.checkUsername(name);
 			}
 			this.displayStep();
--- a/js/firstscreen.js
+++ b/js/firstscreen.js
@@ -257,6 +257,8 @@ enyo.kind({
 			this.step++;
 			if (util.getClientType() == constant.appType && (this.createnew || !this.$.server.getValue())) { // No password for the app when create new or server is null
 				this.step++;
+			} else {
+				this.checkUsername(name);
 			}
 			this.displayStep();
 		} else if (this.step == 3) {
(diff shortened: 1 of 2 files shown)


llaske:
The fix should be compatible with older Sugarizer Server version too.
What do you propose?

dhruvmisra:
Initially I wanted to make a call to the signupURL (un-modified on the server) by passing in dummy data for other parameters. It would return the same error if name already exists. This quickly becomes troublesome as if it's a new username, a new user with the dummy data will be created.

As I dug through the server routes, I found `/api/v1/users?name=__` which might bear some fruit but when I tested it, it requires a token which isn't present before logging in. I don't see any other helpful routes. Anything I'm missing or should look into @llaske?

llaske:
Regarding the UI, I don't like the flash between Name and Password screen. The Password screen should not be displayed before the return of the call.

Regarding the server side, I think using a flag is not a good work around. See my comment on https://github.com/llaske/sugarizer-server/pull/266.

NikhilM98:
> Initially I wanted to make a call to the signupURL (un-modified on the server) by passing in dummy data for other parameters. It would return the same error if name already exists. This quickly becomes troublesome as if it's a new username, a new user with the dummy data will be created.

I think you instead of passing dummy data for other parameters, you should avoid passing other parameters at all, just pass `name` and `beforeSignup`.
- If server returns `Status 401` `Error: Invalid user object` `Code: 2`, means that the Sugarizer Server does not support your fix. --> Show username check on second screen in this case.
- If server returns `exists: true/false`, it means that Sugarizer-Server supports your fix. --> Show username check on first screen in this case.

dhruvmisra:
@NikhilM98, I am currently only sending the `name`, `role` (Required [here](https://github.com/llaske/sugarizer-server/blob/6319475cb73cc2135d9cf3bcc08e677fd422e8b8/api/controller/auth.js#L239)) and `beforeSignup`. I like the idea for backward compatibility by adding this check. I'll add this next.

> The Password screen should not be displayed before the return of the call.

@llaske I'll fix this as well.

dhruvmisra:
@llaske I have fixed the UI flicker between Password and Username screens. 
Added the backward compatibility, the system will use the old workflow if the server hasn't been updated and will switch to the new workflow if the server is updated to support it.

llaske:
Cool. Thanks.

PR #790, opened 2020-04-25
[Issue #789 fix] Splitbar fix in Chrome

### LINKED ISSUE #789: Splitbar in toolbar of Clock activity are half sized in Chrome
Splitbar use to separate buttons in Clock activity are half sized:

image

See the difference with splitbar in Shared notes activity:

image

It works fine in Firefox.

dhruvmisra:
Fixes #789.

The splitbars implemented in the Clock activity use the `<hr>` tags except for the normally used `<div class="splitbar">`. Due to the hr tags being given `display: inline-block`, they create a problem in Firefox where the base of the line starts from the center of the container. Currently, a negative bottom margin was given to fix the problem in Firefox which gave rise to the half-bars in Chrome.

I fixed this issue by setting the `vertical-align` property and fixing margins on the hr tags. Replacing the hr tags with div's can also be implemented but I think this works fine.

Chrome:
image

### DIFF
--- a/activities/Clock.activity/css/activity.css
+++ b/activities/Clock.activity/css/activity.css
@@ -31,6 +31,11 @@ canvas {
   margin: 0;
 }
 
+.toolbar hr {
+  vertical-align: bottom;
+  margin-bottom: 10px;
+}
+
 #main-toolbar #activity-button {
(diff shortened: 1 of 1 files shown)


llaske:
Yes, it works like that.
Thanks.

PR #800, opened 2020-05-08
Vue.js activity template

dhruvmisra:
# Vue.js Activity Template #
This is an Activity Template to create activities in Vue.js. Currently it is an activity named 'ActivityTemplateVue' and runs the basic Pawn activity for testing purposes.

## Structure ##
I have decomposed the major functionalities into importable components which can be found under `/js/components`.

Steps to use any component (Adding the toolbar for example):
1. Import the component using `<script>` tag in `index.html`
```html
  ...
  <script src="js/components/SugarToolbar.js"></script>    //Import the toolbar component
  <script src="js/activity.js"></script>              //Make sure you import activity.js in the end
</body>
```
2. Create and use the component instance in `index.html` (Don't forget the `ref` attribute)
```html
<div id="app">
  <sugar-toolbar ref="SugarToolbar">
    <sugar-toolitem id="add-button" v-on:click="onAddClick" v-bind:title="l10n.stringAddPawn"></sugar-toolitem>
  </sugar-toolbar>
  ...
```

## Benefits ##
* Easy to integrate a specific functionality.
* Decomposed and encapsulated code making it easier to manage. 
* More flexibility in project design.
* Adds a layer of abstraction. New contributors can focus on activity-specific code rather than the working of Sugarizer functionalities.

## Can be improved ##
* Palette workflow
* ~Localization workflow~

This is an initial template structure and is open to changes. All suggestions are welcome.


llaske:
That's not bad.
But it's not exactly what I expected.
In my idea what we call "component" is something generic where no code specific to the activity should be include. Here there is code specific to the Pawn activity in the component files.

More detailed remarks:
- What you call "Journal" is not really the Journal but a mix between activity and initialization
- sendMessage could be a method of the Presence component
- Why not create Vue event on the Presence component to match presence event
- I don't think we need a presence object in activity data since the Presence component should handle everything

Note also that you've mixed different indentation styles in files (TAB/space) and that there is some ES6 syntax (let).

dhruvmisra:
I have the components more generic and handled all the activity specific code outside. 
* Created callback functions and methods to handle data outside the component.
* Emitting Vue events from Presence to handle DataReceived and UserChanged.
* The presence object in activity data is now an instance of presence Vue component, rather than the utility itself.

activities/ActivityTemplateVue.activity/js/activity.js
```diff
@@ -0,0 +1,107 @@
...
+		'toolbar': Toolbar, 'toolbar-item': ToolbarItem, 'localization': Localization, 'tutorial': Tutorial,
+		'presence': Presence, 'journal': Journal
+	},
+	data: {
+		currentUser: {
+			user: {}
+		},
+		full: false,
+		context: null,
+		presence: null,
```
llaske, line 23: I think it's not need

activities/ActivityTemplateVue.activity/js/activity.js
```diff
@@ -0,0 +1,107 @@
...
+			stringAddPawn: ''
+		}
+	},
+	mounted: function () {
+		// this.presence = this.$refs.presence.presence;
+
+		var vm = this;
+		requirejs(["sugar-web/activity/activity", "sugar-web/env", "sugar-web/graphics/icon"], function (activity, env, icon) {
+			// Initialize Sugarizer
+			activity.setup();
```
llaske, line 41: Why not rename your Journal component to an Activity component then implement this in a setup method of the Activity component?

activities/ActivityTemplateVue.activity/js/activity.js
```diff
@@ -0,0 +1,107 @@
...
+			
+			env.getEnvironment(function (err, environment) {
```
llaske, line 44: May be getEnvironment could be a sync method of the Activity component. And the activity could handle by itself the loading of the environment at mount

activities/ActivityTemplateVue.activity/js/activity.js
```diff
@@ -0,0 +1,107 @@
...
+				document.getElementById("user").innerHTML = "<h1>"+vm.l10n.stringHello+" "+environment.user.name+"!</h1>";
+				vm.currentenv = environment;
+			});
+
+			vm.icon = icon;
```
llaske, line 59: Here again you could do a component instead

activities/ActivityTemplateVue.activity/js/components/journal.js
```diff
@@ -0,0 +1,54 @@
+var Journal = {
+  data: {
```
llaske, line 2: Mixed indentation style in this file

activities/ActivityTemplateVue.activity/js/components/journal.js
```diff
@@ -0,0 +1,54 @@
...
+    requirejs(["sugar-web/activity/activity", "sugar-web/env"], function (activity, env) {
+			env.getEnvironment(function (err, environment) {
+				
+				// Load context
+				if (environment.objectId) {
+					console.log("Existing instance");
+					activity.getDatastoreObject().loadAsText(function (error, metadata, data) {
+						if (error == null && data != null) {
+							let context = JSON.parse(data);
+							vm.parent.pawns = context.pawns;
```
llaske: This code is specific it should not be there.

activities/ActivityTemplateVue.activity/js/components/journal.js
```diff
@@ -0,0 +1,54 @@
...
+		this.parent = this.$parent;
+    
+    let vm = this;
+    requirejs(["sugar-web/activity/activity", "sugar-web/env"], function (activity, env) {
+			env.getEnvironment(function (err, environment) {
+				
+				// Load context
+				if (environment.objectId) {
+					console.log("Existing instance");
+					activity.getDatastoreObject().loadAsText(function (error, metadata, data) {
```
llaske: I wonder if all callback should not be implemented as event

activities/ActivityTemplateVue.activity/js/components/journal.js
```diff
@@ -0,0 +1,54 @@
+var Journal = {
+  data: {
+		parent: null,
+	},
+  mounted() {
+		this.parent = this.$parent;
```
llaske, line 9: Think it's not need here

activities/ActivityTemplateVue.activity/js/components/presence.js
```diff
@@ -0,0 +1,101 @@
...
+			});
+		});
+  },
+  methods: {
+    onNetworkDataReceived: function (msg) {
+			if (this.presence.getUserInfo().networkId === msg.user.networkId) {
+				return;  	// Return if data was sent by the user
+			}
+
+			switch (msg.content.action) {
```
llaske: This code is specific if should not be in the component

activities/ActivityTemplateVue.activity/js/components/presence.js
```diff
@@ -0,0 +1,101 @@
...
+						network.onDataReceived(vm.onNetworkDataReceived);
+						network.onSharedActivityUserChanged(vm.onNetworkUserChanged);
+					});
+					console.log(vm.presence);
+					vm.parent.presence = vm.presence;
+        }
+        
+			});
+
+			vm.humane = humane;
```
llaske, line 29: I think we could remove humane stuff here. May be we could add a Popup component to handle all humane features?

activities/ActivityTemplateVue.activity/js/components/presence.js
```diff
@@ -0,0 +1,101 @@
...
+					network.onDataReceived(vm.onNetworkDataReceived);
+					network.onSharedActivityUserChanged(vm.onNetworkUserChanged);
+				});
+				vm.parent.presence = vm.presence;
+			});
+		});
+  },
+  methods: {
+    onNetworkDataReceived: function (msg) {
+			if (this.presence.getUserInfo().networkId === msg.user.networkId) {
```
llaske: Mixed indentation in this file

activities/ActivityTemplateVue.activity/js/components/toolbar.js
```diff
@@ -0,0 +1,73 @@
...
+						vm.$emit(vm.paletteEvent, event);
+					});
+				}
+			});
+		}
+	}
+}
+
+// Toolbar component
+var Toolbar = {
```
llaske, line 39: I wonder if we need a Toolbar component. Only ToolbarItem is a generic component. May be we could be directly all Toolbar code in the main activity file.

activities/ActivityTemplateVue.activity/js/palettes/ballpalette.html
```diff
@@ -0,0 +1,34 @@
+<div class="ball-select">
```
llaske, line 1: Thinks that you could remove this palette and all related images

dhruvmisra: Yes, I'll do that. Will add a palette component later.

activities/ActivityTemplateVue.activity/js/components/toolbar.js
dhruvmisra: It would be a good practice to have the ToolbarItems in their correct container. The ToolbarItems would give us finer control over the buttons and the Toolbar will help us define behaviour of the toolbar. I think the fullscreen methods should be moved to Toolbar component.

activities/ActivityTemplateVue.activity/js/components/presence.js
dhruvmisra: Yes, we can do that.

activities/ActivityTemplateVue.activity/js/activity.js
dhruvmisra: Hmm, interesting. I'll start testing this workflow.

llaske:
My feedbacks:
- I think the version without Event Bus is better. I'm not sure we could have a very complex structure of components in an activity. So I'm not sure is useful.
- As I said, we don't need to reference the activity object, in the Journal object, we could get the sugar-web activity object by a require call.

activities/ActivityTemplateVue.activity/js/components/presence.js
dhruvmisra: Components are more generic now.

activities/ActivityTemplateVue.activity/js/components/presence.js
dhruvmisra: Created the Popup component

activities/ActivityTemplateVue.activity/js/components/journal.js
dhruvmisra: Components are more generic now.

activities/ActivityTemplateVue.activity/js/components/journal.js
dhruvmisra: Removed this property.

llaske:
That's better.

I'm going to look more but here's my first remarks:
- The Toolbar items are defined both in the `toolbar.js` component and in the `index.html`. Items in the toolbar should be defined at one place only. I think it's better if everything is in `index.html`. 
- ToolbarItem are able to handle automatically linked palette (see for example [here](https://github.com/llaske/sugarizer/blob/dev/activities/Calligra.activity/js/toolbar.js#L61)). You don't need to write code for it in `presence.js`
- The test `this.presence.presence` [here](https://github.com/llaske/sugarizer/pull/800/files#diff-2d321821ee927c03050db8743e13b1efR64) should be done by a call to a `isConnected()` method.

dhruvmisra:
> The Toolbar items are defined both in the `toolbar.js` component and in the `index.html`.

Only the buttons which are essential for each activity i.e. Activity, Help, Fullscreen and Stop buttons are present inside the Toolbar component. Also the network button is present which is conditionally rendered automatically if the Presence component is being used. I think it makes for a cleaner `index.html` which should only contain buttons specific to the activity.

>ToolbarItem are able to handle automatically linked palette

Yes you're right, let me try to handle presence palette in ToolarItem.

>The test this.presence.presence here should be done by a call to a isConnected() method.

Yes this will be better. Adding this next.

activities/ActivityTemplateVue.activity/js/components/toolbar.js
```diff
@@ -32,6 +32,16 @@ Vue.component('toolbar-item', {
...
 		}
+
+		if(vm.id == 'network-button') {
+			requirejs(["sugar-web/graphics/presencepalette"], function(presencepalette) {
+				var presencePalette = new presencepalette.PresencePalette(document.getElementById(vm.id), undefined);
+				presencePalette.addEventListener('shared', function () {
+					presencePalette.popDown();
+					vm.$root.$refs.presence.onShared();
+				});
+			});
```
llaske, line 43: You don't need to specify the source code. Just use the `paletteFile`, `paletteClass` and `paletteEvent` in `toolbar-item`. See for example [here](https://github.com/llaske/sugarizer/blob/dev/activities/Calligra.activity/js/toolbar.js#L51).

dhruvmisra: The attributes `paletteFile`, `'paletteClass`, etc are used to create a normal Palette. The presence palette has a different library which needs to be called I believe. Can the presence palette be created with `new palette` instead of `new presencepalette`?

llaske: It's a bit tricky but the `paletteClass` is use to create the object with the right JavaScript class. See the source code of this call [here](https://github.com/llaske/sugarizer/blob/dev/activities/Calligra.activity/js/toolbar.js#L32).
So it will works if you're using `'presencepalette'` value for `paletteClass` property.
In a same way if you're using `'shared'` value for `paletteEvent` property, you will be able to use `v-on:shared` syntax to handle the event.

dhruvmisra: Wow that's clever! I'll test it soon.

activities/ActivityTemplateVue.activity/js/activity.js
```diff
@@ -13,57 +13,50 @@ var app = new Vue({
...
+			// Initialize Sugarizer
+			this.$refs.SugarActivity.setup();
+			this.currentenv = this.$refs.SugarActivity.getEnvironment();
+
+			var vm = this;
+			window.addEventListener('localized', initialLocalization);
+			function initialLocalization() {
+				vm.displayText = vm.SugarLocalization.l10n.get("Hello", { name: vm.currentenv.user.name });
+				vm.SugarLocalization.localize(vm.l10n);
+				window.removeEventListener('localized', initialLocalization);
```
llaske, line 41: Not sure why you need to do that

dhruvmisra: I created this to show the welcome message only once during the activity, and not be affected by any localized event. But I think the event is fired only once during the runtime of any activity. I'll add the message code to the localized method.

dhruvmisra:
> Can you imagine to do it more generic?

Yes I'm currently working to make it more generic. We can send the steps of the tutorial as a prop and extract the `l10n` strings from these steps and handle localization. I had a doubt regarding other methods, Bootstrap Tour API provides a number of functions like `onStart()`, `onEnd()`, `onShow()`, etc., should there be a method for each one of these? As they might be useful in some cases.


> I'm not a big fan of the use of `$nextTick` method. Is there any way to do that?

`$nextTick` function is the recommended way to perform some action right after Vue has completed updating the DOM. (Usually after changing data variables, [see here](https://vuejs.org/v2/api/#vm-nextTick)). An alternative would be to add a small timeout to wait for updation but that would be a bad approach in my opinion.

llaske:
Regarding Tutorial, I don't think we need to encapsulate all bootstrap methods. I think that giving a way to set steps is enough.

Regarding `$nextTick`, I think a better way to do that in Vue.js should be to create a `Pawn` component, handle color as a property of this component then use `v-for` to display all pawns. In that case I think colorize method should not be in the `SugarActivity` component but just call from `requirejs` in the `Pawn` component itselft.

dhruvmisra:
Yes good idea, we can have the Pawn component get the colors as props and set the color in `mounted()`.

As for the Tutorial, I have added basic event emits for these: `onStart()`, `onShow()`, `onEnd()`. I saw these were used in some activities.

activities/ActivityTemplateVue.activity/js/activity.js
dhruvmisra: Made the changes

activities/ActivityTemplateVue.activity/js/components/journal.js
dhruvmisra: Emits an event on data load now.

activities/ActivityTemplateVue.activity/js/components/toolbar.js
dhruvmisra: Added in c428dd8

activities/ActivityTemplateVue.activity/js/activity.js
dhruvmisra: Removed the `initialLocalization()` and made flow better in 06d7134.

activities/ActivityTemplateVue.activity/js/activity.js
dhruvmisra: Created a Pawn component which uses the icon library in a8302fe.

activities/ActivityTemplateVue.activity/js/activity.js
dhruvmisra: Workflow has been changed.

activities/ActivityTemplateVue.activity/js/components/presence.js
dhruvmisra: Should be fixed now.

activities/ActivityTemplateVue.activity/js/components/journal.js
dhruvmisra: Should be fixed now.

activities/ActivityTemplateVue.activity/js/activity.js
```diff
@@ -0,0 +1,107 @@
...
+	},
+
+	methods: {
+
+		onAddClick: function() {
+			let vm = this;
+			this.pawns.push(this.currentenv.user.colorvalue);
+			this.drawPawns();
+
+			if (this.presence) {
```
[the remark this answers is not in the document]
dhruvmisra: Added `isConnected()` to test for presence existence.

llaske:
Good!
I specifically do like your `Pawn` component.

I wonder if we really need a `loaded` event in `SugarActivity`. May be we could remove it and call directly the sugar-web `setup` method in the `mounted` event, so the user don't need to call it (and we could remove the method from the component). 

Minor updates:
- All files should respect the same indentation style (TAB)
- What you call ActivityTemplateVue.activity is the Pawn.activity implemented in Vue.js. In fact, at end you should have two directories: one for Pawn.activity and one for ActivityTemplateVue

dhruvmisra:
Separated the Activity Template from the Pawn activity. Also made some changes to the workflow as per our discussion.
* Moved `activity.setup()` to SugarActivity component
* Removed SugarTutorial's dependency on SugarL10n. Localized strings should be passed in steps now.
* Reworked the insertFromJournal filters; Added support for all type of objects.
* Refactored the code and removed some useless methods.

llaske:
It's perfect. Nice job.

Few (minor) remarks:
- `ActivityTemplateVue` should not be a part of `activities.json` (it's not an activity)
- It will be better to have only on ActivityTemplate directory with two subdirectories `VanillaJS` and `VueJS` (will require to change the text of the initial tutorial to mention that)
- `po` directory should be remove from the activity and from the template
- `README.md` file should be remove from the activity and from the template
- Please update `lib/sugar-web/presencepalette.js` with the latest version in ActivityTemplate in dev branch (this new version automatically handle add/remove users in the palette)
- Please update `LICENSE` file with the latest version in `ActivityTemplate` in dev branch
- Please add `NOTICE` file with the one in `ActivityTemplate` in dev branch
- Update same indentation style (TAB) in all files

dhruvmisra:
Made these changes, updated the tutorial text to reflect the new folder structure.

llaske:
Nice.

Few remarks:
- The default background in the ActivityTemplate should be gray else the background in the tutorial is not the same (white instead of gray on current screen capture)
- For the same reason, the fullscreen button in the toolbar is not in the VanillaJS tutorial, It should not be in the VueJS version
- There is an error in the console when the Pawn activity is shared:

```
TypeError: Cannot read property 'id' of undefined
    at a.sendMessage (SugarPresence.js:44)
    at wn.onAddClick (activity.js:60)
    at He (vue.min.js:6)
    at a.n (vue.min.js:6)
    at He (vue.min.js:6)
    at a.e.$emit (vue.min.js:6)
    at HTMLDivElement.<anonymous> (SugarToolbar.js:31)
    at HTMLDivElement.<anonymous> (pawnpalette.js:18)
```

dhruvmisra:
Fixed the background color and removed the fullscreen buttons. 
I wasn't able to replicate this issue on my side, sharing the activity doesn't show console any error for me. Can you tell me the steps you followed @llaske?

llaske:
Hmmm. Look like it was related to my browser cache.
BTW the background is still white on my side.

dhruvmisra:
Should be fixed now.

llaske:
Thanks.
Not sure where you've updated tutorial text. I've updated `tutorial_step1.md` to mention the new `VanillaJS` directory.

PR #801, opened 2020-05-09
Feature to add a background to Shared Notes activity

### LINKED ISSUE #799: Feature to add a background to Shared Notes activity
It could be nice to add a feature to set the background of the Shared Notes activity.
Currently the background is just white.

image

This new feature should:

- [ ] Add a new icon in the toolbar to select the background from the Journal
- [ ] Set the image as the background
- [ ] Handle resizing of the background image when the board is resized
- [ ] Handle record of the background image when the board is save in the Journal.

sarthak-g:
Fixed #799 

This new feature should:
- [x] Add a new icon in the toolbar to select the background from the Journal
- [x] Set the image as the background
- [x] Handle resizing of the background image when the board is resized
- [x] Handle record of the background image when the board is save in the Journal.


llaske:
Nice.
But the background should be adapted to zoom and move.

sarthak-g:
> But the background should be adapted to zoom and move.

By this mean you mean on zooming, background should also be zoomed?

llaske:
Yes and it should be set at the background position of the top/left of the board and move when the board is move. Using a background during a meeting is something usual, see for example [here](https://www.youtube.com/watch?v=n9B0Wki_HNE). Of course, the image resolution should have a good resolution.

sarthak-g:
ezgif com-video-to-gif (4)
@llaske  made required changes
Is this correct?

llaske:
That's the idea but:
- The background should move when the canvas is moved. Currently only notes are moved
- The background sounds limited to initial size of the board whatever the zoom is. The idea is to zoom out, add a background image with a large size, then to zoom in so the background could cover several screen at normal size

Just a detail but the insert image button should be at the left of the remove button in the toolbar.

sarthak-g:
Raised new pull request (#914 )

PR #805, opened 2020-05-22
Added Vue.js Tutorial

dhruvmisra:
This PR adds the Vue.js tutorial. 

It changes the folder structure to have 2 new directories: `docs/VanillaJS` and `docs/VueJS`. The `docs/tutorial.md` is now used to select the preferred tutorial technology (VanillaJS or VueJS). 

Looking forward to further improvements.


llaske:
Nice.
Here are my first remarks (I need to try to do the full tutorial from scratch to have more):
- There is conflict because I've upgraded the text in your previous PR.
- The new tutorial introduction text to choose between Vue.js and VanillaJS should be improve: not everyone now what is VanillaJS or Vue.JS, you should mention that Vue.js tutorial is for people familiar with Vue.js concept and in case of doubt it's better to choose the VanillaJS version. May be you could say also that the result will be the same at end.
- You should change the screen capture for File Structure in step 1 for Vue.js because new directories are now there (components)
- When you talk about Sugar components in step 1 for Vue.js, it's better to talk about Sugar components for Vue.js
- In step 2, you're explaining the internal of Sugar activity works in **Understanding the working** paragraph, I'm not sure it's useful
- In step 3 for Vue.js your code mention a splitbar but it's not visible in the screen capture, I think it's better to remove it
- In step 4 for Vue.js, it's a bad things that the user can't test itself if it's a new instance or not. I wonder if it's not interesting to send a journal-data-loaded event in the case of a new instance too. Specifically, the console.log message should not came from the component but from the activity itself.
- In Step6, I wonder if you really need to test `if (this.SugarPresence ` is it possible that the SugarPresence object is null/undefined?

dhruvmisra:
After latest commits;

* Resolved the conflict.
* Made the changes in the new tutorial introduction text, please suggest more if needed.
* Added the new screen capture to step 1 in Vue.js.
* Removed splitbar from step 3.

> When you talk about Sugar components in step 1 for Vue.js, it's better to talk about Sugar components for Vue.js

Not sure what you mean here. Should I mention about all components?

> In step 2, you're explaining the internal of Sugar activity works in Understanding the working paragraph, I'm not sure it's useful

I added this mainly to explain about `define` and `require` like in the VanillaJS tutorial. I have removed it as it seems odd to only explain working for one component and not other.

> I wonder if it's not interesting to send a journal-data-loaded event in the case of a new instance too.

I kept this check in the component as it will always be required and the developer will have to add this every time in the the event handler. Do we really need to bring it out?

> I wonder if you really need to test `if (this.SugarPresence  `

It was there to prevent any console errors which might arise due to improper inclusion of the component. Well if the tutorial is followed properly this shouldn't be an issue. I think we can remove it.

llaske:
The tutorial introduction is better. BTW I think it's better to talk about Vanilla JS before talking about Vue.JS.

Regarding Step 1, the idea is just to replace the sentence "contains all Sugar components which..." by "contains all Sugar components for Vue.js which...".

Regarding Journal, the problem is that very often developers need to do a different initialization when activity is loaded from the Journal and when activity is created from scratch. If we don't provide an event to differentiate both it require for developers to initialize from scratch then redo initialization when the `journal-data-loaded` event will be raise. So if suggest to raise an event for new instance too (may be another event if you think it's not pertinent to use the same). I'm also concern about the error since no event is raised when there is an error. 

One more thing: it will be better to put files in `docs/tutorial/VanillaJS` and `docs/tutorial/VueJS` directories instead of `docs/VanillaJS` and `docs/VueJS` directories.

dhruvmisra:
After latest commit:
* Updated the folder structure to `docs/tutorial/{language}`.
* Made the suggested changes to the tutorial introduction and step 1. 
* Added tutorial for new events in SugarJournal added in #806.

prakashujjwal1010:
@dhruvmisra , the image paths and some links are broken ( after you Updated the folder structure to 
`docs/tutorial/{language}`), like `` should be ``, etc. :)

dhruvmisra:
Ah, must have missed these. I'll add a commit to fix it. Thanks! :)

prakashujjwal1010:
very good tutorial :), 
just few comments -
- In step 2, in requirejs.config,  `activity: "../../js"` should be `activity: "../../js"`
- Image paths for background images in css extracts should be `url(../icons/add.svg);` instead of `url(../../icons/add.svg);`
- `locale.ini` is already included in vue template ,so i think we can remove the step showing integration of` locale.ini` in vuejs tutorial.
- In step 5, `this.$refs.SugarActivity.setup();` is giving error , it should be removed,
- In step 6, we should have an instruction  to add `this.SugarPresence = this.$refs.SugarPresence;` in mounted, just like we have for `this.SugarL10n ` in step 5.
- In step 7, css for adding insert-button should be `#main-toolbar #insert-button {
	...
}` instead of `#main-toolbar #picture-button {
	...
}`
- In step 8 , `define([...,"text!pawnpalette.html"]` should be `define([..., "text!activity/palettes/pawnpalette.html"]`
- In step 9, I think we should have a line saying that all the files like bootstrap-tour,...etc are already included in vuejs template
- In step 9, `<script src="lib/jquery-3.4.1.min.js"></script>` should be `<script src="lib/jquery.min.js"></script>` as we have jquery.min.js in our template.
- In step 9, we should have download links for `go-left.svg`, .... etc.

dhruvmisra:
Hey @prakashujjwal1010, thanks for this in-depth review of the tutorial!

I'm sure these relative paths are going to be the end of me. xD I have fixed the paths at places you mentioned. Found some more in step 9, fixed them too.

> `locale.ini` is already included in vue template

I think it was left out while stripping down the Pawn activity to make the template. The goal is to keep the files as similar to the Vanilla JavaScript version as possible. Should we keep the `locale.ini` @llaske?

> In step 9, we should have download links for go-left.svg, .... etc

Hmm, now that I think of it, Vue components should be usable out of the box. That's why we were providing the dependencies (Bootstrap tour, jQuery, etc) already with the template. As the styling for tutorial is already present in `css/activity.css`, I think we should also include the icons `go-left.svg`, ... etc. with the template too. What do you think @llaske?

I agree with all other points and have made those changes. Waiting for more suggestions and fixes. :)

llaske:
Good job @prakashujjwal1010 thanks.

Regarding presence or absence of some files, because we've decided that the VanillaJS and Vue.js are different, there is no reason to force the user to download files even if it's need for the VanillaJS tutorial. Clearly the Vue.js tutorial is easier than the VanillaJS tutorial so it's normal that some files related to Sugar Vue.js components are already there.

So we could keep an empty `locale.ini` file in the template, just change the text to mention that file is already there. It make sense too to provide icons for the tutorial since jQuery and bootstrap are already include.

dhruvmisra:
Sure. I have updated Step 5 mentioning that `locale.ini` already exists. I will add the icons in the next PR along with LZString compression soon.

llaske:
I've tested the full Vue.js tutorial from scratch. I took me some time but I must confess it's not the first activity I've wrote :-)

My remarks:
* Remove the splitbar in index.html in the template else the result don't not match the screen capture
* Step 2, you should detail each time a source code should be updated/modified which file should be change (confusion between index.html and js/activity.js) - except if the same file has just be changed before
* Same remarks in step 3, step 4, step 6 and step 7
* In Step 3, You should mention the full code to create a Pawn component, you don't mention where the mounted should be located. Sure it's obvious for a Vue.js user but it's good to mention it thought.
* In Step 3, for the same reason mention also the location of components property in activity.js file
* In Step 3, you should mention that the pawn.js file should be include before the activity.js file (else it will not work)
* In Step 4, when you integrate the SugarJournal component you should mention that 2 lines are added in the index.html file  (may be you could use comments like in step 6)
* In Step 5, the locale.ini is already integrated in index.html, you should remove the part where you describe how to integrate it
* In Step 5, when you integrate the SugarL10n component you should mention that 2 lines are added in the index.html file  (may be you could use comments like in step 6)
* In Step 5, the Played message is not localized (try with Sugarizer in French). I've tried with your version of Pawn.activity and it doesn't work neither. I think something goes wrong in your SugarL10n component
* In Step 8, all CSS items related to the palette are already in the activity.css file of the template, it should not, it's purely related to Pawn activity

llaske:
I've explored the issue in `SugarL10n` component. It seems related to a second `localized` event received after the one from Sugarizer language change. I suggest you to update the mounted method like that:

```

	mounted: function () {
		var vm = this;
		if (vm.l10n == null) {
			requirejs(["sugar-web/env", "webL10n"], function (env, webL10n) {
				env.getEnvironment(function (err, environment) {
					vm.l10n = webL10n;
					window.addEventListener("localized", function () {
						if (!vm.eventReceived) {
							vm.code = language;
							vm.dictionary = vm.l10n.dictionary;
							vm.$emit("localized");
							vm.eventReceived = true;
						} else {
							if (webL10n.language.code != language) {
							}
						}
					});
				});
			});
		}
	},
```

The `else` part was added, it force the language to the Sugarizer language. So it could happen again.

dhruvmisra:
Made all the required changes to the tutorial text. I'll commit changes to the Activity Template in the LZString PR soon.

> It seems related to a second localized event received after the one from Sugarizer language change.

I was exploring the second `localized` event as well. Thanks I'll update the SugarL10n component. Any reason why you think the second event was being fired with `en-US` (I think default) language code?

llaske:
Thanks.

Regarding the `localized` event, it's raised one time for the browser (the default as you said) and each time a new language is load. So when the language is change like we're doing at start.

PR #806, opened 2020-05-24
Added more event emits in SugarJournal component

dhruvmisra:
Added event emits for new instance and error when loading from journal.


llaske:
Nice. Thanks.

PR #809, opened 2020-05-28
Curriculum Activity

dhruvmisra:
In-progress Curriculum Activity for GSoC 2020

Note: Ignore the `server.bat` file in root, it will be deleted in the end.


llaske:
Nice start. Cool.

My first remarks:
- The toolbar should never be scrolled, only the canvas part should scroll (happen on all screens)
- On categories screen, the category title is too large and the black and white color is not very clear, may be just black will be okay
- On skill screen, the text font color should be black
- On skill screen, the flag position is not ideal. I think it could be better

llaske:
More remarks:
- Could you remove the test categories/skills so we could play with an almost real sample
- I'm not sure if it's good in this activity to use the buddy background color because there is already lot of colors in categories, so I think we could use white instead
- Because categories title could be long, we need to decrease the font size for categories
- I'm no longer sure that using flag analogy is a good idea, it's too large, take too attention and is not so obvious. I think we should go back to checkbox instead. Could you propose something like that?
- The upload icon is too large, some margin should be add around the icon content
- I don't think it's interesting to remind the skill category in the skill detail. The detail screen title should be the skill title itself
- The media zone in the detailed skill should take more place: about half of the screen. Plus it will be nice to have a gray upload icon as background when this no content inside. Something like the journal empty icon in the Journal view
- Back button fill color should be gray instead of black (see for example in Calligra activity)
- Change title of the activity - still My Vue Activity

dhruvmisra:
Hey @llaske, I have added the Medal component and replaced all flags with medals. Also added a new toolitem (with buddy icon) only for testing purposes. Clicking on it will switch between the fixed-colored medals and the buddy-colored medals. This should make it easy to compare and decide which one would be better for the final version.

llaske:
That's very cool! I do like the small animation when the medal is filled. 
I think using the version of medal with buddy colors is better. 
Not sure it's related but there is some error in the console related to colorize.

dhruvmisra:
Yes the error is originating from the Medal component and mainly due to the testing workflow. You can ignore them for now, I'll fix it in the future.

llaske:
Ok. Could you remove buddy icon in the toolbar and use the buddy colors for medal. So I will have a version that I could present to teacher.
Other remarks (lower priority):
- Put the skill image at center (instead at to the left) in the detailed screen
- The detail screen don't work well in smartphone mode

dhruvmisra:
I have updated the activity to use buddy-colored icons. Also changed the skill image placement. Will work on smartphone responsiveness next.

llaske:
Regarding import image, it will be nice to be able to import also directly a drawing coming from the Paint activity. It could be specifically useful for scripting skills where the proof could be a drawing done in Paint activity.
The JournalChooser feature is able to do that, it's done for example in the Fototoon activity. See [here](https://github.com/llaske/sugarizer/blob/dev/activities/Fototoon.activity/js/activity.js#L106).

dhruvmisra:
Added functionality to import drawings from Paint activity.

dhruvmisra:
Added the creation, updatation and deletion of categories and skills. Please review the new workflows.

llaske:
Nice start.

Few comments:
- The icon to change image should be the upload icon instead of the edit icon
- Will be nice to have a default image
- Buttons Confirm/Cancel should use Sugar UI (icons at left)
- It's strange to be able to click on the settings toolitem during editing
- Sorting items is difficult because drag&drop could be done only in on screen, if the user want to move the first category to the end

Not directly related but I'm surprised that categories items could have different size:

image

dhruvmisra:
I have added Curriculum tab to Journal.

> Will be nice to have a default image

Any ideas for this default image?

> Sorting items is difficult because drag&drop could be done only in on screen

If you drag a card to the bottom of the screen, the grid will scroll down automatically. Also I think it'll be better to not show the 3 showcases skills in the settings mode. It will make the category cards smaller and easier to reorder too. What do you say?

> I'm surprised that categories items could have different size

It was due to the flex grow. Added a max and min widths for these cards.

Made all other changes as well.

llaske:
Good. My remarks:

- In the Journal Chooser tab, when the text is too long, it take two lines.
Some encoding issue left too.

image

- The default image for a new skill could be something simple: for example a dark gray medal on a gray background.

- The icon image is not the right one, it should be [this one](https://github.com/llaske/sugarizer/blob/master/docs/images/insert-picture.svg).

- When clicking on a category in edit mode, it should be more logical to go to the skills edit screen. Else the user need to click on the edit button, then click again on the category, then click again on edit button.

- Label "Title", "Color", "Image", "Confirm" and "Cancel" should be localized

- Okay for the sort feature, I suggest to let it like that, without changing categories.

dhruvmisra:
Hey @llaske, I have added localization to categories and skills. The new JSON has the `l10n` variable name in the `title` property. After editing a category/skill, if the title is unchanged (default), it will still be localized but if the title is changed to something else, the new title will be stored for that activity instance (non-localized).
The Curriculum tab in journal chooser is also localized.

*Note:* For testing purposes, I have added Google translated strings for English in `locale.ini`. Try switching the language between English and French.

llaske:
Okay.
It's nice but I think it's too complex. The idea is not to have the opportunity to switch text localization when switching language, the idea is just to ensure that when the activity is open with a new instance, it's fill with a default template using the current user language. So if this same instance is open later in another language, strings should not be localized again.
Plus I find weird than if I'm creating a new category with the title `DefaultSkillSetCategory7` the title is changed by something else :-)

Finally could you clean the `imported` directory? I guess some files inside are no longer need. It could be nice also to rename the JSON file `categories.json` to `default_template.json` and put it at the root of the activity. Then move all images into `images/skills` and remove the `imported` directory.

dhruvmisra:
Changed the folder structure and file names. Made localization simpler, the titles will only be localized in case of a new instance.

> Plus I find weird than if I'm creating a new category with the title DefaultSkillSetCategory7 the title is changed by something else :-)

I think I've taken care of this evil now. :)

llaske:
Cool. That's the idea for Rewards view.
My remarks:
- A+ reward could be get only when notation >= 4
- Rewards icons could be improved, I expect to have a slightly different icon for each reward. Just combine the cup icon with a letter and/or an icon to create an unique reward for each reward
- The popup reward message should integrate the reward icon
- Rewards name should be localized
- Please solve conflict due to recent integration of Chess activity in `activities.json`.

llaske:
Very nice :+1: 
The rewards screen is pretty cool now and the presence handling is exactly what expected. I just wonder if we should not forbid to the hosted users to update users/categories values.

Few remarks regarding the UI:
- Notation palette is very nice now, one little thing: don't do A/B for the level 2 (there is no letter in this case), use a light gray icon and a dark gray icon
- I wonder if we should not allow to change icon state by clicking on the medal itself too (not only on the toolitem)
- Will be better to add splitbar to the right of the presence palette icon
- In detailed view the upload icon arrow should be in the same direction than the one in the palette
- As I said, every reward should have a different icon: add a small text/icon on the cup to remind it's meaning
- A search field should be added in the toolbar to search for a specific category/skill
- It will be better to have a lazy loading for categories else every images are load from the beginning and it could take time on network with limited bandwidth

llaske:
Nice. The lazy loading is pretty cool.

Few remarks:
- The search should not be case sensitive
- On the skill detail screen: the reward toolitem should be to the right of the upload toolitem
- Add a splitbar before the Export toolitem
- I don't think it make sense to see the Export toolitem on the skill detail screen: it could give impression that we could export only the skill. For the same reason I wonder if we should display it on the category detail screen.
- I guess the Export toolitem will become a palette when other export options (PDF, ODT, ...) will be added
- A popup should be displayed at the end of the export
- The name of the Journal entry exported should be "Curriculum Report by {{username}}.txt" (localized)
- In the CSV file the header should be localized

llaske:
The PDF generation is very good. Congrats.

My remarks:
- On stats, will be better to give Acquired percent before Not Acquired percent, order from the higher level to the lower level
- Will be nice to add date under proofs of acquisition
- Look like image proportion are not kept for proofs

image

- Is it possible to keep the name of the template choose at launch? So it could be displayed on the first page of the document and use as title of the export Journal entry. So if the template title is "Kindergarten evaluation", the export first page will display "Kindergarten evaluation" then the buddy icon, then the user name and the entry exported will be "Kindergarten evaluation by <username>.pdf"
- Too bad the you've removed CSV generation, I suggest to transform Export toolitem to a palette. Because CSV mimetype is not supported we could just call it "Export to TXT" but generate a CSV format inside

llaske:
I've detected a bug in Reward screen: when you remove all categories (to create a new template for example), you automatically acquire rewards "50% skills", "All skills", "All categories". Guess it's because you compare to a zero length.

llaske:
When creating a category or a skill, it's always created at the beginning of the list. Should be better to add the new category/skill at end.

dhruvmisra:
Fixed both these issue.

llaske:
My remarks:
- Miss the template for Kindergarten, should be the one imported from JeValide

Capture d’écran du 2020-07-06 22-47-43

- In French there is a black border around the image - due to text size - it should be better to do a white border

Capture d’écran du 2020-07-06 22-50-51

- Image for Montessori school is the one for Primary school and conversely 
- In the template view, the template order should be: Kindergarten, Primary school, Montessori
- Look like Reward popups are no longer displayed
- The exported Journal title should be localized as full string `{{templatename}} by {{username}}` because location of `by` could change in some languages
- When I provide multiple proofs for the same skill, there is an error in ODT generation

image

llaske:
Very nice.
My remarks:
- Remove "." at the end of skills/categories title in the Montessori default template (I've stupidly put it but it's inconsistent with other templates)
- Add margin around Export toolitem icon, it take too much space today
- Remove log messages in the console, specifically "cat: X" each time a category is clicked

dhruvmisra:
I have made the fixes and tested the creation of templates with external images as well (both through URL and by placing JSON in `defaultTemplate` directory). It works as expected.

llaske:
Nice.

Few remarks:

- Be careful with the name use for template file. It's case sensitive on some platforms (Linux, MacOS) so on these platforms, there is an error on the console and the template popup is empty:

image


- The tutorial popup buttons have not the standard colors.

image

image

Should be:

image

image

llaske:
Great works. Thanks a lot. 
This activity is already a legend for teachers!

dhruvmisra:
Glad to be a part of this community! :D

PR #812, opened 2020-05-30
LZString compression and other improvements for Vue.js template

dhruvmisra:
This PR adds:
1. LZString compression/decompression in `SugarJournal` and `SugarPresence`
2. Basic icons required by `SugarTutorial` as discussed [here](https://github.com/llaske/sugarizer/pull/805#issuecomment-636044014)
3. Fixes for `data` property not being a function as required inside `Vue.component()`
4. Fixed a bug in `SugarL10n` causing language to switch back to default (See [here](https://github.com/llaske/sugarizer/pull/805#issuecomment-636482489))
5. Removed splitbar and extra CSS
6. Added `v-visible` directive in `SugarToolbar` which sets the 'visibility' instead of 'display' like `v-show`. It avoids errors and placement issues while using palettes.


llaske:
Nice.

PR #814, opened 2020-05-30
Removed resize option from textarea in Shared Notes

dhruvmisra:
The textarea shows a resize option in the bottom corner but the notes can not be resized.

sharedNotesResize

Removed this option.


llaske:
Thanks. Simple and efficient.

PR #816, opened 2020-06-02
[Issue #815 fix] Fixed Last One Loses after sharing

### LINKED ISSUE #815: Last One Loses activity is unplayable after sharing without second player
Once you share the activity, the computer stops playing and the game is stuck waiting for the second player.

lolNetworkIssue2

Also if the second player leaves during a game, first player is stuck again and can only play the game after restarting the activity.

lolNetworkIssue1

Taken cue from #722

dhruvmisra:
Fixes #815. The user can still play against the computer if the second player isn't connected or if the second player leaves in the middle.

lolNetworkFix


llaske:
Nice. Thanks.

PR #817, opened 2020-06-09
MindMath Activity

prakashujjwal1010:
In progress, MindMath activity for GSOC 2020


llaske:
Congratulation, the new UI is very cool :+1: 
Few remarks:
- Do not use an icon to write "Score", it's typical a string that should be localized
- It will be nice to find an icon for the activity, I suggest a brain with operations inside
- It will be better to start localization now instead of doing it at end

prakashujjwal1010:
- yes, will replace the icon with text.
- okay, sounds good.
- yes,I was just going to start it, ( was working on multiplayer game mode)

llaske:
Nice.
Few UI remarks:
- Sounds like the font is not set, look like the default browser font is use
- Avoid displaying buttons or UI elements not usable. For example all these elements should be hidden instead of disable

image
image

llaske:
Two other remarks resulting to our demo session:
- Do not combine pagination and scrollbar on the result and scoreboard screen. Just use pagination.
- Replace "--" in scoreboard screen when an user is playing by a hourglass icon

llaske:
Few remarks on your latest PR:

- In Result screen, there is an alignment issue with buttons in the footer : back arrow and restart button are not vertically aligned with other buttons

Capture d’écran du 2020-07-01 22-02-41

- In Result screen, the scrollbar is still visible to the right
- I'm not sure to understand the level of difficulty: easy level give higher number than difficult level
- When button in tutorial are disabled, the font/background color use are not the right one

Capture d’écran du 2020-07-01 22-04-01
image

prakashujjwal1010:
well the difficulty level is like( as described [here](https://github.com/sugarlabs/GSoC/blob/master/Ideas-2020.md#sugarizer-game-activity-pack) ):
- an easy level with a chosen number between 10 and 69
- a medium level with a chosen number between 0 and 99

However open to suggestions to change it, we can change the ranges.

llaske:
On remark: in "No Timer" mode it seems that there is no difference between the "Restart" button and the "Pass" button. If so, in this mode it could be interesting to hide the "Restart" button.

prakashujjwal1010:
yes, it sounds good.

llaske:
Thanks a lot for your work.
This activity is very fun and addictive and the UI is pretty engaging.

activities/MindMath.activity/js/Game.js
```diff
@@ -95,7 +95,7 @@ var Game = {
         </div>
         <div class="slots-area-footer">
           <button id="btn-restart"
-            v-if="!disabled"
+            v-if="!disabled && mode=='timer'"
```
llaske, line 98: What happens in multi-player mode? I guess it will hide the restart button too so it will forbid to hosting user to restart the game.

prakashujjwal1010: in multiplayer mode restart button would be in result screen. so it will not have a problem

PR #818, opened 2020-06-10
Updated journalchooser.js to better support feature addition

dhruvmisra:
Exports the `featureLocalJournal`, `features` and `currentFeature` properties and `close()` method to enhance support for feature addition. Also improves `doLocalize()` method.


llaske:
Nice. Thanks. It's very useful to be able to expand JournalChooser.

PR #819, opened 2020-06-11
Updated Fraction to use Vue.js components

dhruvmisra:
Added Vue.js components to Fraction along with the new `SugarDevice` component to handle accelerometer.


dhruvmisra:
@llaske I have updated the activity with latest Sugar components.

llaske:
Nice. Please update with latest components now that there are merged.

dhruvmisra:
Merged the latest dev branch and updated the components.

llaske:
Hmmm. Look like it broke something. On my laptop the ball now continuously try to go to the right of the screen.

dhruvmisra:
Strange, I don't have any issues on my side. Can you try on another device to check if the issue persists?

llaske:
You're right I was not able to reproduce it, even on my initial laptop.
Did you test it also on Android?

dhruvmisra:
Yes I've also tested it on the Android app. It works as expected.

llaske:
I still have the issue on Chromium on Ubuntu.
Here's what happens, the ball go to the right of the screen thought no key is pressed and mouse is not clicked (except for the initial click).

Peek 19-07-2020 21-54

This behavior don't happen on Firefox and don't happen on Chromium on dev.sugarizer.org

dhruvmisra:
Added some debug info (Acceleration x and y values) on the screen and in the console (for Acceleration API). Can you please test the activity again on the same device and send me a screenshot of the values?

llaske:
Thanks.
Here's what happens on my laptop.

Peek 27-07-2020 21-29

dhruvmisra:
This is exactly as we thought. The `y` value stays around 10 and the activity is designed to move the ball if `|y| > 4.75`. It should be due to the different orientation of the accelerometer in the laptop as opposed to other touch devices. Is there a way to detect that we're in this (the laptop) type of device?

llaske:
Hmmm. How much the value is suppose to have? What's the minimum/maximum value?

activities/FractionBounce.activity/js/components/SugarDevice.js
```diff
@@ -33,21 +32,92 @@ Vue.component('sugar-device', {
...
+		},
+		isElectron() {
+			var userAgent = navigator.userAgent;
+			if (/Electron/.test(userAgent)) {
+				return true;
+			}
+			return false;
+		},
+		isWebApp() {
+			if (!this.isMobile() && !this.isElectron()) {
```
llaske, line 50: It's not true, if it's launched using `file:///` it's not mobile, it's not electron but it's not a WebApp neither.
I think you should consider adding a `ifApp` method using the same condition than [here](https://github.com/llaske/sugarizer/blob/dev/lib/util.js#L237).

llaske:
Unfortunately DeviceMotion don't work better on my side.

Peek 03-08-2020 18-44

Plus something is missing in test conditions (see my comment)

activities/FractionBounce.activity/js/components/SugarDevice.js
dhruvmisra: Added the simple `isApp()` method.

dhruvmisra:
The laptop issue should be fixed now. Calling the accelerometer only is the device is mobile. Also it'd be great if the DeviceMotionEvent could be tested on iOS devices to see if it works.

llaske:
That's fine.
Thanks.

PR #820, opened 2020-06-13
Added SugarDevice component to Vue template

dhruvmisra:
Added new `SugarDevice` component with following features:
* Geolocation API
* Accelerometer support (through Cordova)
* Vibration API

Other:
* SugarL10n concurrency fix

### DIFF
--- a/activities/ActivityTemplate/VueJS/js/components/SugarActivity.js
+++ b/activities/ActivityTemplate/VueJS/js/components/SugarActivity.js
@@ -1,4 +1,5 @@
 Vue.component('sugar-activity', {
+	name: 'SugarActivity',
 	data: function() {
 		return {
 			activity: null,
--- a/activities/ActivityTemplate/VueJS/js/components/SugarDevice.js
+++ b/activities/ActivityTemplate/VueJS/js/components/SugarDevice.js
@@ -1,0 +1,63 @@
+Vue.component('sugar-device', {
+	/*html*/
+	template: `<div style="display: none">{{ watchId }}</div>`,
+	data: function() {
+		return {
+      readyToWatch: false,
+      frequency: null
+		}
+	},
+	created: function() {
+		var cordovaScript = document.createElement('script');
+		cordovaScript.setAttribute('type', 'text/javascript');
+		cordovaScript.setAttribute('src', '../../cordova.js');
+		document.head.appendChild(cordovaScript);
+	},
+	mounted: function() {
+		var vm = this;
+    //Accelerometer
+    var useragent = navigator.userAgent.toLowerCase();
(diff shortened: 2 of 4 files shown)


llaske:
Nice work. The inclusion of `cordova.js` is very smart.
Two remarks:
- It seems that some browsers support Accelerometer feature [natively](https://developer.mozilla.org/en-US/docs/Web/API/Accelerometer/Accelerometer). In this case, I think it's better to rely on it, else try to rely on Cordova feature.
- Warning: you mixed TAB/Space indentation style in this file.

dhruvmisra:
I have added the support for native accelerometer whenever it will be available.

Also made changes to SugarL10n to emit the `localized`event after the activity has initialized. 
Working: Added an event listener for SugarAcitivty's `initialized` event which makes `activityInitialized` = true. Added a computed property `readyToEmit` which returns the logical AND of activityInitialized and (vm.dictionary != null) [set after localized event on window], and finally watching this property to emit the `localized` event when true.

llaske:
Regarding accelerometer, do you confirm that the way of using the component will be the same in both case: Cordova and browser? Plus, I wonder it's not better, instead of having an `accelerator-callback` to have a `getCurrentAcceleration` method and let the user of this component decide when he want to get a new value.

Regarding SugarL10n:
- What happens if the activity is initialized before the SugarL10n component has opportunity to run `vm.$root.$refs.SugarActivity.$on('initialized',...` ? 
- It force the user to create the SugarActivity component and give it a reference name `SugarActivity`. It will be better to not force the user to use a specific reference name (or not set a reference name at all). To avoid this constraint, may be we could browse the `$root` property to search for `SugarActivity` component?

dhruvmisra:
If the `accelerometer` object exists, it will be used even in mobile (Not sure about iOS) else the cordova watcher will be added. Let's add the `getCurrentAcceleration` as an additional method as I think the `watchAcceleration` is useful because usually the acceleration is required after a constant interval.

Regarding SugarL10n:
* This situation would not occur as the listener is added directly in the mounted life-cycle hook and being synchronous code, JavaScript Event Loop will complete it's execution before any asynchronous `requirejs` callbacks are resolved.
* In Vue.js the recommended way to access components is through `refs`, so the user should *always* include a ref to access it's methods, but you are right that the user might give some other ref name to the component which can cause issues. Let me see if I can search and attach the listener that way.

dhruvmisra:
I have added a `name` property to SugarActivity and using it to attach the property in SugarL10n, This removes the dependency on ref name altogether.

Not adding the `getCurrentAcceleration` due to the incompatibility with Cordova.

llaske:
Okay, nice.
I guess the tutorial step 5 should be changed to reflect this change.
I suggest to integrate it in the same PR.

dhruvmisra:
Updated the tutorial step.

llaske:
Nice. Thanks. It's better like that.

PR #827, opened 2020-07-01
SugarIcon and other enhancements to Vue components

dhruvmisra:
**Added:**
1. SugarIcon component with `generateIconWithColors()` to get the base64 string and `colorizeIcon()` to set the background (similar to `icon.colorize()`)
2. Promises in place of callbacks wherever possible.

**Updated:**
1. SugarJournal: `createEntry()` to create a new datastore entry.
2. SugarL10n: `timestampToElapsedString()` to convert timestamps to Sugarizer elapsed strings and modified `get()` method which is no longer dependent on webl10n.
3. SugarToolbar: Added `icon` prop which can be used to dynamically set the toolitem icon.
4. SugarTutorial: Code refactoring and fixing icons.


activities/ActivityTemplate/VueJS/locale.ini
```diff
@@ -0,0 +1,207 @@
+[*]
```
llaske, line 1: Now that the `locale.ini` file is no longer empty, the [tutorial step 5](https://github.com/llaske/sugarizer/blob/dev/docs/tutorial/VueJS/step5.md#identify-strings-to-localize) should be updated.

activities/ActivityTemplate/VueJS/js/components/SugarIcon.js
```diff
@@ -0,0 +1,47 @@
...
+		requirejs(["sugar-web/graphics/icon"], function (icon) {
+			vm.icon = icon;
+		});
+	},
+	methods: {
+		colorize: function (element, colors, callback) {
+			this.icon.colorize(element, colors, callback);
+		},
+
+		generateIconWithColors: function (path, colors, callback) {
```
llaske, line 18: Not a big fan of the idea to use callback function here. Other components rely on events. May be it's time here to use promise?

activities/ActivityTemplate/VueJS/js/components/SugarIcon.js
```diff
@@ -0,0 +1,47 @@
...
+			requirejs([`text!${path}`], function (icon) {
+				icon = icon.replace(/(stroke)_color\s\"#?\w*\"/, `stroke_color "${colors.stroke}"`);
+				icon = icon.replace(/(fill)_color\s\"#?\w*\"/, `fill_color "${colors.fill}"`);
+				callback("data:image/svg+xml;base64," + btoa(icon));
+			});
+		},
+
+		colorizeIcon: function (element, colors, callback) {
```
llaske, line 26: What the difference between this method and the `colorize` method.
The name of methods are not obvious: `SugarIcon.colorize()` vs `SugarIcon.colorizeIcon()`...

activities/ActivityTemplate/VueJS/js/components/SugarIcon.js
```diff
@@ -0,0 +1,47 @@
...
+			var path = this.getBackgroundURL(element);
+			this.generateIconWithColors(path, colors, function (src) {
+				element.style.backgroundImage = `url(${src})`;
+				if (callback) {
+					callback();
+				}
+			});
+		},
+
+		getBackgroundURL: function (elem) {
```
llaske, line 36: Do we need to expose this method?

activities/ActivityTemplate/VueJS/js/components/SugarL10n.js
```diff
@@ -75,6 +91,39 @@ Vue.component('sugar-localization', {
 			Object.keys(strings).forEach(function (key, index) {
 				strings[key] = vm.getString(key.substr(6));
 			});
+		},
+
+		// COnvert a UNIX timestamp to Sugarizer time elapsed string
+		timestampToElapsedString: function (timestamp, maxlevel, issmall) {
```
llaske, line 97: It look like to an internal method. May be we could expose an easier method `SugarL10n.localizeTimestamp(timestamp)`?

activities/ActivityTemplate/VueJS/js/components/SugarJournal.js
```diff
@@ -71,5 +71,11 @@ Vue.component('sugar-journal', {
 				}, 0);
 			});
 		},
+
+		createEntry: function (data, metadata, callback) {
```
llaske, line 75: Here again, I'm not in favor of using callback. Events or promise will be more coherent.

activities/ActivityTemplate/VueJS/locale.ini
dhruvmisra: Okay I'll update the step.

activities/ActivityTemplate/VueJS/js/components/SugarIcon.js
dhruvmisra: Yes, we can use a Promise here. Events are useful when we don't need to have a flow after function call and a different method can handle the event (Eg. Acceleration updates).

activities/ActivityTemplate/VueJS/js/components/SugarIcon.js
dhruvmisra: `colorize` is the wrapper for the legacy method `icon.colorize`, I thought about keeping it until the `colorizeIcon()` is perfect. Note that `colorizeIcon()` doesn't use GET request while `colorize()` does.

activities/ActivityTemplate/VueJS/js/components/SugarIcon.js
dhruvmisra: Not really, it is only used by `colorizeIcon()`. Although there is no concept of 'private' methods in Vue templates. We can have a different naming convention for these methods as suggested in the official [Vue Style Guide](https://vuejs.org/v2/style-guide/#Private-property-names-essential). It also shows declaring a private function outside the template object but in our case, that function will then be available everywhere as a global function.

activities/ActivityTemplate/VueJS/js/components/SugarL10n.js
dhruvmisra: Yes we can create a simpler method. By the way, wouldn't it limit the functionality of this method? As we will have to pass default values for `maxlevel` and `issmall`. Maybe we can just add the default values to the parameters of this method itself?

activities/ActivityTemplate/VueJS/js/components/SugarJournal.js
dhruvmisra: Yes I'll convert this to return a Promise as well.

docs/tutorial/VueJS/step4.md
```diff
@@ -110,15 +110,20 @@ Once again, the Sugar-Web environment could help us. When an activity is launche
 
 The good part is, `SugarJournal` handles this for you! It provides some events which you can handle however you want to. The events are:
 1. `journal-new-instance`: Emitted when the activity is a new instance.
-2. `journal-data-loaded`: Emitted when the activity is an existing instance and if there is some data stored in the journal. It returns two parameters: data and metadata.
-3. `journal-load-error`: Emitted when the activity is an existing instance and if some error occurred while loading data.
+2. `journal-shared-instance`: Emitted when the activity is a shared instance.
```
llaske, line 113: I'm not sure it's pertinent to talk about it now because multi-user handling doesn't occur before step 6

docs/tutorial/VueJS/step4.md
```diff
@@ -110,15 +110,20 @@ Once again, the Sugar-Web environment could help us. When an activity is launche
...
+3. `journal-data-loaded`: Emitted when the activity is an existing instance and if there is some data stored in the journal. It returns two parameters: data and metadata.
+4. `journal-load-error`: Emitted when the activity is an existing instance and if some error occurred while loading data.
 
 Let's handle these events from the `SugarJournal` component. Add the following basic methods to `js/activity.js`:
 ```js
 onJournalNewInstance: function() {
 	console.log("New instance");
 },
 
+onJournalSharedInstance: function() {
```
llaske, line 123: Same remarks than before

docs/tutorial/VueJS/step4.md
```diff
@@ -130,7 +135,7 @@ onJournalLoadError: function(error) {
 
 Add bind them at the component instance inside `index.html`:
 ```html
-<sugar-journal ref="SugarJournal" v-on:journal-data-loaded="onJournalDataLoaded" v-on:journal-load-error="onJournalLoadError" v-on:journal-new-instance="onJournalNewInstance"></sugar-journal>
+<sugar-journal ref="SugarJournal" v-on:journal-data-loaded="onJournalDataLoaded" v-on:journal-load-error="onJournalLoadError" v-on:journal-new-instance="onJournalNewInstance" v-on:journal-shared-instance="onJournalSharedInstance"></sugar-journal>
```
llaske, line 138: Same remarks than before

activities/ActivityTemplate/VueJS/js/components/SugarL10n.js
```diff
@@ -54,7 +62,15 @@ Vue.component('sugar-localization', {
 	methods: {
 		// Get a single string with parameters
 		get: function (str, params) {
```
llaske, line 64: What happens when `params` is undefined/null?
Why not merge both `get`/`getString` methods in a simple `get` method?

dhruvmisra: Sure, I think it'll work.

llaske:
Nice. Thanks.

PR #831, opened 2020-07-10
Tangram Activity

prakashujjwal1010:
In progress, Tangram Activity for GSoC 2020.


llaske:
Good start.

Few remarks:
- I'm not sure it's a good idea to use the same color for all pieces. Why not using a different color depending of the type of the piece. So identical pieces will have the same color.
- I think a separation should be done between board and set of pieces. May be a dotted line like in ABCYa Tangram game referenced on the GSoC ideas page:

image

- I think that the title of the Tangram should be closer of the board. May be just under the shape. It's not obvious today to understand it's the name of the shape.
- Is there something wrong on the "arrow" shape? The squared hole in the shape is weird.

image

- The Play/Edit button in the toolbar should be the one button, the button icon changed depending of the current mode.

llaske:
Very good :+1: 

- Regarding Category screen, your view is very nice. But it's too bad that this new view is only reserved to settings. I suggest instead to let the user access to this view in Play mode too to be able to choose the tangram he want to complete. In Play mode, there will be no edit/delete/plus button and clicking on an item just launch the game with this tangram. Replay button will choose randomly a tangram in the category and you could remove this button in edit mode.
- So clicking on an item in the category palette will display the Category screen in Play mode. Of course, this palette will no longer use multi-selection but it doesn't mind. 
- I suggest to not use a palette for difficulty level. Because it's not correlated to the difficulty level but to the play mode (display just the shape or display each piece of shape). So to indicate the mode you could switch between two icons: for example the square tangram with its pieces and the square tangram filled. 
- I was not clear regarding Import/Export data set. You don't need to create buttons for that. If the user create/update its own tangram set, he could just save it in the Journal as usual then copy it to the shared journal or on its device to share it by mail/http/ftp...
- How is it possible to create a new category?
- You should plan to translate default categories/tangram names (of course new categories/tangram names will not be translated)
- Regarding Edit tangram screen, several suggestions:
1. I suggest to add a "Random" button to generate a random tangram. May be the user could try to find a name for it!
2. Remove the Save button: every change should be recorded, at worse the user can restart a new instance of the activity.
3. Prefill the name with "My Tangram", you could remove the label "Enter Tangram Name"; it should be obvious.
4. To go back to the category view, don't use a button in the toolbar, the back button in the view is sufficient
5. Put the category name at the top of the right part of the screen using same color than on the category screen. Don't use label "Category:", it should be obvious
- I'm not sure we need to keep the difficulty level. I don't think it means something for the user. Is there a formula to compute it?

prakashujjwal1010:
Thanks, @llaske.

Regarding creating new activities, we thought we have already enough number of categories including the "miscellaneous" category, user can choose from and also creating new categories would have created the need to place scroll bar in category palette.
Let me add the creating new categories feature then we can have an idea we will keep it if it looks good.:)

llaske:
Good work.

Few remarks:
- When switching from difficulty level, a new game start. It should not.
- The Play button on each Tangram in the Tangrams view is not need, just clicking on the Tangram is okay and obvious.
- The way to create a new category is not ideal and not consistent with other activities like Curriculum/Exerciser. It should be a Plus button in the toolbar that display a form to give name of the category and confirm/cancel buttons.
- For the same reason, using a Combo to select category for the new Tangram is not logical, the process should be to create the category then create Tangrams inside. So the Category should be fixed in this view.
- In Edit mode, play buttons (Random, Timer, Difficulty) should not be visible. Because clicking on it launch the game without really switching to Play mode, it's weird.
- The right order of buttons in the toolbar should be: Activity, Shared, Category, Random, Difficulty, Timer, Settings
- The category palette is still in multi-select mode. As I said, it's not intuitive.
- Valid shape and invalid shape message should be at the same place.

llaske:
The new category form is nice :+1: 
Few remarks:
- I think you don't need the valid/invalid message for category name. It's obvious with the state of confirm/cancel buttons
- When a new category is created, this new category should be the new current one.
- Miss a way to remove a category.

llaske:
Nice.

Regarding the new category form, I think you consider the title valid when it's created. By default, the Confirm button is disable.

image

Also: the Confirm/Cancel label and "New title" should be localized.

llaske:
I've tested the activity on an Android tablet. It works but there is an issue on piece matching.
It seems that you test the matching on the mouse move event. It's fine when there is a mouse but of course on a touch screen this event doesn't happen so the piece is stick until another piece is touched.
See video below:

ezgif com-video-to-gif(2)

May be you could put a timer to test matching?

llaske:
Also, I've got an issue - on the web version.
Not sure how to reproduce it but here the log I had:

image

(Please don't forget to remove debug log)

llaske:
Now works very well. Nice.

There is an issue: a scrollbar appears in the toolbar when the is launched in Electron (see below). I've upgraded Electron version recently to fix a security issue. Could you replicate the fix I've done [here](https://github.com/llaske/sugarizer/commit/bb0afe13c72db62d04bc0b6496efbb65e8200a81) to fix this problem related to Chromium upgrade.

image

llaske:
I've find an issue when reopening a finished game. I've got a strange behavior.
See below:



I've seen it only on iOS. It happens when I'm finished a game, then click back, then click stop and finally reopen the activity. Sound like a mix between finished game and starting game. Could be something related to event synchronization.

llaske:
The Tutorial doesn't work on Electron. There are several errors on the console. See below.

image

I think is related to your HTML file. For Electron/node.js you need to surround integrated JS libraries by:

``if (typeof module === 'object') {window.module = module; module = undefined;}</script>``

and 

``if (window.module) module = window.module;``

Vote activity had the same issue, see fix [here](https://github.com/llaske/sugarizer/pull/833/commits/dd49bd028c8da2f0c94cd6bce17c1fc7fdbe46b6).

llaske:
It look like few strings in the `locale.ini` and in the source code comes from MindMath activity and are not used here.
I've seen:

- `TutoCompulsoryOpTitle`
- `TutoCompulsoryOpContent`
- `TutoTargetTitle`
- `TutoTargetContent`

May be there are some others. Could you check and remove it from `locale.ini` and from the source code?

llaske:
I've got another issue when testing touch on a real iOS device.
Sometimes, some pieces can't be move. See video below where I'm trying to move each piece one by one.
The two last pieces can't be moved.

RPReplay_Final1598779817 2020-08-30 11_37_37

prakashujjwal1010:
@llaske , how frequent this is happening? . let me look into this.

llaske:
About one time on two. So, frequently

llaske:
Another remarks, the loading of categories take lot of time on real device (few seconds).
Could you add a spinner in the background during the loading time?
You could use the same than in e-book reader activity here: https://github.com/llaske/sugarizer/blob/dev/activities/EbookReader.activity/images/spinner-light.gif

llaske:
The spinner is cool but it seems than it's displayed only during loading.
Could you add it also during displaying of a category? It's what took time on some devices.

llaske:
Many thanks for this very cool and addictive activity. Children will love it.

PR #832, opened 2020-07-20
Update tutorial_step2.md

joshuaalm:
Used 'requirejs' (library name) where 'require' (function name) was probably intended.

### DIFF
--- a/docs/tutorial_step2.md
+++ b/docs/tutorial_step2.md
@@ -144,7 +144,7 @@ These lines rely on the framework **require.js** that is used by Sugar-Web to ha
 
 * `define` is a way to define a new module and express its dependencies. Here for example we're going to define a new module that depends on the JavaScript library `sugar-web/activity/activity`. So when the `js/activity.js` is run, **require** will first load the Sugar-Web activity library and put a reference on it in the `activity` variable.
 
-* `requirejs` is pretty the same. It tells to **require**: run the following function but before that, load dependencies and give me a reference to it. There is a small hack here because `domReady!` is a special library used to wait for the end of the HTML page loading.
+* `require` is pretty the same. It tells to **require**: run the following function but before that, load dependencies and give me a reference to it. There is a small hack here because `domReady!` is a special library used to wait for the end of the HTML page loading.
 
 Then comes the most important line of our activity:
 


prakashujjwal1010:
@joshuaalm , requirejs() is the function . so `requirejs` is correct.

dhruvmisra:
Hi @joshuaalm, in RequireJS the keywords `require` and `requirejs` are exactly the same. The creators of the library provided these two aliases because in some environments the `require`  keyword might already be reserved. You can see [this answer](https://stackoverflow.com/questions/13605600/requirejs-difference-between-requirejs-and-require-functions#:~:text=The%20AMD%20spec%20defines%20require,obtain%20a%20reference%20to%20require%20.) or [this commit](https://github.com/requirejs/requirejs/commit/be45948433b053921dc6a6a57bf06d04e13b3b39) for more information.

Sugar activities mainly use `requirejs` keyword, so I think it's correct too.

joshuaalm:
This is helpful, however the paragraph that introduces the definitions, it is explicitly stated that the functions are 'define' and 'require'.  Calling it 'requirejs' in the definition therefore breaks the continuity, and the flow, of a document meant to introduce people with relatively basic HTML and JavaScript experience, to the concepts that requirejs brings to this project. I'm looking forward to learning more about how it works! I only mean to address the ease with which people who wish to create activities can understand this particular introduction to the requirejs library and functions therein.

llaske:
@dhruvmisra is right: we need to use `requirejs` function call because we can't use `require` function in the Electron context.

But you're right, there is a confusion in the sentence before the paragraph you've fixed:

> there are only two functions to understand: `define` and `require`.

It should be:

> there are only two functions to understand: `define` and `requirejs`.

Like that it will be clear.

Warning: you should not send PR on the `master` branch, send your PR on the `dev` branch only. See [here](https://github.com/llaske/sugarizer/blob/dev/CONTRIBUTING.md).

joshuaalm:
Yes, I noticed yesterday that I was making the PR on the master branch and
I will avoid that in the future. Glad to make a new PR on the dev branch
that addresses the lead paragraph, in aligning it with the definitions.
Thank you all for taking the time to patiently explain these things to me.

llaske:
You're welcome.

PR #833, opened 2020-07-20
Vote Activity

dhruvmisra:
Work-In-Progress VoteActivity for GSoC 2020


llaske:
Nice. It's pretty easy to use the activity.

Here are my remarks:
- Sometime colors on graph are too similar:

image

- Icon too show stat to participants is not obvious. I suggest to add an eye on it.
- The back button to go back to the list of poll should be to the left

image

- Not sure to understand the color for caption on rating graph

image

- On graphs, the scale can't be decimal because it's voter count
- Regarding the history view, I suggest to change the background color of the screen (use buddy foreground color) to make the difference with the poll scren

image

- The history view should mention also connected user count (not only voter count)
- The history view should mention the poll timestamp using the Journal format
- The history detail is too similar to the poll screen, could you propose something to avoid a possible confusion? May be changing the header color should be a good start?

image

- The history detail should repeat poll informations: user/voter count, poll timestamp

llaske:
Very cool :+1: 

Few remarks:
- Raised hands should be reinitialized when a new poll start. Currently when a new poll start, hands appears raised for the teacher but not for the user.
- In Rating mode, I think that we need to provide the average value for vote else it's difficult for teacher to estimate the feeling of voters. what do you think?
- The tooltip is "undefined" on rating graph

Capture d’écran du 2020-08-06 17-18-43

- In history detail screen you should display info about vote: #count, #vote, date, ...  may be in the footer
- Template orders should be: yes/no, rating, MCQ, text, images
- Just a reminder: you should implement the autostop poll starting to the second poll
- Look like you're not handling disconnected user (during poll or after a poll)

llaske:
Nice! Settings screen is cool.

Few remarks:
- In settings screen, do not display the poll component as it will be displayed, it take too much space. I think you could just have (in this order): image field, question field, type field, options field, buttons
- When an existing poll with a non-default image is updated, the chosen image is automatically replaced by the default image (bug)
- Order default template like that:  should be: yes/no, rating, MCQ, text, images MCQ. May be we could find more "generic" questions in default to match more common situations.
- When a new Poll is created, I suggest to go back to Play mode automatically so the user could test it immediately
- In history detail screen, the date should be displayed in Journal format and without decoration - yes I know the format is different in generated document :-)
- Is in possible in text graph to adapt dynamically the size of font to try to fit the screen and avoid too much empty space?

image

- In text vote, I think the typed text should be "clean" (to lower case, strip white space before/after). For example "good" and "Good" should be considered the same word in previous example
- A scrollbar appears in the toolbar when Vote activity is launched in Electron (see below). I've upgraded Electron version recently to fix a security issue. Could you replicate the fix I've done [here](https://github.com/llaske/sugarizer/commit/bb0afe13c72db62d04bc0b6496efbb65e8200a81) to fix this problem related to Chromium upgrade.

image

- Users that left activities are not handled (see below) - guess it's a work in progress

image

llaske:
Good.

But I think something goes wrong in graph generation for PDF/ODT.
Pie chart is incomplete since it should be 100% here:

image

Bar graph values are under 1 thought there was 1 user for each value:

image

llaske:
Ok, it works now :+1: 

Percent labels are good but I think they could be larger, bold and black (instead of white) to be more visible on pie charts.

image

llaske:
One other thing: we should manage the case where the organizer leave the activity.

image

In this case voters are all waiting forever :disappointed: 
At least we could display a message to voters and if possible, give an opportunity to the organizer to join again to continue the poll.

llaske:
One remarks on Text poll: will be nice to auto validate the vote of the user when he/she type on ENTER key.

image

More globally it raise the question to know if users could use keyboard to vote on each type of poll?

llaske:
Nice. The activity is pretty usable in fullscreen mode. Both for organizer and for voters.

My remarks:
- Finally H key is not a good idea because in Text poll, if you type a word with a H inside, it raise hand! May be we could use another key or a sequence with a CTRL key
- I'm not sure that you need to call "Poll template" an item in the list. It's just a poll.

image
image
image
image

- Typo in Poll Footer on "Connected" word

image

- Miss Tutorial on Poll Create/Update screen

image

- In Edit mode, add a popup to explain the Play button

llaske:
I've found an error in Electron it seems that the WordCloud component use in the Poll text type doesn't load right and doesn't display.

image

llaske:
Nice, it's fixed.

BTW we should prevent the launch of poll when the user is not connected (when launched via Electron or using file:///). Because once launched, it's impossible to go back to the main screen. See below:

image

I think we should display an error message to say that the user is not connected to a server.

activities/Vote.activity/js/components/SugarPresence.js
```diff
@@ -32,6 +32,14 @@ Vue.component('sugar-presence', {
 	},
 	methods: {
 
+		isNetworkAvailable() {
+			return new Promise((resolve, reject) => {
```
llaske, line 36: I'm not sure we need a promise.
I think to something like:
```
return this.presence != null && this.presence.isConnected();
```

activities/Vote.activity/js/components/SugarPresence.js
```diff
@@ -19,29 +19,25 @@ Vue.component('sugar-presence', {
...
-			})
+		
+		isConnected: function () {
+			return this.presence != null && this.presence.isConnected();
 		},
 
-		isConnected: function () {
-			return this.presence != null;
+		isShared() {
+			return this.presence != null && this.presence.hasOwnProperty('sharedInfo');
```
llaske, line 40: Will be better to call `this.presence.getSharedInfo()` and test if not equal to `null` instead of testing the presence of the internal property.

dhruvmisra: Made the changes.

llaske:
Thanks a lot for this great activity, the most intuitive activity ever!

dhruvmisra:
Had a lot of fun creating this! :) Thank you for your guidance.

llaske:
I've found an issue. 
When trying to update a poll, type and image properties are automatically reinitialized.
See below:

dhruvmisra:
Missed a simple `else` condition here. Adding a PR to fix it.

llaske:
Sound like Confirm/Cancel button are not localized:

image

Could you have a look on it?

dhruvmisra:
Yes you're right, I'll add a PR to fix it.

PR #834, opened 2020-07-22
Added buddy icon and localized strings in SugarPresence popup

dhruvmisra:
Made the `onNetworkUserChanged` method more dynamic:
* If SugarPopup is present: display a popup
* If SugarL10n is present: localize the text inside popup
* If SugarIcon is present: add the buddy icon


llaske:
Sounds good.
Could you localized the `UserJoined` and `UserLeft` values with values from `PlayerJoin` and `PlayerLeft` from [here](https://github.com/llaske/sugarizer/blob/dev/activities/MazeWeb.activity/locale.ini) for `fr`, `es`, `pt`?

dhruvmisra:
Updated the values as per MazeWeb.

llaske:
Nice.

PR #836, opened 2020-07-28
Fixed Curriculum Trophy issues

dhruvmisra:
Fixed the following:
1. First Skill trophy was being acquired on Partially Acquired too.
2. First Category Trophy was being acquired when a new category is created (no skills in the category)


llaske:
Nice. Thanks.

PR #839, opened 2020-08-03
Updated Sugar components in Curriculum and MindMath

dhruvmisra:
Updated the activities to the latest Sugar components, fixing the localization issue in network popups.

@prakashujjwal1010 Please check if the MindMath activity is working fine. :)


activities/MindMath.activity/js/components/SugarPresence.js
```diff
@@ -61,7 +61,6 @@ Vue.component('sugar-presence', {
 				network.createSharedActivity(vm.bundleId, function (groupId) {
 					console.log("Activity shared");
 					vm.isHost = true;
-					vm.$emit('multiplayer-game-started');
```
prakashujjwal1010, line 64: @dhruvmisra ,add this line. After adding this line, the game is working perfectly.

dhruvmisra: Umm, it's probably better not to modify these components as they will be updated quite often. For this use case specifically, I added the `journal-shared-instance` event from SugarJournal. It can be handled similar to `journal-new-instance` handler. We can call the `onMultiplayerGameStarted()` inside that handler.

dhruvmisra:
@prakashujjwal1010 I have made the changes to handle the shared instance using SugarJournal event instead. Please review.

prakashujjwal1010:
@dhruvmisra , its not working :(. actually, when a user shares an activity the game's mode and other things are also changed. so I had added that `multiplayer-game-started` event to indicate that user has shared the activity.
the 'journal-shared-instance' is i think for the users joining the shared instance from join menu. 
so maybe one thing we can do is that we can update the SugarPresence.js to have such an event like `activity-shared` by default for all activities.What do you think? otherwise, you can just add that line :).

llaske:
@prakashujjwal1010 I'm agree with @dhruvmisra that Sugar Vue.js Components should not be changed. We thought precisely these components to be able to ultimately replace Sugar-Web vanilla JS libraries.
Regarding your problem to detect when the user started to share the activity using the presence palette, you could do it by handling the `shared` event from the presence palette like it's done for example [here](https://github.com/llaske/sugarizer/blob/dev/activities/SharedNotes.activity/js/activity.js#L524) in the Shared Notes activity.
@dhruvmisra could you help @prakashujjwal1010 to redefine this event using the Sugar Toolitem components?

I know it could be frustrating to update a working activity but both of you are explorers to show future Sugarizer developers how things should be done using Sugar Vue.js components :fist_oncoming:

dhruvmisra:
@prakashujjwal1010 Ahh, I'm sorry I must have not understood the working the first time.

I think we can have that functionality using SugarToolitem as @llaske suggests. Currently, when the `shared` event is emitted from the presence palette tool item, `SugarPresence.onShared` is called (See [here](https://github.com/dhruvmisra/sugarizer/blob/60f3e528b21e4ccca3715a7837986f622ae12f93/activities/MindMath.activity/index.html#L40)). We can modify this to call a custom intermediate method where we can perform the tasks needed and in the end call `this.SugarPresence.onShared(event, paletteObject)` to keep the existing workflow.

Let me make these changes so it's more clear.

dhruvmisra:
You can view the custom method [here](https://github.com/dhruvmisra/sugarizer/blob/e0ecc7b2c0ed58c3323d76f01fbf31713803538e/activities/MindMath.activity/js/activity.js#L943-L947). I hope it's working now. 😅

prakashujjwal1010:
@dhruvmisra , yes it's now working properly. Thanks for the updation!!

llaske:
Nice. Thanks to both of you.

PR #840, opened 2020-08-11
Updated SugarDevice to add platform methods

dhruvmisra:
Added methods to detect the current platform in an activity: `isApp`, `isWebApp`, `isElectron`, `isMobile`.

### DIFF
--- a/docs/tutorial/VueJS/step6.md
+++ b/docs/tutorial/VueJS/step6.md
@@ -173,21 +173,7 @@ Now, let's update our Pawn activity to integrate presence. Start first by handli
 ```
 The `presence-event` is the name of the event for which you wish to attach a listener for. This `SugarToolitem` detects this and automatically emits a Vue.js event with the same name, when this event is received from the palette. So, when we click the Share button, the `shared` event is received from the palette to the toolitem, and is then emitted as a Vue event to the instance. This will be handled by the `onShared()` method of the component.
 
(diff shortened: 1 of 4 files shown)


docs/tutorial/VueJS/step6.md
```diff
@@ -173,6 +173,22 @@ Now, let's update our Pawn activity to integrate presence. Start first by handli
 ```
 The `presence-event` is the name of the event for which you wish to attach a listener for. This `SugarToolitem` detects this and automatically emits a Vue.js event with the same name, when this event is received from the palette. So, when we click the Share button, the `shared` event is received from the palette to the toolitem, and is then emitted as a Vue event to the instance. This will be handled by the `onShared()` method of the component.
 
+>**NOTE**: If you wish to perform some tasks when an activity is shared, you can add a custom intermediate method as a handler to the `shared` event on this toolitem.
```
llaske, line 176: Hmmm, I think it's too detailed.
The note should explain briefly what is currently done by default in the method. Then just say that it's possible to redefine it a different process is need - without mentioning the detailed code.

Plus your current change prefixed by ">" break the Markdown syntax.

Note also that there is a typo, few paragraphs below `SugarJounral` instead of `SugarJournal`.

dhruvmisra: Made the changes. I'm using `>` only to show a quoted section. Maybe the multiples lines prefixed with `>` were causing the issue.

llaske:
Good. Thanks.

PR #843, opened 2020-08-28
Fixed script imports for Electron and other bugs in Curriculum

dhruvmisra:
Drag-to-reorder for categories/skills will work now in the Electron environment.

Other bugs fixed:
* Template title wasn't being sent to the connected user so the export files had incorrect names.
* Trophies were not visible to the connected user.
* A shared instance, when closed and opened from the Journal opened like the host instance.
* Medal was clickable on the connected user's side allowing change in acquired state.


llaske:
Nice shot. Thanks.

PR #844, opened 2020-09-02
Updated SugarPresence to add isShared() and the new isConnected() methods

dhruvmisra:
Updated SugarPresence component to initialize presence on mount. This allows checking for connection to the network. The new methods:
* `isConnected`: Checks if the activity is connected to the network. (Checks if Web Socket is connected)
* `isShared`: Checks whether the activity instance has been shared with the network.

**NOTE**
Due to modification to `isConnected`, all Vue.js activities utilizing `SugarPresence.isConnected()` should be switched to use `SugarPresence.isShared()` to check if the activity has been shared.

I have made the necessary changes to the tutorial, Curriculum and MindMath activities. 
@prakashujjwal1010 Please confirm that the changes don't break the activity.

### DIFF
--- a/activities/ActivityTemplate/VueJS/js/components/SugarPresence.js
+++ b/activities/ActivityTemplate/VueJS/js/components/SugarPresence.js
@@ -19,21 +19,25 @@ Vue.component('sugar-presence', {
 				// Handle shared instance
 				if (environment.sharedId) {
 					console.log("Shared instance");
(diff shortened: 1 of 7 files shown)


prakashujjwal1010:
The changes don't break the activity. Thanks

llaske:
Nice. Thanks.

PR #845, opened 2020-09-03
Minor fixes and updates to mind math activity

prakashujjwal1010:
after taking inspiration from Tangram activity:
- timer is updated
- adding pulse effect on passing question
- fixing indentation

All the changes are tested and don't break the current activity.


llaske:
Nice. The pulse effect is cool.
Thanks.

PR #846, opened 2020-09-04
Minor fix to Vote's poll edit bug

dhruvmisra:
Fixed the poll editing issue where it resets the poll type.


llaske:
Nice. Thanks.

PR #851, opened 2020-09-19
[Issue #850 fix] Fixed empty context issue on accidental exit of new activity instance

### LINKED ISSUE #850: Leaving Sugarizer app without leaving current activity cause some activities to become empty
It's possible to leave Sugarizer when an activity is still running: by closing the browser or by quitting the app.
In that case when the activity is relaunch, the content of the activity is empty.
Of course it's not a good idea for the user to leave activity on this way because it could lost its current work but having an empty context is not a good experience.

Peek 12-09-2020 22-32

It happens (at least) in Vote, Curriculum, Ebook Reader, Exerciser activities.

dhruvmisra:
Fixes #850.

Fixed the issue by saving the context of the activity right after it is loaded. Added the fix to Curriculum, Vote and Ebook Reader. Exerciser needs to be updated in it's source repository.

### DIFF
--- a/activities/Curriculum.activity/js/activity.js
+++ b/activities/Curriculum.activity/js/activity.js
@@ -764,8 +764,8 @@ var app = new Vue({
 
 		onJournalNewInstace: function () {
 			this.currentView = 'template-selection';
-			// this.importSkills();
-			// this.addAchievementsToUser();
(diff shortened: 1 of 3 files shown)


llaske:
Good job. Thanks.

PR #853, opened 2020-09-21
[Fix #852] fixing small font size problem

### LINKED ISSUE #852: Font is too small in Tangram activity on Safari or iOS
The font use for categories name, tangram name, time, score is too small on Safari or on iOS devices.

See how it appears on Safari:

Capture d’écran 2020-09-20 à 16 07 17

See how it appears on Chrome:

Capture d’écran 2020-09-20 à 16 12 26

prakashujjwal1010:
This PR 
- fixes #852


llaske:
Nice. Thanks.

PR #855, opened 2020-09-23
[Issue #854] Fixed uniquie ID issue on Vote and Curriculum

### LINKED ISSUE #854: Wrong poll is displayed in Vote activity
When there was some creation/remove of poll, it could happens that the wrong poll is displayed when clicking on it.

Steps to reproduce:
- Create a new Vote instance
- Add a new poll labeled 6
- Add a new poll labeled 7
- Remove poll labeled 6
- Create a new poll labeled 8
- Click on poll 8, the poll 7 is displayed instead

dhruvmisra:
Fixes #854.

The issue was caused due to the new poll/category/skill taking it's `id` as the *length* of the corresponding array of that entity. This caused 2 entities to have the same ID.

Fixed the issue by finding the max `id` from the array and setting `maxId + 1` as the new ID for the entity. This makes sure the new `id` is unique. Fixed the issue in both Vote and Curriculum.

### DIFF
--- a/activities/Curriculum.activity/js/Settings.js
+++ b/activities/Curriculum.activity/js/Settings.js
@@ -101,13 +101,13 @@ var CategorySettings = {
 		},
 
 		addCategory: function() {
-			var nextId = this.categories.length;
-			this.category.id = nextId;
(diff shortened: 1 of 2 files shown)


llaske:
Nice. Thanks a lot to be so reactive.

Issue #865, opened 2020-10-03
In FoodChain Activity, the frog currently moves using the keys "I,J,KL" & generates an unexpected behavior 


ricknjacky:
![gif](https://user-images.githubusercontent.com/50985033/94985016-0ac39d00-0570-11eb-87ff-bd43b06ba04a.gif)

The behavior is quite random and to be honest quite confusing, However this is in reference to the edits suggested in #857 
Creating this issue, to bring light on what was additionally wrong with using I,J,K&L

ricknjacky:
Creating a new pull request that solves this and #857 too, waiting for feedback on the same.

ricknjacky:
#866

llaske:
Sounds like the same issue that https://github.com/llaske/sugarizer/issues/857.
So it should be fixed now on the dev branch by https://github.com/llaske/sugarizer/commit/9630d2973641b13e86db7992c97fe3b794f08257. Could you retry with the latest dev branch?

ricknjacky:
closing the issue thanks for the clarification here 
https://github.com/llaske/sugarizer/issues/857#issuecomment-703069302

PR #866, opened 2020-10-03
Fixes #857 and #865

### LINKED ISSUE #857: Allow frog to move with arrow keys in FoodChain activity
In the Play game of the FoodChain activity, it's possible to move the frog using I, J, K, L keys on the keyboard.
It will be better to be able to move it using arrow keys instead.

image

ricknjacky:
Solves #857 and #865 both

### DIFF
--- a/activities/FoodChain.activity/playgame.js
+++ b/activities/FoodChain.activity/playgame.js
@@ -10,14 +10,25 @@ FoodChain.playLevels = [
 
 // Key to use
 FoodChain.playKeys = [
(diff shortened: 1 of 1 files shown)


llaske:
Thanks for your contribution but do not send PR on the master branch, do not send PR to fix multiple issues.
I think this PR is not need because your fix merged in https://github.com/llaske/sugarizer/commit/9630d2973641b13e86db7992c97fe3b794f08257 should already solve the issue.

ricknjacky:
Understood. Closing this request

Issue #867, opened 2020-10-03
Add a popup when image is exported in Moon activity


llaske:
Moon activity allow generation of the current image of the Moon in the Journal by clicking on the dedicated icon in the toolbar (to the right on the screen capture below).

![image](https://user-images.githubusercontent.com/981841/94987323-efec2b00-0565-11eb-941e-e9573535f46d.png)

It will be nice to notify the user when this export is done.
The idea is to add a popup like in the QRCode activity.

![image](https://user-images.githubusercontent.com/981841/94987549-6ccbd480-0567-11eb-827b-4ac5f768e48a.png)

The popup should be displayed using `humane` library and the message should be localized. See [here](https://github.com/llaske/sugarizer/blob/dev/activities/QRCode.activity/js/activity.js#L157) the implementation in QRCode activity.

ricknjacky:
@llaske I'd like to work on this issue can you give me a headstart?

llaske:
Look on the source code I've pointed and do the tutorial on localization it should give you a good introduction. Do not hesitate to ask if you've got more specific questions.

ricknjacky:
![Screenshot (164)](https://user-images.githubusercontent.com/50985033/94992245-45e0c300-05a6-11eb-9566-0a189d3f2fec.png)
I did as you advised me to. I now have an understanding of Localization and I refereed to the QR Code example, for some reason, When I run the activity(MoonActivity) now I am not getting the usual output we get, It's just a blank screen 

Can you please help me out here a bit? What mistake have I done and where should I improve upon?

llaske:
`humane` library is not included in the activity, you should include the `js` file and the `css` file.
Plus you should reference it in the `define` section like explained in the tutorial [here](https://github.com/llaske/sugarizer/blob/dev/docs/tutorial/VanillaJS/step2.md).
Then you will have to create a new string for the message in the `.ini` file.
I suggest you to complete the whole [tutorial](https://github.com/llaske/sugarizer/blob/dev/docs/tutorial/VanillaJS/tutorial.md), it will help you a lot.

ricknjacky:
I will rectify the mistakes and covey the new changes and yes will go through tutorials again 🙏 thanks for guiding 😊

ricknjacky:
![Screenshot (166)](https://user-images.githubusercontent.com/50985033/95009908-0916d980-0643-11eb-8295-dab702fefc88.png)

[...]

PR #876, opened 2020-10-06
Fixing the moon activity

ricknjacky:
The issue #867 finds a resolution to it, in this pull request

### DIFF
--- a/activities/Moon.activity/js/moon-activity.js
+++ b/activities/Moon.activity/js/moon-activity.js
@@ -1,7 +1,7 @@
 define(['activity/data-model', 'activity/draw', 'webL10n', 'sugar-web/env', 'sugar-web/datastore', 'moment-with-locales.min'], function(DataModel, Draw, l10n, env, datastore, moment) {
 
     'use strict';
-
+    requirejs(['humane'], function (doc, humane) { 
     var toggleGridBtn = document.querySelector('#toggle-grid-button'),
         toggleHemisphereBtn = document.querySelector('#toggle-hemisphere-button');
 
@@ -319,7 +319,8 @@ define(['activity/data-model', 'activity/draw', 'webL10n', 'sugar-web/env', 'sug
 			file_size: 0
 		};
 		datastore.create(metadata, function() {
-			console.log("export done.")
+            humane.log(webL10n.get('MoonImageSaved'));
+            console.log("export done.")
 		}, inputData);
     }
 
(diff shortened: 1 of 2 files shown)


llaske:
The activity no longer work, the screen is blank and there is the following error in the console:

image

Look like neither `humane.js` neither `libnotify.css` are included in your PR.

activities/Moon.activity/js/moon-activity.js
```diff
@@ -1,7 +1,7 @@
 define(['activity/data-model', 'activity/draw', 'webL10n', 'sugar-web/env', 'sugar-web/datastore', 'moment-with-locales.min'], function(DataModel, Draw, l10n, env, datastore, moment) {
 
     'use strict';
-
+    requirejs(['humane'], function (doc, humane) {
```
llaske, line 4: That's not correct, `humane` should be added in the `define` method above.

llaske (changes requested):
Some issues to fix.
Please ensure the activity works on your side.

activities/Moon.activity/js/moon-activity.js
```diff
@@ -319,7 +319,8 @@ define(['activity/data-model', 'activity/draw', 'webL10n', 'sugar-web/env', 'sug
 			file_size: 0
 		};
 		datastore.create(metadata, function() {
-			console.log("export done.")
+            humane.log(webL10n.get('MoonImageSaved'));
```
llaske, line 322: `webL10n` is not define in this context. I guess you should use `l10n`

ricknjacky: I did the changes the activity just doesn't load up now :( 
I am struggling to find the bug, please help me out here

activities/Moon.activity/js/moon-activity.js
ricknjacky: sorry for that, did the change no difference

ricknjacky:
Result after I did the requested changes.

image

prakashujjwal1010:
- firstly you need to add `humane.js` ([here](https://github.com/llaske/sugarizer/blob/master/activities/QRCode.activity/lib/humane.js)) in `lib` directory to use it.
- then you need to add `libnotify.css` ([here](https://github.com/llaske/sugarizer/blob/master/activities/QRCode.activity/css/libnotify.css)) in `css` directory 
- add the `humane` in `define` method like `define([........, 'humane'], function(......., humane)`  instead of `requirejs` to

this would solve the problem. test it at your end.

PR #878, opened 2020-10-06
Fixing the Labyrinth activity

ricknjacky:
The issue #869  finds a resolution to it, in this pull request


llaske:
The PR should not include multiple fixes. You should use a different branch for each fix to avoid it.

See my remarks on PR https://github.com/llaske/sugarizer/pull/876.
More specifically:
- Miss humane files into the PR
- Ensure the activity is still working and your fix is effective (the popup should be displayed and no error message in the console)

Issue #870, opened 2020-10-03
Add a popup when image/sound is exported in Abecedarium activity


llaske:
Abecadarium activity allow export of an image or a sound in the Journal by clicking on the dedicated icons in the toolbar (to the right on the screen capture below).

![image](https://user-images.githubusercontent.com/981841/94987767-152e6880-0569-11eb-904b-4b3fa1212b87.png)

It will be nice to notify the user when this export is done.
The idea is to add a popup like in the QRCode activity.

![image](https://user-images.githubusercontent.com/981841/94987549-6ccbd480-0567-11eb-827b-4ac5f768e48a.png)

The popup should be displayed using `humane` library and the message should be localized. See [here](https://github.com/llaske/sugarizer/blob/dev/activities/QRCode.activity/js/activity.js#L157) the implementation in QRCode activity.

ricknjacky:
fixed, waiting for a feedback

ricknjacky:
#879 fixes this

ricknjacky:
![image](https://user-images.githubusercontent.com/50985033/95646151-26d0bc80-0ae3-11eb-8afb-4c64d03737ca.png)


I have to add the humane library in this file and particularly call it in the highlighted part of the code ?

llaske:
Hmmm. In this activity all define are load one time [here](https://github.com/llaske/sugarizer/blob/dev/activities/Abecedarium.activity/js/activity.js#L3) and store in the `Abcd` global variable. I think it's better to do the same for the `humane` library.

RinkiyaKeDad:
Hey @llaske! I'm new to Sugarizer development and would like to take this issue up. @ricknjacky are you still working on it? If yes would you like some help?

llaske:
Hi @RinkiyaKeDad, sure you can complete the current work by @ricknjacky. That's how open source works!

ricknjacky:
>@ricknjacky are you still working on it? If yes would you like some help?

@RinkiyaKeDad  
Sorry for the late reply, I just lost track of this one, and recently I've been busy with exams at my University. 
this issue is solved pls refer #889 { I had some ambiguity pertaining to testing the changes in this activity, though that's solved now}

>I'm new to Sugarizer development and would like to take this issue up. 

Even my experience with the codebase is pretty nascent, maybe we should explore the platform and it's codebase a bit more so that we don't make silly mistakes(like I did above) and help each other out 😃

ricknjacky:
[...]

PR #879, opened 2020-10-06
fixing the Abecedarium activity

ricknjacky:
The issue #870 finds a resolution to it, in this pull request

### DIFF
--- a/activities/Abecedarium.activity/learn.js
+++ b/activities/Abecedarium.activity/learn.js
@@ -3,7 +3,7 @@
 // Collections size on the screen
 var entriesByScreen = 8;
 
-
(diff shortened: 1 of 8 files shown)


ricknjacky:
some files are added from previous commit, will squash them soon
waiting for review :)

llaske:
The activity no longer work, an error message is display in the console:

image

The PR should not include multiple fixes. You should use a different branch for each fix to avoid it.

See my remarks on PR https://github.com/llaske/sugarizer/pull/876.
More specifically:
- Miss humane files into the PR
- Ensure the activity is still working and your fix is effective (the popup should be displayed and no error message in the console)

PR #889, opened 2020-10-10
Fix for Abecedarium activity

ricknjacky:
fix for #870

### DIFF
--- a/activities/Abecedarium.activity/css/libnotify.css
+++ b/activities/Abecedarium.activity/css/libnotify.css
@@ -1,0 +1,104 @@
+html,
+body {
+  min-height: 100%;
+}
+.humane,
+.humane-libnotify {
+  position: fixed;
+  -moz-transition: all 0.3s ease-out;
+  -webkit-transition: all 0.3s ease-out;
+  -ms-transition: all 0.3s ease-out;
+  -o-transition: all 0.3s ease-out;
+  transition: all 0.3s ease-out;
+  z-index: 100000;
+  filter: progid:DXImageTransform.Microsoft.Alpha(Opacity=100);
+}
+.humane,
+.humane-libnotify {
+  font-family: Ubuntu, Arial, sans-serif;
+  text-align: center;
+  font-size: 15px;
+  bottom: 10px;
+  right: 10px;
+  opacity: 0;
+  width: 150px;
+  color: #fff;
+  padding: 10px;
+  background: #000;
+  -moz-transform: translateY(-40px);
+  -webkit-transform: translateY(-40px);
+  -ms-transform: translateY(-40px);
+  -o-transform: translateY(-40px);
+  transform: translateY(-40px);
+}
+.humane p,
+.humane-libnotify p,
+.humane ul,
+.humane-libnotify ul {
+  margin: 0;
+  padding: 0;
+}
+.humane ul,
+.humane-libnotify ul {
+  list-style: none;
+}
+.humane.humane-libnotify-info,
+.humane-libnotify.humane-libnotify-info {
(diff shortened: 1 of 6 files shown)


ricknjacky:
@llaske the activity doesn't seem to work if I 
>In this activity all define are load one time here and store in the Abcd global variable. I think it's better to do the same for the humane library.

ricknjacky:
Ps:- I still haven't called the Abcd for the humane instance and yes I have included humane.js in lib and libnotify too in css

llaske:
> the activity doesn't seem to work if I do the same for humane library

What you mean by "doesn't seem to work"?

> yes I have included humane.js in lib and libnotify too in css

It's not in your PR.

ricknjacky:
I added it in the files locally, tried testing it, the activity didn't load up is what I meant, hence I didn't push the commit 
sorry for not being specific

ricknjacky:
I have added the humane and libnotify files, also stored it in the Abcd global variable.

ricknjacky:
In the previous commit i added a comment , because that was causing the activity to not load in my local machine, Can someone please guide me, as to how I should call the humane which is stored in the Abcd variable? 

image

llaske:
You didn't include link to the `libnotify.css` file in `index.html`.
Plus `l10n` is not defined in this context. You probably have to define it locally like [here](https://github.com/llaske/sugarizer/blob/dev/activities/Abecedarium.activity/play.js#L316).

activities/Abecedarium.activity/play.js
```diff
@@ -313,7 +313,7 @@ enyo.kind({
 		this.$.colorBar.removeClass("themeColor"+this.theme);
 		this.$.colorBar.addClass("themeColor-1");
 		var playMode = this;
-		requirejs(["webL10n"], function(webL10n) {
+		requirejs(["webL10n", "l10n"], function(webL10n,l10n) {
```
llaske, line 316: I don't understand the idea here?

llaske (changes requested):
It doesn't fix the issue. 
You should define `l10n` around the humane log like I mention in my previous comment.

activities/Abecedarium.activity/play.js
ricknjacky: I actually did the change(took it literally,sorry) that you suggested in the previous commit, will rectify it soon

ricknjacky:
The activity doesn't work for some reason still

I have, 
>define l10n around the humane log

done this

llaske:
Did you test it? It doesn't work. Worse: it's no longer possible to export anything :disappointed: 
I think you're going in a wrong direction: the message should be add the message [here](https://github.com/llaske/sugarizer/blob/dev/activities/Abecedarium.activity/entry.js#L126). It's where the data is stored.

ricknjacky:
image

sorry for dragging this issue for so long, but i just cant find a way out here 😔
i have attached a screenshot in that i have called the l10n by just defiing humane and l10n and i also tested using Abcd.humane.log() and it just doesn't fix 😥 

please can you tell me where exactly, i'm getting it all wrong? i really want to know my mistake

llaske:
`define` is an async function, `exportToImage` is expected to have a sync answer.
It's probably the reason of your issue. You should call the `define` inside the `Abcd.datastore.create`.

ricknjacky:
@llaske Extremely sorry for late reply, I just missed out on this one. I rebased all my commits and I ran the activity again, It occurred to me today that I actually had solved this long time back and I have to select a card after clicking on save image/sound button in the activity, after testing it in that manner ig I have solved the issue. I await your feedback on this :)

llaske:
Nice. Thanks.

Issue #894, opened 2020-10-18
Add more fun at end of Calligra


llaske:
When an user complete a letter in Calligra activity. Nothing happens.

![Peek 18-10-2020 22-34](https://user-images.githubusercontent.com/981841/96385165-660da600-1192-11eb-8c64-9ac2531c8324.gif)

It will be nice to add:
- [Confetti](https://github.com/catdad/canvas-confetti) to celebrate it like in Tangram activity
- A button to propose to do the next letter

llaske:
Fixed in https://github.com/llaske/sugarizer/pull/896

PR #896, opened 2020-11-03
Fixed Issue #894

AdithyaKrishnaK:
Added confetti and Try Next button


activities/Calligra.activity/index.html
```diff
@@ -24,6 +24,7 @@
 <script>if (window.module) module = window.module;</script>
 <script src="lib/require.js"></script>
 <script type="text/javascript" src="../../cordova.js"></script>
+<script src="https://cdn.jsdelivr.net/npm/canvas-confetti@1.3.1/dist/confetti.browser.min.js"></script>
```
llaske, line 27: Please copy the confetti file locally so the activity will be able to work offline.

activities/Calligra.activity/js/activity.js
```diff
@@ -147,6 +147,23 @@ var app = new Vue({
...
+			var vm = this;
+			var index = -1;
+			for (var i = 0 ; i < vm.currentTemplate.images.length ; i++) {
+				if (vm.currentTemplate.images[i] == item) {
+					index = i;
+					break;
+				}
+			}
+			index++;
+			if (index === 26) {
```
llaske, line 161: 26 is not always the number of items in the template! Try with letters.

activities/Calligra.activity/js/player.js
```diff
@@ -8,6 +8,7 @@ var Player = {
 				<canvas id="letter"></canvas>
 			</div>
 			<button id="player-restart" class="player-restart"></button>
+			<button id="player-next-letter" class="player-next" v-on:click="nextItem()">Try next</button>
```
llaske, line 11: Will be better to use an icon instead. For example something like https://github.com/llaske/sugarizer/blob/dev/activities/MindMath.activity/icons/pass-icon.svg

llaske (changes requested):
Nice start!

See my comments.
Plus, you should update the tutorial to explain what is this new button.

activities/Calligra.activity/index.html
AdithyaKrishnaK: While adding confetti file with npm I am getting message 2 low severity vulnerabilties found

llaske: Include the same file than the one in Tangram activity, there is no message in Tangram activity

AdithyaKrishnaK:
I am having trouble adding to tutorial. I don't understand where the title and description are stored and how they are accessed here https://github.com/AdithyaKrishnaK/sugarizer/blob/841bc39abf0413f146f80ae76be46aa4fa61ab32/activities/Calligra.activity/js/tutorial.js#L46

AdithyaKrishnaK:
What am I supposed to do for other languages?

llaske:
Good job. Thanks.

PR #900, opened 2020-11-27
Add West Bank and Somaliland

ccosse:
The geojson files define the known countries in English.  The locale.ini file provides translations to other languages, and is also consulted in the case of English.  English is the only language activated in the Sugarizer version, so only necessary to add entries to the [en] and [*] sections.  The .po translation files in the repo tree do not seem to be used, and are not part of the original.


llaske:
English is not the only language but the activity use the language from Sugarizer settings.
So to see another language you need to change it in Sugarizer: go back on home view, right click on buddy icon to display menu then click on settings item then language.

.po files are generated from .ini so you don't need to update it.

ccosse:
Updated West Bank image; updated several missing entries in locale.ini, including correct fr translations for N. Cyprus and West Bank.

llaske:
Thanks.
I've merged it manually [here](https://github.com/llaske/sugarizer/commit/d9dd7bd39017354f28878e0c46da9486cfa2be08) because you've sent a PR on the master branch.

PR #904, opened 2020-12-12
Pop up added to notify user when exported to jounral

b18050:
@llaske In this PR, I have tried to resolve issue #870. Pop is notifying user when export is done.


llaske:
Look good but I can't merge it. You must send a PR on dev branch only. 
See https://github.com/llaske/sugarizer/blob/dev/CONTRIBUTING.md for more.

PR #907, opened 2020-12-15
Pop up added to notify user when exported to jounral

b18050:
Pop up added to notify user when exported to journal .


llaske:
Thanks but in the meantine it was already fixed by another PR: https://github.com/llaske/sugarizer/pull/889

b18050:
@llaske  Thanks.

Issue #895, opened 2020-10-31
Wrong values for inverse trigonometric function in Calculate Activity


saurabhhere:
Below mentioned bugs are achieved from Sugarizer web (http://try.sugarizer.org/)

Getting __wrong value of inverse trigonometric function when angle type is 'degree'__ :

For asin(0.5) expected value is 60 degree but we are getting 0.000152
Also the range for x in asin(x) is from -1 to 1 but calculator is showing value for x =2 also: 

![Screenshot from 2020-11-01 00-45-35](https://user-images.githubusercontent.com/60233336/97787990-96fec980-1bdb-11eb-8b17-6dc45ae36055.png)

Same problems are with acos(x) & atan(x) when angle type is 'degree'.

[Suggestion] In case of tan, calculator is showing finite values for both tan(90°) & tan(π/2), we can either write infinite or not defined instead of showing a finite number.

![Screenshot from 2020-11-01 00-55-29](https://user-images.githubusercontent.com/60233336/97788181-fd381c00-1bdc-11eb-8c37-04f2c49b230b.png)

llaske:
Good shot. Thanks.
And it look like it's only in degree mode.

![image](https://user-images.githubusercontent.com/981841/98444620-fcd6de00-2112-11eb-8e43-36615d3abb5b.png)

@mikklfr do you think it could be related to the automatic degree to radian conversion [here](https://github.com/llaske/sugarizer/blob/dev/activities/Calculate.activity/js/calculate-app.js#L270)?

Parthkhandelwal0:
hey @llaske! is anyone working on this issue? i would like to contribute should i go ahead?

llaske:
Hi @Parthkhandelwal0, you're welcome to work on it.

saurabhhere:
Hi @Parthkhandelwal0, are you working on this issue?

amitchakraborti9:
Hi @llaske,
I just submitted a pull request which fixes the issue without implementing the suggestion.

llaske:
Fixed by https://github.com/llaske/sugarizer/pull/908

PR #908, opened 2020-12-19
Fix for wrong inverse trigonometric values in degree mode

amitchakraborti9:
This pull request fixes issue #895.


llaske:
Good job. Thanks.

PR #909, opened 2020-12-20
Handle canvas responsiveness in FotoToon Acitivity

### LINKED ISSUE #559: Fototoon don't handle window resize

saurabhhere:
Fixes #559 

Preview: 

FotoToon

### DIFF
--- a/activities/Fototoon.activity/js/activity.js
+++ b/activities/Fototoon.activity/js/activity.js
@@ -37,9 +37,9 @@ define(["sugar-web/activity/activity","sugar-web/datastore","sugar-web/env","tex
         var mainCanvas = document.getElementById("mainCanvas");
         var sortCanvas = document.getElementById("sortCanvas");
         // remove 5 more to be sure no scrollbars are visible
-        mainCanvas.height = window.innerHeight - sugarCellSize - 5;
-        mainCanvas.width = mainCanvas.height * 4 / 3;
-        mainCanvas.style.left = ((window.innerWidth - mainCanvas.width) / 2) + "px";
+        mainCanvas.height = window.innerHeight*.9;
+        mainCanvas.width = window.innerWidth*3/4;
+        mainCanvas.style.left  = ((window.innerWidth-mainCanvas.width)/2) + "px";
 
         var previousButton = document.getElementById("previous-button");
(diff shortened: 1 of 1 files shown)


saurabhhere:
Have a look, @llaske

llaske:
Hmmm. I think it don't solve the problem.
The current behavior is to resize images but don't change globes position. So it's not possible to obtain a nice comic when the window is resized.

Peek 21-12-2020 22-46

The correct behavior should be to keep the image to its original size, so the globe will always match.

saurabhhere:
I thought that the user will use it on the same window size while making a single comic.

I think if we keep the image size to the original size and then if we reduce the window size to a lower size (case opposite to shown above) the canvas size will decrease and hence image size will decrease as the image can't exceed the canvas, changing the relative position of globes with respect to images.

Image size is updating here as per the size of canvas keeping the pixels constant.
So if we are making the outside canvas responsive I think it will make the image responsive too.

llaske:
An user could do a comic using a specific window size then change the window size - for example set the browser window to full screen.
The fix should ensure that whatever resolution used at design time the comic will continue to be readable when it will be read in another resolution or another device.
If you think that playing with outside canvas is enough to solve the issue, I'm fine with that. But currently it's not the case, so I'm afraid it need a more complex fix.

llaske:
Close due to lack of activity. @saurabhhere feel free to reopen it if need.

Vishalk91-4:
> Hmmm. I think it don't solve the problem. The current behavior is to resize images but don't change globes position. So it's not possible to obtain a nice comic when the window is resized.
> 
> Peek 21-12-2020 22-46 [ Peek 21-12-2020 22-46 ](https://user-images.githubusercontent.com/981841/102825223-9fef6880-43de-11eb-8261-e0b6038a2e40.gif) [ ](https://user-images.githubusercontent.com/981841/102825223-9fef6880-43de-11eb-8261-e0b6038a2e40.gif)
> 
> The correct behavior should be to keep the image to its original size, so the globe will always match.

@llaske, Can you please explain this point, I couldn't get what did differ from what you want

PR #912, opened 2020-12-25
Added export image option in Paint activity

### LINKED ISSUE #911: Option in Paint Activity to export the current drawing to the journal as an image
It would be a nice feature if we have a button to export the current drawing in Paint activity to the journal as an image ( like we have in Moon Activity) in JPEG or PNG format.
for this, a simple toolbar button would work.
The user/child can later use his/her drawing and can import it to various other activities like Write Activity, Curriculum Activity, etc. from the journal.

saurabhhere:
Fixes #911 

Preview: 

PaintActivity

### DIFF
--- a/activities/Paint.activity/js/activity.js
+++ b/activities/Paint.activity/js/activity.js
@@ -51,6 +51,24 @@ define(["sugar-web/activity/activity","tutorial","webL10n","sugar-web/env","acti
       webL10n.language.code = language;
     });
 
+     // Export as PNG image
+		document.getElementById("save-image-button").addEventListener('click', function() {
+			var mimetype = 'image/png';
+			var inputData = document.getElementById("paint-canvas").toDataURL(mimetype, 1);
+			var metadata = {
+				mimetype: mimetype,
+				title: "Image Paint",
+				activity: "org.olpcfrance.MediaViewerActivity",
+				timestamp: new Date().getTime(),
+				creation_time: new Date().getTime(),
+				file_size: 0
+			};
+			datastore.create(metadata, function() {
+				humane.log(webL10n.get('PaintImageSaved'))
+				console.log("export done.")
+			}, inputData);
+		});
+
     document.getElementById("stop-button").addEventListener('click', function(event) {
       var data = {
         width: PaintApp.elements.canvas.width / window.devicePixelRatio,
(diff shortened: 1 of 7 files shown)


activities/Paint.activity/js/activity.js
```diff
@@ -51,6 +51,24 @@ define(["sugar-web/activity/activity","tutorial","webL10n","sugar-web/env","acti
       webL10n.language.code = language;
     });
 
+     // Export as PNG image
+		document.getElementById("save-image-button").addEventListener('click', function() {
```
llaske, line 55: Respect indentation. SPACE instead of TAB here.

activities/Paint.activity/js/activity.js
```diff
@@ -51,6 +51,24 @@ define(["sugar-web/activity/activity","tutorial","webL10n","sugar-web/env","acti
...
+			var mimetype = 'image/png';
+			var inputData = document.getElementById("paint-canvas").toDataURL(mimetype, 1);
+			var metadata = {
+				mimetype: mimetype,
+				title: "Image Paint",
```
llaske, line 60: Will be better to call it "Paint by {username}" so localize the title and integrate the user name as variable.

llaske (changes requested):
Thanks.

You should also handle responsiveness. The Stop button should always be visible. So the new button should be hidden when there is no enough space  to display it.

image

You should also update the tutorial with a new step to describe this new button.

activities/Paint.activity/js/activity.js
saurabhhere: In my Visual studio code, it is showing proper indentation. This problem occurs only when I push code to GitHub, due to different tab sizes used by GitHub and Visual Studio.
Thanks for pointing, I will cross check it in the next commit.

activities/Paint.activity/js/activity.js
saurabhhere: Sure!

llaske:
Nice.
But there is still indentation issue in `activity.js`.
See how it appears in Atom:

image

saurabhhere:
In my recent commit,

This is what I pushed from Visual Studio:

Screenshot from 2020-12-29 11-31-49

And after pushing, this is what it is showing on my forked repo:

Screenshot from 2020-12-29 11-32-04

Can you please confirm for me that it is showing proper indentation in your code editor also?
If not can you help me with what may be the reason for this?

Thanks

llaske:
It seems that Visual Studio Code don't display difference between TAB and SPACE.
Some TAB were still there at the beginning of updated lines.
I've fixed it before merging.
Thanks for your PR.

Issue #826, opened 2020-06-30
Is it possible to use JavaScript ES6 syntax in activities ?


llaske:
Previously, JavaScript ES6 syntax was prohibited in Sugarizer and in Sugarizer activities.
Is it still the case?

llaske:
After a bunch of testing (see [here](https://github.com/llaske/sugarizer/tree/feature/es6)) and a fix on minify process (see [here](https://github.com/llaske/sugarizer/pull/783)), the ES6 syntax is now officially supported in source code for Sugarizer and for activities.

The only exception is the [ES6 modules feature](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Modules) that is not supported on Chrome on `file:///` - and fail also for the same reason on Electron (Linux/MacOS/Windows) and Android.

dhruvmisra:
During our discussion on the Knowledge Activity pack, you mentioned about some incompatibilities with iOS as well. Do those still exist?

llaske:
@dhruvmisra No, I've tested on all platforms so it works also on iOS.

dhruvmisra:
Wonderful! Finally we can use Promises. 😍 It's great milestone for Sugarizer.

PR #914, opened 2020-12-26
Feature to add background image in SharedNotes activity (Issue #799)

### LINKED ISSUE #799: Feature to add a background to Shared Notes activity

sarthak-g:
Fixed #799 

https://user-images.githubusercontent.com/42293606/103151860-f4f8f900-47a7-11eb-9307-ce6eb6ebc346.mp4


llaske:
Nice work. It's a good start.
I think something goes wrong with the zoom (see below).

Peek 28-12-2020 22-42

Plus I'm not sure why you changed the content of `sugar-web` directory. May be your dev branch is not up to date?

sarthak-g:
https://user-images.githubusercontent.com/42293606/103280607-d9376200-49f6-11eb-8e82-5557b31493cb.mp4

I've added files in sugar-web directory for journal functionality (journalchooser.js).

llaske:
Good job.

Two remarks:
- I'm not sure we should adapt the initial size to the screen size (`cover` option), I think it's better to use the original image size
- How is it possible to remove the background? May be we could add a custom sheet in Journal popup with all standard colors. So the user could choose to go back to a white background or fill it with another color. See Curriculum activity [here](https://github.com/llaske/sugarizer/blob/dev/activities/Curriculum.activity/js/CurriculumChooser.js) to understand how it works.

sarthak-g:
@llaske 
- Removed cover option
- Added button to add standard colour to background and save it into the journal.

https://user-images.githubusercontent.com/42293606/104046558-47093800-5206-11eb-906f-d09cc52417a9.mp4

llaske:
Hmmm. Removing `cover` doesn't seems to change anything.
When including an Abecedarium (image size 210x210) it take all the screen.
Could you try others options?

Peek 10-01-2021 22-35

Plus it seems that the image is not drawn at the top left of the current screen.

Peek 10-01-2021 22-34


Regarding color selection in a new sheet, good job.
Two remarks anyway:
- It will be better to localize the new sheet at the second position (just after the Journal sheet)
- Could you localize color names?

sarthak-g:
@llaske made changes. Please review.

https://user-images.githubusercontent.com/42293606/104215861-5f27c400-545f-11eb-9566-40ce678540f5.mp4

llaske:
Good work @sarthak-g 

The image has now the right size. But it's always at top/left start position. It should be at top/left of the current position. See below.

Peek 12-01-2021 22-19

Another issue: the share action no longer work. I guess the handling of presence should be updated to take into account the new background stuff.

sarthak-g:
> The image has now the right size. But it's always at top/left start position. It should be at top/left of the current position. See below.

I've kept that to keep the background in synchronization with the initial pan shift but I've updated in latest commit.

> Another issue: the share action no longer work. I guess the handling of presence should be updated to take into account the new background stuff.

Latest commit resolved it. 
If background image is added in shared mode then should it be visible on all screens?

llaske:
> I've kept that to keep the background in synchronization with the initial pan shift but I've updated in latest commit.

Nice. It works now.

> Latest commit resolved it.

Hmmm. Don't works on my side. See below.

Peek 14-01-2021 22-02

> If background image is added in shared mode then should it be visible on all screens?

Yes background should be in sync on all screens.

sarthak-g:
@llaske 
Latest commit resolves

- error in sharing
- synchronization of background

See below:

https://user-images.githubusercontent.com/42293606/104818269-9bd33100-584c-11eb-95be-fb4aacd178a0.mp4

sarthak-g:
Right now in shared mode, if one note is moved then it is reflected on connected users but movement of screen(i.e pan event) does not reflect on another screen. See below:

https://user-images.githubusercontent.com/42293606/104818396-5e22d800-584d-11eb-9cb2-0694eecce72b.mp4

therefore I've not enabled movement of image on another screen. Do I have to do that?

llaske:
Each user should be able to move its view but the background should be at the same place on every view. So I guess that you should pass the x-y shift from top/left (zoom?) to be sure to rightly set the background for each user.

Plus, you've got updated the `journalchooser.js` file. You must not, it should be the same for all activities.

sarthak-g:
@llaske 
In latest commit:

- background image is at same point on all shared screens and when load existing instance
- Shift background color code from journalchooser.js to new file

llaske:
Nice work.

One more issue: Journal entries created before your PR can't be open now.

Peek 21-01-2021 21-48

sarthak-g:
> One more issue: Journal entries created before your PR can't be open now.

@llaske , it was because of change in datastore object. I've updated it in latest commit.

llaske:
Good.

One more thing is missing: the background is not there when the content is exported to the Journal as a PNG.

Peek 23-01-2021 15-07

sarthak-g:
https://user-images.githubusercontent.com/42293606/105628872-95087780-5e65-11eb-869d-a1c8f6df493b.mp4

@llaske , it is properly working now.
It will also work for older browser versions despite of use of promise in html2canvas because of loading of es6-promise polyfill before html2canvas library.

llaske:
Cool.
BTW you don't need to use es6-promise because ES6 syntax is now supported on all platforms. See https://github.com/llaske/sugarizer/issues/826

sarthak-g:
Great, removed file in latest commit.

llaske:
Great job. Thanks a lot for your work and your perseverance.

PR #917, opened 2020-12-30
Updated 'tutorial_step1_1VueJS.png' image in 'docs/images' folder

prabhsharan36:
The 'tutorial_step1_1VUEJS.png' image was outdated because now the activity template have 'icons' folder added to it (referring to VueJS activity template). So the new screenshot image is now updated according to the activity template file structure.

tutorial_step1_1VUEJS


llaske:
Hmmm. I don't think it make a meaningful difference.
Plus your capture don't show the `Pawn.activity` directory like the original one.
Also files `package-lock.json` and `.vs/slnx.sqlite` are not related to your change.

llaske:
Close due to lack of activity. @prabhsharan36 feel free to reopen it if need.

PR #922, opened 2021-01-15
Added country flag in Color My World activity

### LINKED ISSUE #920: Add country flag in Color My World activity
Color My World display country name at the bottom right in Coloring mode:

Peek 13-01-2021 21-56

and in a popup in Tour/Interactive mode.

Peek 13-01-2021 21-57

It would be nice to add to the name, the flag of the country.
A simple solution could be to use emoji flag matching the country (when exist).

saurabhhere:
Fixes #920 

Preview of fix:

ColormyWorld1

ColorMyWorldBottom


llaske:
Nice. That's the idea.
But I'm not a big fan of your solution to change name in all localization.
The flag is always the same whatever the current language is. Plus we cannot ask to translator to include the flag in its translation.

saurabhhere:
We can also have the same result by creating a different js file that contains all maps with their corresponding country and then we can define them in our existing activity file to use for Tour and Colouring Mode.

Preview of js file for flags:

Screenshot from 2021-01-17 01-37-19

And then we can use flag.js like this inside existing js files:

```
flag[`${target_name.replace(/ /g,'_')}`]
```
to show the corresponding flag in Tour and Colouring Mode.

If this fix looks good to you I will push the changes

llaske:
Sounds good. 
Be careful to handle localization. The country name in your map should be the unlocalized name.

saurabhhere:
Please have a look @llaske

llaske:
Good job. Thanks.

Issue #921, opened 2021-01-13
Implement a full 24h mode in Clock activity


llaske:
A new feature https://github.com/llaske/sugarizer/issues/903 of Clock activity added a 24h mode to display 24h value for hours.

The idea is to improve this mode to:
- Use this settings for digits too
- Change background of the clock for PM hours (may be from white to black)
- Allow to pass to AM and PM when rotating the clock arrows
- Ask questions for AM and PM in play mode
- 
![Peek 13-01-2021 22-14](https://user-images.githubusercontent.com/981841/104511333-d5a5fc80-55ec-11eb-993c-13967556b5ce.gif)

saurabhhere:
For PM hours, do we need something like this?

![Screenshot from 2021-01-21 18-38-26](https://user-images.githubusercontent.com/60233336/105355324-fc30ec80-5c17-11eb-8196-bdb0248bc610.png)

llaske:
That's the idea.
Did you change the hours colors? It seems brighter than the AM version. I think it's better if colors are the same in AM/PM mode.

saurabhhere:
No, I haven't updated the hours colors. Those are the same as before. Please have a look at #923

llaske:
Fixed in https://github.com/llaske/sugarizer/pull/923

PR #923, opened 2021-01-23
Implement a full 24h mode in Clock activity

saurabhhere:
Fixes part of #921 

Updated Background to Black and boundary to white for PM hours.

### DIFF
--- a/activities/Clock.activity/js/activity.js
+++ b/activities/Clock.activity/js/activity.js
@@ -471,8 +471,15 @@ define(["sugar-web/activity/activity","sugar-web/env","sugar-web/graphics/radiob
             var lineWidthBackground = this.lineWidthBase * 4;
             ctx.lineCap = 'round';
             ctx.lineWidth = lineWidthBackground;
-            ctx.strokeStyle = this.colors.black;
-            ctx.fillStyle = this.colors.white;
+            var date = new Date();
+            var hours = date.getHours();
+            if (hours<=12){
+              ctx.strokeStyle = this.colors.black;
+              ctx.fillStyle = this.colors.white;
+            } else {
+              ctx.strokeStyle = this.colors.white;
+              ctx.fillStyle = this.colors.black;
+            }
             ctx.beginPath();
             ctx.arc(this.centerX, this.centerY,
                     this.radius - lineWidthBackground - x, 0, 2 * Math.PI);
--- a/activities/Clock.activity/js/activity.js
+++ b/activities/Clock.activity/js/activity.js
@@ -136,6 +136,9 @@ define(["sugar-web/activity/activity","sugar-web/env","sugar-web/graphics/radiob
             this.isDrag = false;
             this.isSmiley = false;
             this.initiateAngles = false;
+            this.toggleAMPM = false;
+            this.firstQuad = false;
+            this.fourthQuad = false;
 
 
             // These are calculated on each resize to fill the available space.
@@ -388,8 +391,23 @@ define(["sugar-web/activity/activity","sugar-web/env","sugar-web/graphics/radiob
             var hours = date.getHours();
             var minutes = date.getMinutes();
             var seconds = date.getSeconds();
-
-            this.displayTime(hours, minutes, seconds);
+            if (show_am_pm){
+               this.displayTime(hours, minutes, seconds);
+            } else {
+               if (hours==0) {
+                  var txt = " AM"
+                  this.displayTime(hours+12, minutes, seconds, txt);
+               } else if (hours < 12) {
+                  var txt = " AM"
+                  this.displayTime(hours, minutes, seconds, txt);
+               } else if (hours==12) {
+                 var txt = " PM"
+                 this.displayTime(hours, minutes, seconds, txt);
+               } else {
+                  var txt = " PM";
+                  this.displayTime(hours-12, minutes, seconds, txt);
+               }
+            }
             this.displayDate(date);
 
             this.handAngles.hours = Math.PI / 6 * (hours % 12) + Math.PI / 360 * minutes;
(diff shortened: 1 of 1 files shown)


llaske:
That's the idea but the background should be change only in 12/24 mode.

saurabhhere:
When 12/24 mode is clicked it shows 24 hours on Clock like this: 

Screenshot from 2021-01-24 15-43-20

Can you please explain whether we need it when Clock is showing upto 24 hours as shown above or when it is showing only upto 12 hours.
I think when it's in 12 hours mode we have to apply these changes and in 24 hours mode background will always remain white. Please confirm.

saurabhhere:
Changes made:
1. Background will remain white in 24 hours mode and white/black in 12 hours mode.
2. Time shown at the bottom will be updated in AM/PM in 12 hours mode and remain same in 24 hours mode.

Preview of changes: 

ClockActivity

If these changes look good to you, I will update the PR.

llaske:
Sounds good.
BTW as mentioned at https://github.com/llaske/sugarizer/issues/921
24h mode should be use also to:
> * Allow to pass to AM and PM when rotating the clock arrows
> * Ask questions for AM and PM in play mode

saurabhhere:
Regarding:
 > * Allow to pass to AM and PM when rotating the clock arrows

I have toggled AM and PM whenever the clock passes angle 0. But in that case, it is only updating from AM to PM or PM to AM when the drag is smooth so that the clock is able to detect when the angle becomes 0.
 
Preview of what I explained above:

ClockSetTime

I am searching for a better alternative to it. If you can help me with what else we can do, it will help me a lot.



Regarding:
> * Ask questions for AM and PM in play mode

I didn't get you exactly what we need for this. Do we just need to add AM and PM in the bottom time when in play mode or do we need any changes in Clock also?  
Like for both 5:00:00 AM and 5:00:00 PM, we have to drag the hands at 5'o clock if I am correct.

llaske:
> I am searching for a better alternative to it. If you can help me with what else we can do, it will help me a lot.

Hmmm. Not sure. May be @prakashujjwal1010 could help here. He implemented the set time feature.

> Ask questions for AM and PM in play mode

Currently the computer ask value between 1H-12H.
In 24H mode it should ask value between 1H-23H.

prakashujjwal1010:
@saurabhhere for toggling AM/PM in set time mode, I would suggest suggest you to toggle it wrt the time ( `timeToBeDisplayed` ) that has been calculated already via the angles [see here](https://github.com/llaske/sugarizer/blob/841bc39abf0413f146f80ae76be46aa4fa61ab32/activities/Clock.activity/js/activity.js#L675). so use `timeToBeDisplayed` rather than `handAngles` directly e.g. toggle whenver it passes hour 12 rather than angle 0.

saurabhhere:
@prakashujjwal1010 If I use ```timeToBeDisplayed``` instead of  ```handAngles``` and rotate minute hand when ```timeToBeDisplayed['hours']==12``` it will toggle again and again till the hour is between 12 and 1.

Preview: 

ClockActivitySetTime

I have placed toggle function inside ```moveHandsInSetTime``` and called when ```isDrag``` is ```true``` :

Screenshot from 2021-01-30 22-21-43

Can you please help me with that?
Thanks

prakashujjwal1010:
@saurabhhere obviously this simple logic won't work you have to consider the "change" not only the current hour value. By "toggle whenever it passes 12" I meant:
- check if the hour hand moves from the top-left clock's half( 9 to 12) to another top-right clock's half(12 to 3).
- check if the hour hand moves from the top-right clock's half( 12 to 3) to another top-left clock's half(9 to 12).

so plan above logic by comparing the previous value and current value of hour :)
(Note: Also test for the situation when the drag is very fast)

saurabhhere:
@prakashujjwal1010  Thanks! It worked :)

saurabhhere:
Changes Made:
1. Toggle AM/PM in Set-Time Mode
2. Ask questions till 23H in Set-Time-Game Mode

Preview of Changes:

Clock

All changes mentioned in #921 are made. Please review If any change is required.

llaske:
That's the idea.
But:
- In 24h mode when rotating hands and hour pass 12h, the background should change
- In 12h mode the Clock should be like before: background always white and no display of AM/PM after digit.

saurabhhere:
Updated PR with new changes.

Preview:

Clock

llaske:
Nice job.
Still one issue: in play mode in 24h mode the background should change be adapted to the question asked.
In the following screen capture, the background should be white.

Capture d’écran du 2021-02-07 21-26-47

Plus finally I'm not a big fan of "AM" "PM" format for digits when in 24h mode. It will be more consistent to display it in 0-23 format like in play mode.

saurabhhere:
> in play mode in 24h mode the background should change be adapted to the question asked.
> In the following screen capture, the background should be white.

I didn't get it exactly.
In 24h mode, the background is white for AM hours (i.e. 0-11h) and black for PM hours (i.e. 12-23h). So according to this, at the above-shown screen capture (i.e. 14h) background should be black.
Please confirm.

llaske:
@saurabhhere oops you're right

saurabhhere:
Changes made :
* In 12h mode Clock will show 1-12h with white background and in 24h mode Clock will show 0-23h (black background during PM hours).
* During Set-Time-Game Mode, in 24h mode, the background will be white for 0-11h and black for 12-23h.
* During Set-Time-Mode, in 24h mode, format will be digit only (without AM/PM) and background will toggle between white and black for AM & PM hours respectively.

Preview of all changes:

ClockActivity


If all changes looks good to you, I will update the PR.

llaske:
Sounds good.

saurabhhere:
PR Updated. Please Review

llaske:
Good job @saurabhhere.
Thanks for your contribution.

PR #938, opened 2021-03-14
Updating markdown files

### LINKED ISSUE #937: The way to launch Sugarizer locally with Chrome has changed
A recent discussion on sugarlabs mailing list has rightly brought this to our attention that the path{as mentioned in markdown files} via which we open sugarizer in our browser is actually incorrect.

Even I faced this issue where, upon running the following command:

`chrome --allow-file-access-from-files  index.html `

A blank page opens up, 

Screenshot (593)

to tackle this I always opened the index.html file navigating into the directory where it resides. This is however a heuristic approach and I presume not the right way to access sugarizer in browser locally.

ricknjacky:
fixes #937

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,7 +62,16 @@ To run locally **Sugarizer Application into the Web Browser** (GNU Linux/Mac OS/
 
 For **Chrome**, close ALL running instances of Chrome and re-launch it using the command line:
 
- 	chrome --allow-file-access-from-files  index.html
+ 	chrome --allow-file-access-from-files file:\\\Directory:\sugarizer\index.html
+
+In windows, \ is the way path is written and / is the way for Linux/OSX backward and forward slashes differ here, hence path for Linux/MacOS will be:
+
+	chrome --allow-file-access-from-files file:///Directory:/sugarizer/index.html
+
+Directory in the path above is where you have stored/cloned your sugarizer repo into. Notice the path in the image below to get the gist of it.
+
+![](images/Chrome-Sugarizer.png)
+
 
 On Windows, you should launch:
 
--- a/README.md
+++ b/README.md
@@ -62,15 +62,13 @@ To run locally **Sugarizer Application into the Web Browser** (GNU Linux/Mac OS/
 
 For **Chrome**, close ALL running instances of Chrome and re-launch it using the command line:
 
- 	chrome --allow-file-access-from-files file:\\\Directory:\sugarizer\index.html
+ 	chrome --allow-file-access-from-files file:\\\<Directory>\sugarizer\index.html
 
-In windows, \ is the way path is written and / is the way for Linux/OSX backward and forward slashes differ here, hence path for Linux/MacOS will be:
+In the path above, `<Directory>` is where you have stored/cloned your sugarizer repo into.
 
-	chrome --allow-file-access-from-files file:///Directory:/sugarizer/index.html
+In Windows, \ is the way path is written and / is the way for Linux/MacOS backward and forward slashes differ here, hence path for Linux/MacOS will be:
 
-Directory in the path above is where you have stored/cloned your sugarizer repo into. Notice the path in the image below to get the gist of it.
-
-![](images/Chrome-Sugarizer.png)
+	chrome --allow-file-access-from-files file:///<Directory>/sugarizer/index.html
 
 
 On Windows, you should launch:
--- a/docs/tutorial/VanillaJS/step0.md
+++ b/docs/tutorial/VanillaJS/step0.md
@@ -26,7 +26,9 @@ We will use **Chrome** in this tutorial but any other browser will work as well.
 
 To run Sugarizer locally in Chrome, close any running instances of Chrome and re-launch it using the command line:
 
-    chrome --allow-file-access-from-files index.html
+    chrome --allow-file-access-from-files file:\\\Directory:\sugarizer\index.html
+
+Directory in the path above is where you have stored/cloned your sugarizer repo into. Notice the path in the image below to get the gist of it.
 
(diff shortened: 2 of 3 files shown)


ricknjacky:
New Behavior: 

Screenshot (594)

ricknjacky:
@llaske please have a look at this and share your thoughts.

llaske:
@ricknjacky I wonder if there is no a more simple way in Chrome CLI to open a local file using relative path.
Did you try other way to do that? Did you explore CLI options in Chrome? It work in previous version, may be there is another way to do the same.

ricknjacky:
> @ricknjacky I wonder if there is no a more simple way in Chrome CLI to open a local file using relative path.
> Did you try other way to do that? Did you explore CLI options in Chrome? It work in previous version, may be there is another way to do the same.

There is, we can use `http-server` 
We'd have to install this package accordingly.
`npm install http-server -g`

Now, in the sugarizer directory, to serve it on port 8000, we can type:

`http-server -p 8000`

Should we approach the problem by this method?

llaske:
No it's worse, it need mean installing an HTTP server.
The idea is to be able to start Sugarizer development just by copying files in a local directory then using a browser and a text editor.

ricknjacky:
> No it's worse, it mean installing an HTTP server.
> The idea is to be able to start Sugarizer development just by copying files in a local directory then using a browser and a text editor.

I think then, the way proposed above is the correct one.
I say this because, we don't know the directory user is cloning this repo into, hence mentioning the steps accordingly will be correct. What do you think? please share your suggestions.

README.md
```diff
@@ -62,7 +62,16 @@ To run locally **Sugarizer Application into the Web Browser** (GNU Linux/Mac OS/
 
 For **Chrome**, close ALL running instances of Chrome and re-launch it using the command line:
 
- 	chrome --allow-file-access-from-files  index.html
+ 	chrome --allow-file-access-from-files file:\\\Directory:\sugarizer\index.html
```
llaske, line 65: Will be better to write `file:\\\<Directory>\sugarizer\index.html`

README.md
```diff
@@ -62,7 +62,16 @@ To run locally **Sugarizer Application into the Web Browser** (GNU Linux/Mac OS/
...
+
+In windows, \ is the way path is written and / is the way for Linux/OSX backward and forward slashes differ here, hence path for Linux/MacOS will be:
+
+	chrome --allow-file-access-from-files file:///Directory:/sugarizer/index.html
+
+Directory in the path above is where you have stored/cloned your sugarizer repo into. Notice the path in the image below to get the gist of it.
```
llaske, line 71: Write `<Directory>` instead. This paragraph should be before the one mentioning difference between Windows and Linux/MacOS

README.md
```diff
@@ -62,7 +62,16 @@ To run locally **Sugarizer Application into the Web Browser** (GNU Linux/Mac OS/
 
 For **Chrome**, close ALL running instances of Chrome and re-launch it using the command line:
 
- 	chrome --allow-file-access-from-files  index.html
+ 	chrome --allow-file-access-from-files file:\\\Directory:\sugarizer\index.html
+
+In windows, \ is the way path is written and / is the way for Linux/OSX backward and forward slashes differ here, hence path for Linux/MacOS will be:
```
llaske, line 67: Write Windows with an uppercase.
Write MacOS instead of OSX.

README.md
```diff
@@ -62,7 +62,16 @@ To run locally **Sugarizer Application into the Web Browser** (GNU Linux/Mac OS/
...
+
+	chrome --allow-file-access-from-files file:///Directory:/sugarizer/index.html
```
llaske, line 69: Write `file:///<Directory>/sugarizer/index.html`.

README.md
```diff
@@ -62,7 +62,16 @@ To run locally **Sugarizer Application into the Web Browser** (GNU Linux/Mac OS/
...
+
+Directory in the path above is where you have stored/cloned your sugarizer repo into. Notice the path in the image below to get the gist of it.
+
+![](images/Chrome-Sugarizer.png)
```
llaske, line 73: The image is useless.

docs/tutorial/VanillaJS/step0.md
```diff
@@ -26,7 +26,9 @@ We will use **Chrome** in this tutorial but any other browser will work as well.
 
 To run Sugarizer locally in Chrome, close any running instances of Chrome and re-launch it using the command line:
 
-    chrome --allow-file-access-from-files index.html
+    chrome --allow-file-access-from-files file:\\\Directory:\sugarizer\index.html
```
llaske, line 29: See previous comments.

docs/tutorial/VueJS/step0.md
```diff
@@ -25,7 +25,9 @@ We will use **Chrome** in this tutorial but any other browser will work as well.
 
 To run Sugarizer locally in Chrome, close any running instances of Chrome and re-launch it using the command line:
 
-    chrome --allow-file-access-from-files index.html
+    chrome --allow-file-access-from-files file:\\\Directory:\sugarizer\index.html
```
llaske, line 28: See previous comments.

llaske (changes requested):
Okay guess that there is no CLI option in Chrome to avoid prefixing with `file:///<Directory>` ?
See my request changes below.

ricknjacky:
> Okay guess that there is no CLI option in Chrome to avoid prefixing with `file:///<Directory>` ?

No, I searched for one like that, couldn't find one. Please do tell me if there is one in future 

> See my request changes below.

Committing them now. Thanks for the feedback.

llaske:
Nice. Thanks.

PR #941, opened 2021-03-22
[Fixed #940] Dollar Street Activity not working on browser

### LINKED ISSUE #940: Dollar Street Activity no longer work
Dollar Street Activity is Showing an error when opened on the browser.

Used latest dev branch

Preview: 

Screenshot from 2021-03-22 00-02-40

sarthak-g:
Fixed #940 
dollarStreet


saurabhhere:
I have tried the same fix at the first shot, but there are few other issues we have to look at.

1. `See things` palette shows "No place match your filter" I don't think that may be the expected result. 

Screenshot from 2021-03-22 12-41-17


2. When we open any family, it is showing "No description." 

Screenshot from 2021-03-22 12-44-17


I am not sure what are the expected result for the above two now (After API change), but earlier, they both were working fine.

sarthak-g:
@saurabhhere ,
these issues are due to change in API.
Ex- there is no description property therefore it is showing "No description"

@llaske , Is there any documentation for these api's?I'm not able to find it.

saurabhhere:
> these issues are due to change in API.
> Ex- there is no description property therefore it is showing "No description"

Yes, I agree

llaske:
Thanks @sarthak-g , I've fixed the issue in https://github.com/llaske/sugarizer/commit/b1fa6e07bb0aab4c5db23ec265d344e0ea8ed0f7

PR #948, opened 2021-04-18
[Fixed #945] Position of scrollbar after activity is quit

### LINKED ISSUE #945: Default position of scrollbar (top)  after everytime an activity is quit.
Rather than displaying  the activity list from top (by default) after every time we quit an application , the activity list is to be displayed at the point till the place up to where it was already scrolled. 
I thought for this as when i was trying all activities one by one in the activity list , when i reached to activities present in the last of activity page each time when i quit an activity i was by default directed to top of activity list . So for going to the activity next to previous activity i would have had to scroll again to bottom.

sarthak-g:
@llaske , Fixed issue #945 
Tested on browser and android mobile.

https://user-images.githubusercontent.com/42293606/115149960-fbc3b880-a083-11eb-97a2-705ef5e08e29.mp4


js/homeview.js
```diff
@@ -456,6 +456,13 @@ enyo.kind({
 		this.currentView = newView;
 		stats.trace(constant.viewNames[oldView], 'change_view', constant.viewNames[newView]);
 
+		if (
+			(constant.viewNames[oldView] == 'list_view' && constant.viewNames[newView] == 'home_view') || 
+			(constant.viewNames[oldView] == 'list_view' && constant.viewNames[newView] == 'neighborhood_view')
+		) {
+			sessionStorage.setItem("scroll-value", scrollbar_position());
```
llaske, line 463: Not a big fan of using `sessionStorage`. A context storage class is available in settings class: https://github.com/llaske/sugarizer/blob/dev/lib/settings.js

js/listview.js
```diff
@@ -1,3 +1,5 @@
+var scrollbarTop = 0;
```
llaske, line 1: Global variables should be avoided.

js/listview.js
```diff
@@ -251,3 +288,16 @@ function sorted(activities) {
...
+
+function scrollbar_position() {
+	var diff = scrollbarTop - scrollbarMaxTop;
+	if (diff > 0) {
+		return (scrollbarTop - diff);
+	}
+	return scrollbarTop;
+}
+
+window.addEventListener("beforeunload", (e) => {
```
llaske, line 300: Not a good idea to use this type of global event. Will be better to save position before running activity [here](https://github.com/llaske/sugarizer/blob/dev/lib/settings.js#L413).

llaske (changes requested):
Thanks for this proposal. See my comments below.

sarthak-g:
@llaske , made suggested changes. Please review.

js/listview.js
```diff
@@ -43,6 +48,16 @@ enyo.kind({
 		}
 	},
 
+	rendered:function() {
+
+		this.inherited(arguments);
+		this.scrollbar_session_value = preferences.getOptions("scroll-value");
```
llaske, line 54: Options in the `preferences` object is reserved for settings options of the user (specifically options coming from the Privacy dialog, see [here](https://github.com/llaske/sugarizer/blob/dev/js/dialog.js#L1014) for an example of use).
So instead of using it I suggest to create a new property and new setter/getter in the `preferences` object.

js/listview.js
```diff
@@ -129,13 +144,42 @@ enyo.kind({
...
 	onscroll: function(inSender, inEvent) {
+
 		var scrollBounds = inEvent.scrollBounds;
+		if(scrollBounds != undefined) {
+			this.scrollbarTop = Math.ceil(scrollBounds.top);
+			this.scrollbarMaxTop = Math.ceil(scrollBounds.maxTop);
+			if(!this.move_scrollbar) {
+				var diff = this.scrollbarTop - this.scrollbarMaxTop;
+				if (diff > 0) {
+					preferences.setOptions("scroll-value", (this.scrollbarTop - diff));
```
llaske, line 155: See previous comment on options use.

js/listview.js
```diff
@@ -129,13 +144,42 @@ enyo.kind({
...
+				}
+				else {
+					preferences.setOptions("scroll-value", this.scrollbarTop);
```
llaske, line 158: See previous comment on options use.

llaske (changes requested):
Good progress. I suggest below a change regarding preferences usage.

sarthak-g:
@llaske made suggested changes

llaske:
Good job. 
It's a very cool enhancement.
Thanks.

PR #949, opened 2021-04-20
Remove name and XO icon from Memorize Activity when user left

### LINKED ISSUE #794: Name and XO visible even after removing in Memorize Activity
In memorize activity, more than two users are not allowed. Users are automatically removed but their name and XO icon are still present.
In the image below, "sarthakGupta" and "hellomnop" are automatically removed but their name and XO are still present.
NameAndXOMemorize

saurabhhere:
Fixed #794 

Automatically remove the name and xo icon when more than two users joined. 
Also, will remove the name and xo icon if one out of two users left and add it again when the second user rejoins.

Preview of Fix:

Memorize


saurabhhere:
@llaske Please review

llaske:
Good job. Thanks.

PR #950, opened 2021-04-20
Fixed incorrect error message in Calculate activity

### LINKED ISSUE #798: calculate activity outputs the incorrect error message if variable is changed in function
function f(y) { if (arguments.length != 1) { throw new SyntaxError("Wrong number of arguments in function f (" + arguments.length + " provided, 1 expected)"); } return y }

This is output when you change the x in f(x) to any other letter. the correct error message should be "invalid left hand side of assignment operator =".

saurabhhere:
Fixes #798

This error occurs as inside [`calcEqualClick`](https://github.com/llaske/sugarizer/blob/f489cd323b08a41830426336b7c8a94da191933b/activities/Calculate.activity/js/calculate-activity.js#L188) function only treats `f(x)` as a graph equation and all other as equations to solve. So when we enter any variable other than x it tries to solve it using the [`launchCalculation`](https://github.com/llaske/sugarizer/blob/f489cd323b08a41830426336b7c8a94da191933b/activities/Calculate.activity/js/calculate-activity.js#L162) function instead of the [`launchGraph`](https://github.com/llaske/sugarizer/blob/f489cd323b08a41830426336b7c8a94da191933b/activities/Calculate.activity/js/calculate-activity.js#L132) function and return error when unable to solve it.

In the fix, I have added a check for graph equations in launchCalculation function, if it is a graph equation it will return from the function along with the error in the display. 

Preview:

Screenshot from 2021-04-20 23-03-25

### DIFF
--- a/activities/Calculate.activity/js/calculate-activity.js
+++ b/activities/Calculate.activity/js/calculate-activity.js
@@ -164,8 +164,9 @@ function launchCalculation(calcInputValue, labelValue) {
     calculation: calcInputValue,
     graph: false
   };
-
-  if (calcInputValue[0]=='f' && calcInputValue[1]=='('){
+  
+  var reg = /f\(.*\)=[\w]+/;
+  if (reg.test(calcInputValue)){
     calculation.error = "Invalid left hand side of assignment operator =";
     CalculateApp.displayCalculation(calculation);
     return;
--- a/activities/Calculate.activity/js/calculate-activity.js
+++ b/activities/Calculate.activity/js/calculate-activity.js
@@ -165,6 +165,12 @@ function launchCalculation(calcInputValue, labelValue) {
     graph: false
   };
 
+  if (calcInputValue[0]=='f' && calcInputValue[1]=='('){
+    calculation.error = "Invalid left hand side of assignment operator =";
+    CalculateApp.displayCalculation(calculation);
+    return;
+  }
+
   if (labelValue) {
     calculation.label = labelValue;
   }
--- a/activities/Calculate.activity/js/calculate-activity.js
+++ b/activities/Calculate.activity/js/calculate-activity.js
@@ -154,7 +154,12 @@ function launchGraph(calcInputValue, labelValue) {
   CalculateApp.persistCalculations();
   CalculateApp.displayCalculation(calculation);
   if (!calculation.error) {
-    CalculateApp.graph(calculation.calculation);
+    try {
+      CalculateApp.graph(calculation.calculation);
+    } catch (error) {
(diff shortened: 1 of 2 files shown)


llaske:
Hmmm. Very specific fix. May be it could be possible to catch error in a more generic way?

saurabhhere:
@llaske We can use regex `/f\(.*\)=[\w]+/` to catch error in more generic way. It will look for `f(` and `)=` in the expression.
If this looks good to you, I will update the PR.

llaske:
Okay but my question was more generic. Is it the only error that could happen during expression evaluation?
I don't want to create a specific code for each possible error type.
Is it possible to do something like a more global `try/catch`?

saurabhhere:
I checked the functions for different errors. There is already a [`try/catch`](https://github.com/llaske/sugarizer/blob/1421602f66359b7f8f2448ff9b2a454b6f411bd8/activities/Calculate.activity/js/calculate-activity.js#L174-L180) present which is handling the maximum general errors. This error is specific and returns directly from `try` block with an error in the console.

Screenshot from 2021-05-03 15-38-13

It only detects `x` as an argument. So when there is a variable with `x`  like `f(x+y)` then it is handled by `try/catch` itself. This case is specific when no `x` variable is present inside `f()` then it takes the number of arguments as 0 inside `f()` and returns above mentioned error in the console.

llaske:
@saurabhhere I think it's possible to add a `try/catch` section around the `mustache.render` then raise a generic error if something goes wrong.

saurabhhere:
Yes, I agree we can add a `try/catch` around the `mustache.render`, but there are three `mustache.render` for which we have to use `try/catch` instead I was thinking to add `try/catch` block around `displayCalculation` so if anything goes wrong inside `displayCalculation` it will raise a generic error.

Preview of `try/catch` around `displayCalculation`:

Screenshot from 2021-05-05 18-35-54

What do you think about this?

llaske:
Why not? But in that case `try/catch` could be also be inside the `displayCalculation` method.
It will avoid to have some calls with `try/catch` and some calls without `try/catch`.
We should just ensure that in any case `displayCalculation` will display a result or an error message.

saurabhhere:
So should I add `try/catch` around both `displayCalculation` and `mustache.render` to ensure that every time it displays a result or an error message

llaske:
Hmmm. No, I guess inside the `displayCalculation` method, it should be enough.

saurabhhere:
@llaske Done

llaske:
Nice. Thanks.

PR #960, opened 2021-05-29
[WIP] Story activity

saurabhhere:
This is a new activity for Sugarizer.


llaske:
When I'm trying to test it, the browser is stuck and I've got a continuous list of error messages.
Any idea?

image

saurabhhere:
Is this error coming on every browser every time? Or on any specific browser (Like Safari)?

llaske:
I've tested only on Chrome.

saurabhhere:
> When I'm trying to test it, the browser is stuck and I've got a continuous list of error messages.

@llaske Thanks for the testing. Fixed.

llaske:
Nice @saurabhhere it works now. Thanks.
I've got two remarks:

- I think that next/previous icons are too small. It will be better to increase size. To easily be usable on touch screens, icons should have the size of a finger. See for example icons size when reading in eBook activity.

image

- It could be nice to navigate directly to an image by clicking on it from the 9 images mode.

saurabhhere:
@llaske Thanks for your feedback.
I have implemented both suggestions and you can test them now.

llaske:
@saurabhhere I think there is an issue when switching context. See video capture below.

Peek 21-07-2021 21-42

saurabhhere:
@llaske Thanks. Fixed.

llaske:
Nice.
I suggest few improvements:
- Will be nice to ensure that the cursor is always at the end of text when switching from an image to another or from a view to another
- Will be nice to disable Play/Speak button when there is nothing to Play/Speak so the user will have a visual indication that he doesn't need to click on it
- I don't see any way to restart or change the image set. It was possible on the original activity via a Restart button but I think we should be better to do a palette for this. May be it could be a way to decide the number of image at start: 2x2 or 3x3 (take inspiration from Memorize activity)

saurabhhere:
@llaske Thanks for your suggestions. I have implemented all these features along with Shared mode and you can test them now.

llaske:
The export feature is now like expected. Two issues however:

* In the generated ODT, there is questions marks instead of images for the grid view.

image

* During PDF generation it will be nice to hide items and show a spinner during the process instead of letting images change.

Peek 01-09-2021 22-35

There is also two issues regarding storage.

* When the user logged off, there is an error when reopening the instance:

Peek 01-09-2021 22-48

* When the browser window is closed (instead of stopped), there is an error when reopening the instance:

Peek 01-09-2021 22-51

saurabhhere:
> During PDF generation it will be nice to hide items and show a spinner during the process instead of letting images change.
What kind of spinner are you expecting here? We have to render images here because `html2canvas` is taking screenshots of them and adding them to pdf.

llaske:
Could you hide the HTML elements then add a spinner GIF like [this one](https://github.com/llaske/sugarizer/blob/dev/activities/Calligra.activity/images/spinner-light.gif) on the top.

llaske:
@saurabhhere Nice job on the PDF export :+1:

saurabhhere:
@llaske When we log off and then re-login again the `onJournalDataLoaded` function is returning `null` as data in Story Activity, it's working fine if we don't log off. Do you have any idea about this? Anything else we have to add to our activity to handle this. Maybe the synchronization with the server is not working fine.

llaske:
@saurabhhere you're right it come from a synchronization issue in Sugarizer core so the problem is not related to the activity. I'm working on it. Could you fix other issues and clean your code (commented code)?

saurabhhere:
@llaske Both issues which you mentioned related to storage are due to synchronization issue in Sugarizer core?
I have fixed the issue with export to ODT and added spinner in export to PDF and will update the clean code.

llaske:
@saurabhhere No, the error reopening instance when the window is closed (instead of activity stopped) should be related to the fact that you probably save the instance only when button stopped is clicked. A better practice is to save it when the instance is initialized (to be sure that it's not null) then to save it regularly, for example when changing view/picture.

saurabhhere:
@llaske Thanks.
I have fixed that issue which is coming while closing activity without the stop button and also updated the clean code.

llaske:
Thanks @saurabhhere, it works now.
I've got another issue. There is an error when trying to export sound:

image

Look like you've copy/paste a code that use an undeclared `that` variable.  Please double check that the same issue is not present elsewhere in your code.

saurabhhere:
@llaske My bad, Thanks.
I have updated the PR and checked if the same issue is present somewhere else also.

llaske:
Nice. Thanks @saurabhhere it works now.

Could you also ensure the activity is reopen on the same image displayed before leaving?
In the following sample the activity stop on the Books images and is reopen on the G letter.

Peek 05-10-2021 22-33

saurabhhere:
@llaske Earlier it was implemented to open the first image in grid mode always whenever we reopen the activity in grid mode.
Now, I have updated the implementation to open the same image displayed before leaving as suggested by you.

llaske:
Good job. Thanks @saurabhhere .

llaske:
@saurabhhere I've found the issue on Android - the problem will be the same on iOS.

Because the app size is limited on Google/Apple store the Abecedarium activity is built without the database (images and sounds). See the option [minsize](https://github.com/llaske/sugarizer-apkbuilder#generate-your-apk) in Sugarizer APK Builder. When this option is active, all images/sounds are removed from the package (see [here](https://github.com/llaske/sugarizer-apkbuilder/blob/master/src/make_android.sh#L134)).

The Abecedarium activity and the Journal chooser popup are both able to dynamically detect if the database is stored locally or if the database should be load from internet. It's done by trying to load the [_ping.png](https://github.com/llaske/sugarizer/blob/dev/activities/Abecedarium.activity/images/database/_ping.png) dummy image during init. If this image can't be loaded locally it mean that the database should be load from internet. See the source code [here](https://github.com/llaske/sugarizer/blob/dev/activities/Abecedarium.activity/network.js#L38) in Abecedarium and [here](https://github.com/llaske/sugarizer/blob/dev/activities/ActivityTemplate/VueJS/lib/sugar-web/graphics/journalchooser.js#L454) in Journal chooser popup.

Currently, the Story activity doesn't detect if the database is stored locally or not so all images loading failed when images are not stored locally.
Could you add the same detection process for Story?

llaske:
@saurabhhere I think I understand the issue. The problem comes from your `checkFileExists` function [here](https://github.com/llaske/sugarizer/pull/960/commits/f6c72cce7c6b2b99137aae837401b73b17caa061#diff-d25f0aa22b02a2881580f1a6251ceed827e0d07e22c6e3230dd2afeed2b5d66bR177).
Your implementation consider that `XmlHttpRequest` is run synchronously but:
- It's not the [default behavior](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/send)
- Synchronous `XmlHttpRequest` are [deprecated](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/Synchronous_and_Asynchronous_Requests#synchronous_request)
- Due to the need of [cordova-plugin-wkwebview-file-xhr](https://github.com/llaske/cordova-plugin-wkwebview-file-xhr) plugin, `XmlHttpRequests` in iOS are necessarily asynchronous.

So the current implementation will not work on iOS and has risks of failure in future browsers versions.

Could you refactor the implementation of the `checkFileExists` function? It will require to pass a parameter to the function for callback and handle asynchronous calls of this method in the activity.

llaske:
@saurabhhere thanks for your work. Good idea to rely on async and promises.

BTW it seems that it raise a new issue::when a new board is created it's added to the existing one instead of replacing it.
See below:

Peek 30-10-2021 18-52

saurabhhere:
@llaske, Thanks for pointing that.
Just forgot to clear the image array as per our new asynchronous function.

llaske:
Thanks @saurabhhere. Good job 👍 
Both issues are now fixed: the one on iOS and the one on all platforms.

BTW I find another issue related to generation this time. The generation no longer works as expected. Some images are not generated. Here's for PDF but it's the same issue for other formats and it's not depending of the platform.

saurabhhere:
@llaske It's working fine on my side in all formats. 

story-activity

Is this issue appear in your system after switching to asynchronous function?

saurabhhere:
No, I don't think recent changes should affect the export feature. 
Maybe you should try once by clearing the browser cache.

llaske:
I've reproduced the issue on my second PC too.
Just by generating several times PDF I get randomly a PDF with between 3 and 6 images inside.
@saurabhhere could you try the same on your side?

Peek 05-11-2021 22-12

PR #988, opened 2022-03-03
Improve responsivity and some styling change in Calculator 

### LINKED ISSUE #982: Improve calculator screen responsivity
Screenshot below :arrow_down: 
we just have to set the width of container under media query

WhatsApp Image 2022-02-28 at 3 10 45 AM

aman-netizen:
Fixes #982 and also improve some styling.

image
image
image

Please review my work.

### DIFF
--- a/activities/Calculate.activity/css/activity.css
+++ b/activities/Calculate.activity/css/activity.css
@@ -1,6 +1,6 @@
 #body {
 	overflow-x: hidden;
-    overflow-y: hidden;
+  /* overflow-y: hidden; */
 }
 .toolbar {
   height: auto;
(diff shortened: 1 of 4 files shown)


llaske:
https://github.com/llaske/sugarizer/issues/982 is already fixed.
If you want to propose an improvement please first open an issue an explain what sort of improvement you propose. So we could discuss about it before the PR.

llaske:
The dimension you mention is not realistic. I don't think it's interesting to add/update the media query for that.
Regarding resizing of result square I think it's better to keep always the same thing, it's easier to understand for children.

aman-netizen:
Okay sir .

Issue #913, opened 2020-12-26
No Tooltip for journal button on home page, also no tooltip on some toolbar buttons in journal page.


prabhsharan36:
Tooltip is very essential part in Sugarlabs because it is used mostly by childrens. So, It would be easy for them to read tooltip and know what the button will do.

prabhsharan36:
I would like to work on this issue.

llaske:
Tooltips are not so essential because it can't be displayed on touch screens.
Plus Help button give an explanation of each element of the UI.
BTW, sure you could propose a PR to fix tooltips where there missing.

hetanthakkar1:
can i take this issue?

llaske:
@hetanthakkar1 sure.

llaske:
Fixed in https://github.com/llaske/sugarizer/pull/995

PR #995, opened 2022-03-06
Fix for tooltips issue on the journal page.

AnandChourasia007:
Added tooltips in the toolbar buttons as well as footer buttons. Added the translations in all languages specified in locale.ini as well.
Issue #913


llaske:
Good job.
BTW just add translation in English. We avoid automatic translation in Sugarizer. It's better to let native speakers do their own translation.

AnandChourasia007:
@llaske , please check the latest commit.

llaske:
Nice. Thanks.

PR #1005, opened 2022-03-15
Update filter pallete in the paint activity

### LINKED ISSUE #1004: Paint activity, filter activity options not clear.
In the paint activity, under the filter activity there are two options. However, there is no tooltip explaining what they do nor the help button explains it during the tutorial. Suggesting a very minor enhancement, i.e. simple tooltips explaining what they do.

AnandChourasia007:
To include tooltips for the two options under the filter button.
Fixes #1004 
Before:
filter pallete before
After:
filter pallete after1
filter pallete after2


llaske:
Why not but tooltips should be localized.

AnandChourasia007:
Yeah, sure.
(BTW, I saw that in the translate.ini file, the translations for two languages are not yet done, instead plain english text is written there, the languages being german [de], and portuguese [pt].  Since its not expected to be done by contributors I dont think I should raise this as an issue. Will you please look into it?)

AnandChourasia007:
> Why not but tooltips should be localized.

What exactly do you mean by localized? I assumed its about storing the tooltip text in a variable and fetching it from translate.ini or locale.ini. But then I saw, in the index.html file none of the buttons' tooltips text are defined in this way. Please tell me what 'localizing' tooltips means. 😓

llaske:
> But then I saw, in the index.html file none of the buttons' tooltips text are defined in this way

All tooltips should be localized. It's a good opportunity to fix it.

AnandChourasia007:
@llaske , please review the new commits.

PR #1013, opened 2022-03-16
generate qr-code on enter

### LINKED ISSUE #1007: Get qr code with enter button
Currently, we need to click the ``generate-button`` to generate the QR code. We can also use enter to generate the QR code, it will be easier for the user.
image

dhwaj1902:
fix #1007


llaske:
Nice. Thanks.

PR #1017, opened 2022-03-17
Delete playstore.png

jigyasunigam:
This playstore.png file had no use in whole project, that's why I deleted it. And the same image is in the applestore.png file which has been used in the project's Readme section. If you still want that file I can rename it from "playstore.png" to "appstore.png" as it's a appstore's image not playstore's.


llaske:
You're right. This file could be deleted.
BTW PR should be send only on the dev branch. See [here](https://github.com/llaske/sugarizer/blob/dev/CONTRIBUTING.md) for more.

jigyasunigam:
Ok I am sorry  I didn't read the contribution guidelines carefully. I will make a pr in dev branch.

PR #1019, opened 2022-03-18
Delete playstore.png

jigyasunigam:
This playstore.png file had no use in whole project, that's why I deleted it. And the same image is in the applestore.png file which has been used in the project's Readme section.


llaske:
Nice thanks.

jigyasunigam:
@llaske 
Thank you for guiding me to make my first pull request.

Issue #1003, opened 2022-03-15
MediaViewer activity empty.


AnandChourasia007:
The mediaviewer activity shows an empty page with no option of adding media to be viewed. The readme file too is empty and doesn't shed light on what the activity actually intends to do and how to use it. Also as the help button is missing, there is no way of knowing about the activity and its functions.
![Empty mediaviewer activity](https://user-images.githubusercontent.com/89119829/158320920-8a87944f-e0f6-41c0-beea-91d41794be24.jpg)

llaske:
It's deliberate. MediaViewer activity is not intended to be launch alone. It's a viewer for existing media content in the Journal.

AnandChourasia007:
In that case, I think it would be helpful to users if this information is written somewhere like a help button or the future contributors if this is mentioned int the readme file for the activity.
What do you think, @llaske ?

llaske:
Hmmm. Why not. It could be an icon + a localized message in the background like the one when the Journal is empty.

<img width="1051" alt="image" src="https://user-images.githubusercontent.com/981841/158604534-33a89060-7aba-4617-8502-05bcda27a999.png">

AnandChourasia007:
Sure. I get it. I will do this and submit a PR soon.

AnandChourasia007:
Hey @llaske , is this good?
![MediaViewer empty message](https://user-images.githubusercontent.com/89119829/158663864-ccd741dc-b5ef-4dd7-b840-0ef83e393a64.jpg)

llaske:
That's the idea 
The message should be more simple: "Please open an existing media from the Journal".

llaske:
Fixed in https://github.com/llaske/sugarizer/pull/1081

PR #1022, opened 2022-03-20
Displaying message when MediaViewer activity is empty

AnandChourasia007:
Fix attempt at #1003 .
MediaViewer empty message

### DIFF
--- a/activities/MediaViewer.activity/index.html
+++ b/activities/MediaViewer.activity/index.html
@@ -8,6 +8,7 @@
           content="user-scalable=no, initial-scale=1, maximum-scale=1, minimum-scale=1, width=device-width, viewport-fit=cover"/>
     <link rel="stylesheet" media="not screen and (device-width: 1200px) and (device-height: 900px)"
           href="lib/sugar-web/graphics/css/sugar-96dpi.css">
+          <link rel="prefetch" type="application/l10n" href="locale.ini" />
     <link rel="stylesheet" media="screen and (device-width: 1200px) and (device-height: 900px)"
           href="lib/sugar-web/graphics/css/sugar-200dpi.css">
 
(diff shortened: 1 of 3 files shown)


llaske:
Sounds good but I'm surprise that the empty message is display every time.
I'm expect to have a condition to display it only if no content is read.

AnandChourasia007:
Its working fine on my local computer. Please let me know how do I reproduce the issue.

https://user-images.githubusercontent.com/89119829/159202337-14c27a84-5136-4c6a-aa6c-d991c8a36eb0.mp4

llaske:
Your latest commit no longer work (see error below).

image

The previous one worked but I'm expect to change visibility of the message according of the way the activity is launch.

AnandChourasia007:
Umm, my latest commit works fine on my pc. Both when opening sugarizer in browser as well as electron. @llaske

https://user-images.githubusercontent.com/89119829/159894296-202e8883-3805-4bf5-94b0-335057fa248c.mp4

As can be seen, the message is visible only when no content is read, and when content is viewed the message is not visible.

llaske:
I've reproduced the issue on a new environment.
Could you test again on your side?

AnandChourasia007:
Tested again on my side, seems to work as shown in the above video though. Please wait, I will once again thoroughly try to check where this inconsistency is coming from.
Just so that we are on the same page, the problem is that on your side the code I want to get merged does not work as shown in the video, right?

llaske:
@AnandChourasia007 Yes it doesn't work like in your video, I've got an error message in the Javascript console. So I never seen the message.
I suggest you to try to reapply your update on the latest dev branch.

llaske:
Fixed in https://github.com/llaske/sugarizer/pull/1081

PR #1030, opened 2022-03-22
hover effect for all planets

### LINKED ISSUE #1029: Hover effect to Planet Activity
We can provide the hover effect to all the planets so that it will be easier for the users to use this activity

https://user-images.githubusercontent.com/57860123/159445351-936390df-6fce-401b-a449-73a915a9eae3.mp4

dhwaj1902:
fix #1029

### DIFF
--- a/activities/Planets.activity/css/activity.css
+++ b/activities/Planets.activity/css/activity.css
@@ -283,3 +283,67 @@ canvas {
   top: 45%;
 
 }
+
+.Mercury:hover{
+  transform: scale(1.1);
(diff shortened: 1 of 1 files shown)


gitkaarma:
Rather than writing hover effect for each planet, you can add a class to each planet(Maybe already exist) and then write single CSS. This way unnecessary code can be removed. What do you think?

dhwaj1902:
yes @gitkaarma! I have pushed it.

llaske:
It's better like that.
Thanks.

PR #1037, opened 2022-03-23
Added functionality QR code with enter button #1007

jigyasunigam:
Now user can get QR code by hitting enter button inside the input box.


jigyasunigam:
fixes #1007

PR #1042, opened 2022-03-24
generate qr-code on enter

### LINKED ISSUE #1007: Get qr code with enter button

jigyasunigam:
fix #1007


llaske:
Already fixed in https://github.com/llaske/sugarizer/pull/1013

jigyasunigam:
> Already fixed in #1013

Yes I already know that. That's why I changed the code. And I think my code is more cleaner than previous code.
Please just take a look at files changed.

PR #1046, opened 2022-03-27
Update activity.js

jigyasunigam:
Minimised the code.


llaske:
We don't need to minimize the code. There is a deployment task dedicated to that: https://github.com/llaske/sugarizer/tree/dev#optimize-performance

jigyasunigam:
Ok I understood that.

Issue #1055, opened 2022-04-09
Overlapping of "activity-palette" wrapper container with "back-button" in some activities


S-kus:
In **Tangram.activity** and **Curriculum.activity** the `activity-palette`'s `container` is overlapping with `back-button`. As shown:

![curr1](https://user-images.githubusercontent.com/86892991/162558648-35519f8d-3b13-400a-b226-1bc1201c35c8.png)

![curr2](https://user-images.githubusercontent.com/86892991/162558652-426c7a27-df0d-423b-89b8-492c13bb8318.png)

![ten](https://user-images.githubusercontent.com/86892991/162558656-cd704cac-853b-4b58-aa4a-d94206f54cc0.png)

S-kus:
Hello @llaske, I fixed this, Could you please review its pull request.

llaske:
Good point.

llaske:
Fixed in https://github.com/llaske/sugarizer/pull/1056

PR #1056, opened 2022-04-09
Corrected overlapping of back-button in some activities

S-kus:
Fixes: #1055 

By correcting the `z-index` of the `back-button` is fixed. As shown:

fix1

fix2

### DIFF
--- a/activities/Curriculum.activity/css/Skills.css
+++ b/activities/Curriculum.activity/css/Skills.css
@@ -13,7 +13,6 @@
 	width: 50px;
 	height: 50px;
 	cursor: pointer;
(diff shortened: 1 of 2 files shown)


llaske:
It fix the issue but the back button no longer works in Tangram.

S-kus:
> It fix the issue but the back button no longer works in Tangram.

Thanks @llaske, Yes you are right. Only half of it is working because the `z-index` of `tangram-name-content` i.e. the **title** is higher than the `back-button`. 
I fixed it now. Could you please review it once again!

llaske:
Nice. Thanks.

Issue #1058, opened 2022-04-12
Overlapping of back button with the activity palette in Tangram Activity


nikkuAg:
<img width="233" alt="overalp" src="https://user-images.githubusercontent.com/77184239/163022328-0be628d3-2b29-4a6d-ba5b-860caf5fb9ec.png">
Back button in Tangram activity was coming above the activity palette.

nikkuAg:
Hey @llaske,

I have solved this issue in the PR #1059

llaske:
Duplicate https://github.com/llaske/sugarizer/issues/1055

PR #1059, opened 2022-04-12
Back button sent back to avoid overlapping

nikkuAg:
I have solved the issue about the overlapping of back button with activity palette
Issue: #1058 

s1
s2


llaske:
Duplicate https://github.com/llaske/sugarizer/issues/1055

Issue #1060, opened 2022-04-12
Settings of Falabracman activity was overlapping with activity palette


nikkuAg:
The settings are not visible when we open `activity-palette` in Falabracman activity.

<img width="497" alt="s3" src="https://user-images.githubusercontent.com/77184239/163026538-f2528149-170e-47d9-ae12-80f673e1544e.png">

nikkuAg:
Hey @llaske,

I have solved this issue 
PR: #1061 
Can you please review it

llaske:
You're right.

llaske:
Thanks but you're not supposed to send a PR to master. See [here](https://github.com/llaske/sugarizer/blob/dev/CONTRIBUTING.md).
I've merged it manually here: https://github.com/llaske/sugarizer/commit/15433b480f1afde09890f85f95035cb4b787f769

PR #1061, opened 2022-04-12
In Falabracman activity settings were overlapping with activity palette

nikkuAg:
I have removed the z-index from settings css so that the `activity-palette` is above the settings text and form.

s5
s4

Issue: #1060


llaske:
Thanks but you're not supposed to send a PR to master. See [here](https://github.com/llaske/sugarizer/blob/dev/CONTRIBUTING.md).
I've merged it manually here: https://github.com/llaske/sugarizer/commit/15433b480f1afde09890f85f95035cb4b787f769

Issue #1064, opened 2022-04-12
Physics element are coming above toolbar


nikkuAg:
If we drag the physics elemnts in Physics JS activity above the toolbar then we see that the elements overlap with toolbar.

<img width="556" alt="s8" src="https://user-images.githubusercontent.com/77184239/163033522-1395f05c-39a0-46a8-84fc-b6bb189a9e06.png">

nikkuAg:
Hey @llaske,

I have solved this issue
PR: #1065 
Please review it

llaske:
You're right, it will be better to not overlap the toolbar

llaske:
Fixed in https://github.com/llaske/sugarizer/pull/1170

PR #1068, opened 2022-04-18
Physics elements are properly aligned

nikkuAg:
I have changed the z-index of the canvas and aligned by setting the position from top.
s9

Issue: #1064


llaske:
With this fix, the activity no longer work when starting from a blank instance.

Issue #1048, opened 2022-03-29
Loading error on slow network


MananJain777:
It happened few times with me like whenever I play a certain game it just doesn't respond at all and to fix it I need to restart the website and re-login which isn't a good user experience.

https://user-images.githubusercontent.com/96513500/160713419-83426768-325d-47dd-b386-b4a01fdeaaf1.mp4

AnandChourasia007:
Hey @MananJain777 , it would be helpful if you could tell us how to reproduce this issue because Memorize is working fine on my computer.

Shreyas-SAS:
> whenever I play a certain game it just doesn't respond at all.

@MananJain777 Can you tell me once whether you are facing the problem with local file being run in a browser or using the try sugarizer web app on the website?

If its local file that is being run resulting an error then I had faced a similar scenario whenever I tried to launch Sugarizer in presence of another chrome tab/window (with or without the same account).

Please try to relaunch sugarizer after closing all chrome instances. If the issue persists please inform us. We will try best to resolve the issue.

MananJain2002:
Hi @AnandChourasia007 I don't know sometimes that error occurs sometimes it won't. It happened three times for me. I guess it might be cause of backend issues server might not be loading fast enough...

MananJain2002:
@Shreyas-SAS  I am using it on the website https://try.sugarizer.org/ as well as https://dev.sugarizer.org/ we can improve some backend code for it to render properly...

Shreyas-SAS:
> @Shreyas-SAS  I am using it on the website https://try.sugarizer.org/ as well as https://dev.sugarizer.org/ we can improve some backend code for it to render properly...

@MananJain2002 Change in Backend? I still don't get it what exactly seems to be the error caused. Can you please explain when the error is occurring, its not showing up at our ends.

It might also be that there may be a network issue or browser issue at the moment and no problem with the code. 

If you feel that there is a problem with the code, feel free to put up a PR. We will surely have a look and suggest if we feel any change is necessary.

MananJain2002:
@Shreyas-SAS it happened like three times for me with different wifi connections I don't think network is the problem and it occures sometimes and I thought it might be because the code is not rendering properly on the site might need to use some async/await and promises from javascript so that it can load properly surely I'll go through the code and try to figure out something if found will post a PR

llaske:
@MananJain2002 It look like a connection issue.
Please display the console/error window to see if there is any error.

[...]

PR #1070, opened 2022-04-19
Fixed Error

MananJain2002:
I have went through require.js documentations https://requirejs.org/ and have seen many articles about it Just by changing the loading time to fixes the bug

#1048

### DIFF
--- a/activities/Memorize.activity/lib/require.js
+++ b/activities/Memorize.activity/lib/require.js
@@ -191,7 +191,7 @@ var requirejs, require, define;
         var inCheckLoaded, Module, context, handlers,
             checkLoadedTimeoutId,
             config = {
-                waitSeconds: 7,
+                waitSeconds: 0,
                 baseUrl: './',
                 paths: {},
                 pkgs: {},


llaske:
It doesn't fix the issue on my side.
Could you explain how it's suppose to fix the timeout problem?

MananJain2002:
Hi @llaske sir, I referred this article https://www.moreonfew.com/load-timeout-for-modules-error-in-requirejs/ and then reequirejs documentations here https://requirejs.org/docs/api.html#config-map and https://requirejs.org/docs/errors.html even on stackoverflow page https://stackoverflow.com/questions/14279962/require-js-error-load-timeout-for-modules-backbone-jquerymobile after trying out different options it seams like by changing loading time of the page to zero solves it cause it render the page as fast as possible. I tried it on my local system as well it does solves in my system...

llaske:
You're right it solve the issue.
BTW I'm skeptical about this fix:
- In Slow 3G condition it means that the user has to wait more than 2mn to see the activity screen. It will be unusable 
- I'm not in favor of changing code in standard libraries

MananJain2002:
Hi @llaske sir actually I'm getting this error with 50mbps of internet speed as well as 80mbps I don't think it occurs only on slow 3g networks

https://user-images.githubusercontent.com/63538502/164953404-a168e40c-556d-4514-aea0-fea39d4a5ab3.mp4

by changing the code it actually fixes it...

llaske:
> by changing the code it actually fixes it...

It doesn't fix the issue.
The issue is that it take too long time to load the page.
Waiting few minutes is not acceptable.
As I said, it's better in this case to use the offline version of Sugarizer.

Issue #1047, opened 2022-03-28
Chess multiplayer bug when a third player join


AnandChourasia007:
Issue: If a chess match between two players is already going on, attempt to join by a third player is allowed. Any such attempts then actually BREAKS the game and the 1st and 2nd players lose the game status, returning to singleplayer screens.
Suggestion: Since, this activity is supposed to be used by kids, expecting them to 'not click when a chess game is full' is not a good idea. Instead, we can prevent clicking of the join game button after the second player has joined the game.

https://user-images.githubusercontent.com/89119829/160396030-f8d21940-923d-49c5-b786-1a7aa5fd08e3.mp4

llaske:
Good point.
We can't prevent joining the game but we can stop the activity of the third player in this situation.
I guess some other activities do the same. May be Memorize?

jcvails:
I was unsure of the player counts of other shared games so I only made the fix for chess, should be easy to add other player counts in the constant file though it may be worth adding player counts to the activities themselves for simplicity in creating new activities.

llaske:
@jcvails the fix should be done in Chess activity

UtkarshSiddhpura:
@llaske I've fixed the issue and tested it.
- 3rd user won't interrupt the game
- if any player leaves the opponent will be played by computer.

llaske:
Fixed in https://github.com/llaske/sugarizer/pull/1345

PR #1071, opened 2022-04-20
Chess multiplayer bug when a third player join #1047 - fix

jcvails:
- Chess now only allows 2 players to join a game
- Fixes game crashing bugs
- add player counts for other games in constant.js -> maxPlayerCount 
- #1047


llaske:
The fix should be done in Chess activity, not in Sugarizer core.

llaske:
The fix should be done in Chess activity, not in Sugarizer core.

Issue #1073, opened 2022-05-08
Toolbar improvements in LabyrinthJS.activity


S-kus:
1. Currently `sub-toolbar` is overlapping with `activity-palette`, as shown:

    ![laby1](https://user-images.githubusercontent.com/86892991/167293899-d334aff6-546c-462f-9f49-a48f9871557e.png)
    
    It should be like this:
    ![2laby](https://user-images.githubusercontent.com/86892991/167294174-ca2fb224-c09f-4d34-9f0d-17c12edbaec4.png)

2. The button title of Idea text and the background color is Stroke Color and Fill Color respectively, As shown:

    ![Screenshot from 2022-05-08 16-38-55](https://user-images.githubusercontent.com/86892991/167294141-132def15-4bd5-49d9-a5d5-4e633220c22a.png)

    It can be set to Text Color and Background Color for more specificity.

S-kus:
@llaske I raised the pull request for 1st issue. 
Could you please review it and also suggest is 2nd issue needed to be solved or not? If it's required, should I just update the title message in the locale.ini file, or is anything else required?
I will commit for this change in the same pull request.

llaske:
You're right for 1.
I don't think the text for 2 is confusing.

S-kus:
Thanks, @llaske,
I have fixed the 1st issue. Could you please review [this](https://github.com/llaske/sugarizer/pull/1074) pull request?

llaske:
Fixed in https://github.com/llaske/sugarizer/pull/1074

PR #1074, opened 2022-05-08
Improvement in LabyrinthJS.activity toolbar

S-kus:
Fixes: #1073 
- Fixed overlapping of `sub-toolbar` with `activity-palette` by setting the z-index of palette higher than sub-toolbar.


activities/LabyrinthJS.activity/lib/sugar-web/activity/activity.js
```diff
@@ -58,6 +58,8 @@ define(["webL10n",
         var activityPalette = new activitypalette.ActivityPalette(
             activityButton, datastoreObject);
 
+        document.getElementById("activity-palette").style.zIndex = "3";
```
llaske, line 61: You must not change code in sugar-web library. The code for this library must be the same for all activities.

S-kus: Okay sorry @llaske. 
I will try to solve this bug via styling changes in .css files.

S-kus:
@llaske I fixed it by adding style in activity.css, could you please review it again.

llaske:
Nice. Could remove change on activity.js from the PR?
Only the updated file should be in the PR.

S-kus:
> Nice. Could remove change on activity.js from the PR? Only the updated file should be in the PR.

Yes sorry for that.
I corrected it now.

llaske:
Nice. Thanks.

PR #1080, opened 2022-05-31
Added Translation for Planet Types in locale.ini(Fixes #1076)

### LINKED ISSUE #1076: Planet type is not localized
The planet type in information popup in Planets activity is not localized.
It's always in English (Terrestrial, Ice Giant, Gas Giant).

image

dante381:


dante381:
Have added the translations in locale.ini and added a .replace() function in activity.js

llaske:
Nice. Thanks.
Just one thing: do lot localize strings because it's better to let native people do the translation instead of using Google translate.

PR #1081, opened 2022-06-01
Message for MediaViewer activity empty

S-kus:
Fixes: #1003 

The message appearance is fixed without any error in the console, but the message is still not localized properly, it's kind of getting overwritten. As shown:

https://user-images.githubusercontent.com/86892991/171504463-51d1bb28-6b34-4d02-9861-3f1d6d84fac5.mp4

@llaske could you please guide me on the localization part.
Thank you!

### DIFF
--- a/activities/MediaViewer.activity/css/activity.css
+++ b/activities/MediaViewer.activity/css/activity.css
@@ -10,7 +10,3 @@ body {
     position:absolute;
     display: none;
 }
-
-#EmptyMessage {
-    margin: auto;
-}
--- a/activities/MediaViewer.activity/css/activity.css
+++ b/activities/MediaViewer.activity/css/activity.css
@@ -10,3 +10,7 @@ body {
     position:absolute;
     display: none;
 }
+
+#EmptyMessage {
+    margin: auto;
+}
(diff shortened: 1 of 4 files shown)


llaske:
This [line](https://github.com/llaske/sugarizer/blob/dev/activities/MediaViewer.activity/js/activity.js#L7) prevent your localization code to be executed. The test is the same that `if (!environment.objectId)` in the tutorial.

You should put your code before and it should work.

llaske:
Okay nice.

Just two things:
- For the message I suggest "You need to open an existing media from the Journal" (without dot at end)
- Use the English message for all translation. It's better to let translation to native speakers instead of using Google translate.

llaske:
Also, the message is not responsive. It should be in the middle of the screen whatever the size is.

Peek 03-06-2022 21-58

S-kus:
Sure @llaske, I have corrected these changes in the recent commit.

llaske:
Nice. Thanks.
On localization you should create strings for common languages (en, fr, es, de) but use the English value for all. So translators will know they have work to do.

S-kus:
> Nice. Thanks. On localization you should create strings for common languages (en, fr, es, de) but use the English value for all. So translators will know they have work to do.

It's merged now! Should I still commit with suggested changes?

llaske:
> This pull request merged now! Should I still commit with suggested changes?

No that's fine. I've done it.
Just remember that for next time.

S-kus:
Okay Thanks @llaske, I will be careful about this next time.

Issue #1075, opened 2022-05-12
Add Previous/Next button in DollarStreet activity to navigate in images


llaske:
It could be nice to add a Previous/Next button in DollarStreet activity on the detailed screen.
It will allow to navigate into the image gallery.

![Peek 12-05-2022 21-51](https://user-images.githubusercontent.com/981841/168157160-c3e83d9b-57df-4506-bbb9-abc3140a921d.gif)

codebloded:
hey, @llaske  nice it's good for user experience, I.m working on this !

S-kus:
@llaske Could you please review the attached pull request.

llaske:
Fixed in https://github.com/llaske/sugarizer/pull/1082

PR #1082, opened 2022-06-03
Added previous and next button to navigate things of family in DollarStreet Activity

S-kus:
Fixes: #1075 

I added the previous and next buttons to navigate through "things of families". Single images like family profile images will still be working as earlier i.e. when their popup opens, there is no previous and next button.


llaske:
Good job.

I wonder if it's a good idea to have an infinite loop: the one before the first is the last, the one after the last is the first.
It's seems too complex to understand for a child.

S-kus:
Hey @llaske, yeah you are right. Now, I removed the infinite loop and added a function so that, if the user clicked the previous button and it's already at index "0" then it will work as the back button and it will close the carousel. This insures users that there is no previous image (if it will not close then it seems something is not working properly as the image is not changing). Similarly for the next button.

llaske:
Hmmm. Closing the carousel it's not what I've got in mind. 
I think it's more consistent with other Sugarizer activities to just forbid to go before first and to go after last. 
So I suggest to hide the previous button on first image and to hide the next button on last image.

S-kus:
Hey @llaske, I made the above changes in the recent commit. Could you please review it again?
Now, it's working as suggested, i.e. the previous button hides if the current index of the image is 0 and similarly the next button for the last image. Also, I have to add `vm.destroy();` because if someone closed the modal and opens it again then the images do not update anymore if the previous or next button is clicked.

https://user-images.githubusercontent.com/86892991/172321419-09c110a4-b28a-4f37-a826-f5095b4cf50c.mp4

llaske:
It works, except when the user click directly on the first or on the last image. In this case buttons are not hidden and clicking on it raise an error in the console.

Peek 07-06-2022 21-41

Regarding implementation, it will be better to merge `showImage` and `showImageCarousel` function to avoid code duplication. You could for example differentiate the call to the method by an argument and you could hide buttons by default.

S-kus:
Hey @llaske, I updated the pull request with suggested reviews. Could you please check it again?

llaske:
Good job. Thanks.

PR #1083, opened 2022-06-14
[WIP] Replaced bootstrap-tour with introJs tour library in MindMath Activity

S-kus:
The following steps are followed:
- replaced bootstrap-tour .min.js, .min.css files and links with [introJs](https://introjs.com/docs/getting-started/install#:~:text=Intro.js%20consists%20of%20js%20and%20css%20files%20that%20need%20to%20be%20linked%20externally%20in%20the%20HTML%20page.%20The%20JS%20(intro.js)%20file%20should%20be%20linked%20before%20the%20closing%20%3C/body%3E%20tag.%20The%20CSS%20file%20(introjs.css)%20needs%20to%20be%20linked%20in%20the%20%3Chead%3E%20of%20the%20HTML%20page.) scripts.
- Replaced the keywords for description and placement of tour-box with intro and position to make steps according to introJs' requirements.
- Added function to start the introJs tour in `SugarTutorial.js`
- Added custom styling to make tour UI similar to old bootstrap-tour UI.


S-kus:
Hey @llaske, Could you please review this and let me know if any corrections are required!

llaske:
Nice start. It's pretty good.
Here are my remarks:

- The font size is not the same (13.3px instead of 14px)
- Buttons label are not localized

image

- Is it possible to center dialog in the middle of the targeted item?

image

- I'm not a big fan of the "Done" button when there is nothing left. Is it just possible to hide it or to set it disabled?

image

- Same question for the "Previous" button when on the first item

image

- The popup disappears when the user click outside the popup. Is it possible to disable this behavior like in bootstrap tour?
- The Pagination step is displayed every time, I think it should be displayed only when the pagination control is displayed

image

S-kus:
Hey @llaske, I corrected some of the suggested corrections, following is the description: 
- [x] The font size problem (13.3px instead of 14px)
- [x] Buttons label are not localized
- [ ] Is it possible to center dialog in the middle of the targeted item?
  I am not able to do this, as if I tried to get the current position manually and fix it by js but it's not updating for every step. Also, there is no feature I find in docs to update it. It's updating by intro.js, so I can't update styling at every step. Still looking to correct it.
- [x] Is it just possible to hide it or to set it disabled? (set to disabled)
- [x] Same question for the "Previous" button when on the first item
- [x] The popup disappears when the user click outside the popup. Is it possible to disable this behavior like in bootstrap tour?
- [ ] The Pagination step is displayed every time, I think it should be displayed only when the pagination control is displayed
  I didn't get the problem, could you please explain it again. Also, I have just updated the `placement` to `position` and `content` to `intro` of the `steps` array.

llaske:
- Okay for the font size but the line spacing is not sufficient, set the `line-height` value at the same value than in Bootstrap Tour
- Okay for buttons label localization
- Okay for centering the dialog, let's see if you find a simple way to fix it
- Okay for Previous and Next disable
- Okay for popup disappearing
- Regarding Pagination, the problem is global: it's what happens when the referenced element is not present in the page. It seems that in Bootstrap Tour the matching step is not displayed but in IntroJs the step is displayed anyway. Change for example `element: ".list-numbers"` into `element: ".list-numbersxxx"`, the matching popup is still displayed. Do the same thing in Bootstrap tour, the popup is no longer displayed. Could you try to analyze this behavior and try to find a fix? It could be a major compatibility issue.

S-kus:
Hi @llaske, 
In recent commit:
- Fixed the line height as earlier in bootstrap-tour.
- Fixed the pagination element problem by filtering the final steps array. Now, the array only contains the elements if it's actually there or if it is the default element (this step description will show in the middle of the screen).

Also, I still cannot find a good way for the middle position that works smoothly and uniformly for every step. 
On going through docs there is the [position option](https://introjs.com/docs/intro/attributes#data-position) of `bottom-middle-aligned` and in the min.js file there is `top-middle-aligned` as well, but both are not working uniformly for every step. 
For exa: setting the position of the step of element `.list-operators` and `.slots-area-footer` as `top-middle-aligned`, only work for `.list-operators`. i.e. they adjust themself accordingly even if there are space and no option for the right-middle and left-middle position.
Should I use `bottom-middle-aligned` and `top-middle-aligned` or not?

llaske:
Good job @S-kus .

Okay for line height.

Very good idea the idea to filter. But I think it could be improved. Why not testing the presence of the `element` property instead of forcing the developer to set the property to `.default` in each step?

Regarding position, do not change anything. Developers will choose their favorites position options. 

One more thing: you've mixed TAB and SPACE indentation in `activity.css`, it's important to respect the initial indentation choice of the file. Here it was SPACE.

llaske:
That's nice.

While reviewing your code I've seen you've removed the events `start`, `show` and `end` raised by the component: see [here](https://github.com/llaske/sugarizer/pull/1083/files#diff-43815ec1886368b47450dfe256cc92d8594fd7c643a362644fc9aeeef5fbc519L48-L55). Is there a way to reproduce these events in IntroJS?

S-kus:
There are some functions [here](https://introjs.com/docs/intro/api) in introJs, these functions are not exactly similar to functions of bootstrap, i.e. there are no methods like `end()` in introJs. Also, it's not working as a new component. Although we can execute some events like, `alert()` or `console.log()`.
I guess we are not even using these functions for any specific task still, If it will be needed in the future, anyone can follow the docs and use these functions to execute their custom functions.

llaske:
> I guess we are not even using these functions for any specific task still

Not in Mindmap activity but these Bootstrap Tour events are used elsewhere, for example in Sugarizer Core [here](https://github.com/llaske/sugarizer/blob/dev/lib/tutorial.js#L54-L88) to initialize/clean the tutorial.
So we should at least provide a migration process to explain how these calls could be reproduce with IntroJS.
Regarding `end` event it seems to be equivalent to [onexit](https://introjs.com/docs/intro/api#introjsonexitprovidedcallback).

S-kus:
> So we should at least provide a migration process to explain how these calls could be reproduce with IntroJS.

Yeah, that would be really helpful!



> Regarding `end` event it seems to be equivalent to [oncomplete](https://introjs.com/docs/intro/api#introjsoncompleteprovidedcallback).

Yes, similarly for `onshow()` ig we can use [onbeforechange](https://introjs.com/docs/intro/api#introjsonbeforechangeprovidedcallback)

S-kus:
Yeah, Thanks @llaske.
It's corrected now.

activities/MindMath.activity/js/activity.js
```diff
@@ -1127,188 +1127,166 @@ var app = new Vue({
...
-          element: "",
-          orphan: true,
-          placement: "bottom",
           title: this.l10n.stringTutoExplainTitle,
-          content: this.l10n.stringTutoExplainContent
+          intro: this.l10n.stringTutoExplainContent
         }, ];
       }
-
+      steps= steps.filter(function (obj) {
```
llaske, line 1287: Should be done in the SugarTutorial.show method.

activities/MindMath.activity/index.html
```diff
@@ -15,12 +15,12 @@
 	<link rel="stylesheet" href="css/game.css">
 	<link rel="stylesheet" href="css/result.css">
 	<link rel="stylesheet" href="css/leaderboard.css">
-	<link rel="stylesheet" href="css/bootstrap-tour-standalone.min.css">
+	<link rel="stylesheet" href="css/introjs.css">
 	<link rel="stylesheet" href="css/libnotify.css">
 	<script>if (typeof module === 'object') { window.module = module; module = undefined; }</script>
 	<script src="lib/vue.min.js"></script>
 	<script src="lib/jquery.min.js"></script>
```
llaske, line 22: `jquery.min.js` is no longer need. it was required by bootstrap tour.

activities/MindMath.activity/js/activity.js
```diff
@@ -1127,188 +1127,166 @@ var app = new Vue({
...
+        return ($(obj.element).length && document.querySelector(obj.element) && document.querySelector(obj.element).style.display != 'none') || !('element' in obj);
```
llaske, line 1288: $(obj.element) is a jQuery syntax. It can't be used now.

S-kus:
Hey @llaske, corrected the above changes, could you please review it once again!

PR #1092, opened 2022-08-24
[WIP] Add assignment Feature

codebloded:


codebloded:
- In these two commits

1. Restrict students from duplicating the assignment Entry. stop the uploading if it detects the `assignmentId` within `metadata`.
2. Add an assignment icon in the home view, by clicking on the icon journal view will open.

locale.ini
```diff
@@ -4080,4 +4080,4 @@ PrivacyRemove=Display delete account features
 PrivacyRemoveLocal=Delete local account
 PrivacyRemoveRemote=Delete server account
 TutoActivityMeasureactivity=Measure activity is an activity to learn the concepts of sound.
-TutoActivityStoryactivity=Story Activity uses images to prompt you to tell a story. You should try to tell a story that ties the images together into a comprehensive narrative.
+TutoActivityStoryactivity=Story Activity uses images to prompt you to tell a story. You should try to tell a story that ties the images together into a comprehensive narrative.
```
llaske, line 4083: This file should not be updated.

llaske:
In the Journal view the assignment icon should appear only if there is an assignment in the Journal.
The icon should be responsive (change position when window size change).

image

When clicking on the icon, it should display the Journal view with the assignment filter selected.

In the Journal View the Assignment filter should be near the favorite icon.
It should be responsive too.

image

When clicked, only Assignment items should be visible.

codebloded:
I know the filter option is left to build.
I'm figuring out how to implement it

codebloded:
@llaske Check the last commit!

llaske:
I can't test it with the Server side for now.
Here are my first remarks:

- Assignment icon on the home view should not be visible when there is no assignment in the Journal. 
- Assignment icon on the home view, when visible, should be colored (using buddy color). The number of assignments should be at the bottom right. It will be better to use a dark gray background for the circle. 
- When the assignment icon is clicked it should open the Journal view with the assignment filter active. 
- In the Journal view, the assignment icon in the toolbar should be to the right of the favorite icon.

llaske:
Two quick remarks before testing with Sugarizer Server:
- [ ] In the Journal view, the assignment icon in the toolbar should be to the right of the favorite icon.
- [ ] The Home view tutorial and the Journal view tutorial should be updated (new steps to explain new UI elements)

codebloded:
> Two quick remarks before testing with Sugarizer Server:
> - [ ] In the Journal view, the assignment icon in the toolbar should be to the right of the favorite icon.
> - [ ] The Home view tutorial and the Journal view tutorial should be updated (new steps to explain new UI elements)

Sure ! I'm doing this right now

codebloded:
@llaske I have added tutorials in both home and journal view and also change the due date format in assignment line.

llaske:
In the Journal view when the Favorite filter is activated and nothing match, there is a button to clear filter.
The button should appear also when the Assignment filter is activated.
See below.

Peek 07-09-2022 21-30

codebloded:
Okay , sure it'll be implemented in the next commit . I didn't even realise this bug 😅.!

codebloded:
@llaske i have check this above bug in the code of my branch it is working fine for me as you said above.
I have made journal view responsive ! Give me reviews for changes

llaske:
There is an issue with the return feature, when the an assignment was returned by the teacher, it's no longer possible to send it again. A message "The assignment has already been returned" appears.

Capture d’écran du 2022-11-11 11-11-42

Issue #1090, opened 2022-07-09
Replace Bootstrap tour by IntroJS in tutorial for all activities


llaske:
[Bootstrap tour](https://bootstraptour.com/) is the library use for writing activities tutorial. The library is no longer maintain and doesn't support recent version of Bootstrap. So we decided to replace it by another libraries named [IntroJS](https://introjs.com/).
IntroJS provide similar features and a lightweight UI.

![new_tutorial](https://user-images.githubusercontent.com/981841/178120591-b952de0e-a191-4c9a-a96c-819619437003.png)

Migration from Bootstrap tour to IntroJS require some change in code for each activity.
The work has been done for Sugarizer tutorial [here](https://github.com/llaske/sugarizer/commit/8e58af63412eca667f138b3f15864d884fba3af3) and for Mindmath activity [here](https://github.com/llaske/sugarizer/pull/1083?notification_referrer_id=NT_kwDOAA77UbEzODA3OTQ2NTc1Ojk4MTg0MQ#event-6964289669). The migration process is described [here](https://github.com/llaske/sugarizer/pull/1083#issuecomment-1167648209).

The objective of this issue is to migrate to IntroJS each activity:

- [x] Abacus
- [x] Abecedarium	
- [x] Blockrain
- [x] Calculate
- [x] Calligra
- [x] ChatPrototype
- [x] Chess
- [x] Clock Web
- [x] Color My World
- [x] Constellation
- [x] Curriculum
- [x] Dollar Street
- [x] Ebook Reader
- [x] Exerciser
- [x] Falabracman	
- [x] Flip
- [x] FoodChain	
- [x] FotoToon
- [x] Fraction
- [x] Game Of Life
- [x] Get Things
- [x] Grid Paint
- [x] Implode	
- [x] Labyrinth JS
- [x] Last One Loses
- [x] Maze Web
- [x] Measure	
- [x] MediaViewer
- [x] Memorize
- [x] MindMath
- [x] Moon
- [x] Paint
- [x] Physics JS
- [x] Planets	
- [x] Pomodoro
- [x] QRCode
- [x] Record
- [x] Reflection
- [x] Shared Notes
- [x] Speak
- [x] Sprint Math
- [x] Stopwatch
- [x] Story
- [x] TamTam Micro
- [x] Tangram	
- [x] Tank Operation
- [x] Video Viewer
- [x] Vote
- [x] Write
- [x] XO Editor
- [x] Xmas Lights

Please make a PR for each activity.

[...]

PR #1117, opened 2022-11-29
Replaced bootstrap-tour with introJs tour library in Last One Loses Activity

Ayeshashaik759:
I have migrated Last One Loses activity tutorial from bootstrap-tour to IntroJs


llaske:
Nice. Thanks.

PR #1118, opened 2022-12-05
Replaced bootstrap-tour with introJs tour library in QR code Activity

Ayeshashaik759:
I have migrated QR code activity tutorial from bootstrap-tour to IntroJs


llaske:
Good. Thanks.

PR #1123, opened 2022-12-25
Replaced bootstrap-tour with introJs tour library in Write Activity

Ayeshashaik759:
I have migrated Write activity tutorial from bootstrap-tour to IntroJs


llaske:
There is an issue: the tutorial popup is displayed for icons not visible (due to window size) in the toolbar.

image

Ayeshashaik759:
@llaske  Can you any suggestions to solve this issue

llaske:
The filter part in tutorial method should handle this: filter step pointing on hidden element.
Check if it works.

Ayeshashaik759:
The order of icons visible in your toolbar is different from order I am getting and I got the order of icons same as in the sugarizer.org website so I am little bit confused as I can't figure out the issue clearly. But I will take time and try to analyse and solve the issue.

PR #1126, opened 2022-12-30
Replaced bootstrap-tour with introJs tour library in flabracman activity

Ayeshashaik759:
I have migrated Flabracman activity tutorial from bootstrap-tour to IntroJs


llaske:
Good work. Thanks.

PR #1134, opened 2023-01-06
Replaced bootstrap-tour with introJs tour library in Xmas Lights Activity

Ayeshashaik759:
I have migrated Xmas Lights activity tutorial from bootstrap-tour to IntroJs


llaske:
Good @Ayeshashaik759 . Thanks.

PR #1145, opened 2023-01-19
Replaced Bootstrap tour by IntroJs in Chess activity

saakxshi:
I have migrated Chess activity tutorial from bootstrap-tour to IntroJs


llaske:
The board is not fully displayed and there is errors in the console.

image

Did you test it on your side?

PR #1099, opened 2022-09-01
Replaced Bootstrap-tour by IntroJs in Paint activity

disha1202:
#1090 
Replaced Bootstrap-tour by IntroJs in Paint activity

[screen-recorder-thu-sep-01-2022-14-54-08.webm](https://user-images.githubusercontent.com/58461801/187882830-e2c5ec65-797f-48cd-8ef1-6098350e3893.webm)

PR #1096, opened 2022-08-31
Replaced bootstrap-tour with introJs tour library in Story Activity

abhishek-sultaniya:
Fixes: #1090 
I have replaced bootstrap-tour with introJs tour library in Story Activity.All the suggestions,improvements and modifications are welcome.I followed all the steps as described in [here](https://github.com/llaske/sugarizer/pull/1083#issuecomment-1167648209).

PR #1149, opened 2023-01-22
Replaced Bootstrap tour by IntroJs in Chess activity

saakxshi:
I have migrated Chess activity tutorial from bootstrap-tour to IntroJs

[screen-capture.webm](https://user-images.githubusercontent.com/91880528/213904578-512cdcaa-8556-497f-a00f-14dfd7642e7e.webm)


llaske:
Thanks for your contribution but SPACE//TAB should not be change in original files.

saakxshi:
@llaske please check and let me know if any changes are to be made.

llaske:
Icons are no longer displayed in the tutorial.

image

See in other activities (for example GameOfLife) how to keep this feature.

saakxshi:
Sure i'll look into it.

Ayeshashaik759:
Hlo @saakxshi you can solve this issue by making the following changes in the tutorial.js file and you can even remove the showPieces (from line 66 to 90) method in tutorial.js after this.

Screenshot (118)

The output view.
Screenshot (119)

saakxshi:
@Ayeshashaik759 Thanks for your input.
@llaske Now the icons are visible, but there's one issue regarding the closebutton position in the pawn title. In order to solve this issue, I looked into PR #1099 and #1096, which suggested downgrading the introjs version, but even after doing that, there was no change, so can you suggest what is needed to be done?

image

Ayeshashaik759:
@saakxshi this is not an issue it is a normal case when the text length is more.

saakxshi:
> @saakxshi this is not an issue it is a normal case when the text length is more.

How it can be fixed?

Ayeshashaik759:
> > @saakxshi this is not an issue it is a normal case when the text length is more.
> 
> How it can be fixed?

No need to fix it actually 😅

llaske:
Nice. Thanks for your work @saakxshi

PR #1154, opened 2023-01-22
Replaced bootstrap-tour with introJs tour library in Speak Activity

dikwickley:
Replaced Bootstrap tour by IntroJs in Speak activity
issue #1090


llaske:
Good. Thanks @dikwickley

PR #1155, opened 2023-01-22
Replaced bootstrap-tour with introJs tour library in Record Activity

dikwickley:
Replaced Bootstrap tour by IntroJs in Speak activity
#1090


llaske:
Thanks for your contribution @dikwickley

PR #1162, opened 2023-01-24
Replaced Bootstrap tour by IntroJs in Food ChainActivity

dikwickley:
@llaske I have migrated the Game of Life activity tutorial from bootstrap-tour to IntroJs
Issue: #1090


Ayeshashaik759:
Hlo @dikwickley I just want to give some suggestion on your pr. did you got the output as excepted if you got the output as excepted its ok or else in tutorial.js file in the line  you have used "steps = tour.addSteps([ " you have deleted tour variable in the beginning I think so this may cause some error and have checked the tour output in learn and build and play mode.

dikwickley:
you are right @Ayeshashaik759 thanks for the heads up

PR #1166, opened 2023-01-25
Fixed Tutorial steps missing at first launch

### LINKED ISSUE #1089: Tutorial steps missing at first launch
At first launch, the starting tutorial has only 2 steps (instead of 9).

Peek 26-06-2022 11-12

Restarting the tutorial solve the issue.

UtkarshSiddhpura:
Fixes  #1089 issue
@llaske pls review it

### DIFF
--- a/js/firstscreen.js
+++ b/js/firstscreen.js
@@ -91,11 +91,6 @@ enyo.kind({
 		toolbar.style.backgroundColor = "white";
 		document.getElementById("canvas").style.overflowY = "hidden";
 		util.hideNativeToolbar();
-	},
-
-	// Render
-	rendered: function() {
-		this.inherited(arguments);
 
 		// At first launch, display tutorial
 		var that = this;
@@ -104,7 +99,6 @@ enyo.kind({
 				that.startTutorial();
 			}
 		}, constant.timerBeforeTutorial);
-
 	},
 
 	localize: function() {


llaske:
I don't understand your fix.
The issue is related to the home view, not to the first screen.
Have a look on the video included with https://github.com/llaske/sugarizer/issues/1089

UtkarshSiddhpura:
> I don't understand your fix. The issue is related to the home view, not to the first screen. Have a look on the video included with #1089

I'm still exploring codebase my fix might be wrong, but does it fix the issue pls check it on your side.

UtkarshSiddhpura:
@llaske Here is the explanation why my fix works :
- For some reason the firstscreen gets rendered twice on the screen
   **1st** time on FirstLaunch
   **2nd** just before the homeview 
- Hence, launching 2nd tutorial when the view is actually **homeview**  (since, we're calling tutorial once view gets rendered)
- now, all elements related to the **view** will get added in tutorial steps.
- then the tutorial steps will  **filter** out steps if the element is not present on screen.

What about the tutorial launch once homeview being created
- It'll get cancelled bcoz of this line : 
https://github.com/llaske/sugarizer/blob/f766b334a6f59af0bb04319b1171e4c89195e887/lib/tutorial.js#L508

Solution : Start the tutorial at the end while creating firstscreen (just like it is in homeview)

UtkarshSiddhpura:
Thanks 👍

PR #1170, opened 2023-01-28
Fixed physics element are coming above toolbar

### LINKED ISSUE #1064: Physics element are coming above toolbar

UtkarshSiddhpura:
Fixes #1064 issue


llaske:
Nice. Thanks.

PR #1172, opened 2023-01-29
Fixed creation of elements that're bigger than canvas

### LINKED ISSUE #1171: Creating Elements in PhysicsJs.activity that're Bigger than canvas
- Collisions are only detected for the **top edge** of view-port (i.e base of toolbar) for bodies/elements while **creating**.
- So elements bigger than canvas are easily possible to create from other edges
- e.g
phy

Solution: consider each edge of viewport and stop increase in size once edge is touched.

UtkarshSiddhpura:
Fixes  #1171 issue


llaske:
Good job. Thanks.

PR #1173, opened 2023-01-29
Fixed Bad handling of multitouch in PhysicsJs.activity

### LINKED ISSUE #849: Bad handling of multitouch in Physics activity
In Physics activity on touch devices, when creating a new item with one finger then using another finger, a new item is created but the first one is stick in the air at its original position.
The only action possible on the first one is now to remove it.

See in action below:

ezgif com-video-to-gif(2)

UtkarshSiddhpura:
**Fixes** #849 **Multi-touch** issue for both **Play** & **Pause** state.
**Solution**:  make previous body dynamic if not in paused state.

**more explanation**: `'interact:release':` event for bodies will not trigger if multi-touching. so Handle it in `'interact:poke':` event

### DIFF
--- a/activities/PhysicsJS.activity/js/activity.js
+++ b/activities/PhysicsJS.activity/js/activity.js
@@ -490,6 +490,10 @@ define(["sugar-web/activity/activity","tutorial","webL10n","sugar-web/env"], fun
 			var createdStart = null;
 			world.on({
 				'interact:poke': function( pos ){
+					// make previously created body dynamic
+					if (createdBody && physicsActive) {
+						createdBody.treatment = "dynamic";
+					}
 					// create body at a static place
 					if (currentType != -1 && pos.y > 55) {
 						createdBody = dropInBody(currentType, pos);


llaske:
Did you test it on a real device?
For example using https://github.com/llaske/sugarizer-apkbuilder

UtkarshSiddhpura:
Not really, what I did was opened the sugarizer in my phone desktop mode and tested it with multitouch.

Previously it was not working like you mentioned
And after fixing it worked well

UtkarshSiddhpura:
It should work if you still want me to test i can

llaske:
@UtkarshSiddhpura yes, please.

llaske:
It perfectly work on a real device. Good job @UtkarshSiddhpura

UtkarshSiddhpura:
> It perfectly work on a real device. Good job @UtkarshSiddhpura

Thank you for checking it by yourself, been busy lately...

PR #1179, opened 2023-01-31
Fixed Template container interfering the Calligra.activity

UtkarshSiddhpura:
Fixes #1174 issue

PR #1182, opened 2023-02-02
Fixed zoom container interfering the Calligra Activity

### LINKED ISSUE #1180: Zoom container interfering the Calligra Activity
### Issue
When the **zoom drop-down** is clicked, the interaction with the upper half of the canvas is **hindered** as it is on top of it.
This can lead to uneasiness as the drop-down menu can be easily left open until the desired size is achieved. And zooming in and out with the drop-down menu open further contributes to it.

image

### Solution
This can be avoided by making the display property inline so that it does not conflict. 

image

If the above seems fine then I'd like to implement it.

saakxshi:
Fixes #1180 issue.


llaske:
In fact this fix seems useless, https://github.com/llaske/sugarizer/pull/1179/files already solved the issue

saakxshi:
Yeah true.

Issue #1181, opened 2023-02-02
background colour of journal-container and body are same in somes activities


Ayeshashaik759:
when we click on insert option in write activity then a pop up opens in that when we go to search in abecedarium activity then as the background colour of the journal-container and body are same due to this there is no special highligth for background of scrollbar

![Screenshot (123)](https://user-images.githubusercontent.com/101845170/216251588-7abb67d1-12ef-4c78-959f-efc361e40f14.png)
  I just want to suggest that it will nice if the scroll bar have some background colour which can be implemented by changing the background colour of the journal-container. @llaske please share your view on this

somewhat like this 
![Screenshot (124)](https://user-images.githubusercontent.com/101845170/216252040-614f5a23-f65d-481c-b4a5-046bbff78b5b.png)

llaske:
Hmmm. Why not?
BTW I'm not sure that fixing this without side effect is easy.
This feature rely both on CSS and JS components.

Ayeshashaik759:
I tried to make the changes by changing the code in activities/Write.activity/css/quill/quill.core.css file. By manual testing (i.e by opening it in browser) I didn't observe any side effects. so @llaske please check the pr and share your view on it. If any side effects occur I will try resolve them.

llaske:
You must not change styles in external libraries because if the library is upgraded your change will be lost.
Plus the issue is the same in other activities (Write for example) so it's not related to Quill library.

Ayeshashaik759:
Ok i will not change the external libraries instead i will try to get the excepted changes using activity.css

Ayeshashaik759:
@llaske I think now it is working fine this time I made changes to the activity.css. And the remaining scrollbars are also working fine without any change. I reopened the pr please share your view on it.
![Screenshot (126)](https://user-images.githubusercontent.com/101845170/216813074-f0a6017f-a29d-44ca-8bd5-e89d40f76988.png)

llaske:
Same issue in Paint, Shared Notes and Fototoon.

Ayeshashaik759:
All the activities inherit same properties and opacity from the library lib/picoModel.js but still difference occurs. Another way is to solve this issue and make the scrollbar visible is to change the opacity for this activities but confused with were to make the changes as the changes are not applied by making changes to acticity.css. Ans as we cannot make changes to the libraries.

llaske:
Fixed in https://github.com/llaske/sugarizer/pull/1184

PR #1184, opened 2023-02-05
changing background colour of scroll bar in write activity

Ayeshashaik759:
I tried changing the background color of scroll bar in write activity by doing the following changes. Related to the issue #1181

### DIFF
--- a/activities/Fototoon.activity/css/activity.css
+++ b/activities/Fototoon.activity/css/activity.css
@@ -449,3 +449,7 @@ background-size: 100% 1.2em;
   }
   
 }
+
+.pico-overlay {
+  opacity: 0.66 !important;
+}
--- a/activities/Paint.activity/css/activity.css
+++ b/activities/Paint.activity/css/activity.css
@@ -370,3 +370,7 @@
 	background-color: #808080;
 	border: 2px solid #808080;
 }
+
+.pico-overlay {
+  opacity: 0.66 !important;
+}
--- a/activities/SharedNotes.activity/css/activity.css
+++ b/activities/SharedNotes.activity/css/activity.css
@@ -390,3 +390,7 @@ body {
     display: none;
   }
 }
+
+.pico-overlay {
+  opacity: 0.66 !important;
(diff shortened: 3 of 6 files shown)


llaske:
Smart fix :+1: 
BTW I wonder if the issue could not be fixed in another way.
The Memorize activity use the same Journal component but don't have the issue:

Capture d’écran du 2023-02-09 22-18-28

Ayeshashaik759:
Actually the background color of both are a little bit different.

Screenshot (5)

I inspected both the activities and observed that both have same properties but still difference occurs.

Screenshot (6)

I tried changing the opacity in write activity (by increasing the opacity from 0.5 to 0.66) then I observed that it is working fine and we can see the scroll bar.

Screenshot (16)

images compared after making changes to opacity in write activity.
Screenshot (11)

And we get this opacity from div with class "pico-overlay" which is an event present in the library lib/picoModal.js. I tried changing the opacity of class pico-overlay in activity.css by it is not working changes are not visible. And as we should not make changes to the libraries so @llaske  can you give suggest any idea to solve this.

Screenshot (17)

llaske:
Thanks for this analysis.
Should be tested but may be forcing the opacity by adding following lines in `activity.css` could do the job:

```
.pico-overlay {
  opacity: 0.66 !important;
}
```

Ayeshashaik759:
> Thanks for this analysis. Should be tested but may be forcing the opacity by adding following lines in `activity.css` could do the job:
> 
> ```
> .pico-overlay {
>   opacity: 0.66 !important;
> }
> ```

Yes it is working now got the output as expected. Shall I continue to do this changes for the remaining activities also In which this issue is occurring

llaske:
> Shall I continue to do this changes for the remaining activities also In which this issue is occurring

Yes. Please do the change in the same PR and ensure everything works like expected.

Ayeshashaik759:
Yah sure 😊

llaske:
Nice. Thanks.

Issue #1032, opened 2022-03-22
Fractions Activity: Add score counter


Shreyas-SAS:
Issue: 
There seems to be no score counter in Fractions activity but only a check list on the top-left corner.

Suggestion: 
Add a score display so that progress can be better mapped using high scores and motivate kids to score higher.

Shreyas-SAS:
@llaske please have a look once if this would be a good issue to resolve since we have a table already but not a score to map the number of correct answers.

llaske:
Why not? Let's suggest a nice place and design to add the current score.

Shreyas-SAS:
> Why not? Let's suggest a nice place and design to add the current score.

thank you for your confirmation. I will be working on the issue asap and put a PR soon.

Shreyas-SAS:
I am not able to work on the issue at the moment due to some academic commitments so if anyone is interested in working on the issue please feel free.

Shreyas-SAS:
> I am not able to work on the issue at the moment due to some academic commitments so if anyone is interested in working on the issue please feel free.

I will continue to work on the issue.

Shreyas-SAS:
> Hlo @Shreyas-SAS if you don't mind can I work on this issue. My suggestion is I have observed that in fraction activity at each turn we have 10 chances so the max high score can be 10 so is it okay to display just score each time. I tried to display the score in the following way(i.e on the top right side in case of mobile we can change its position to center) the score is set to 0/10 at the start of each turn. @llaske please share your view on this changes.
> 
> ![Screenshot (128)](https://user-images.githubusercontent.com/101845170/217252867-cf402635-451b-4966-9fcd-fbf642aa0c05.png)
> 
> ![Screenshot (127)](https://user-images.githubusercontent.com/101845170/217252700-04164d3f-1e8f-4639-927c-3896b5ca31f7.png)
> 
> in the below video you can see how the score is reset of each turn(i.e after 10 chances).
> 
> https://user-images.githubusercontent.com/101845170/217256818-8d70a8c0-6dd9-4bd3-bfb8-517685f31198.mp4
> 
> 
> 

Yeah sure. Please do. I lost the code that I implemented before creating a pr. Please continue. I do not wish to work on it.

Ayeshashaik759:
@llaske I tried adding score counter to the fraction activity. I tested it in the web browser I hope it is working fine. please check my pr and share your view on it.

llaske:
[...]

PR #1185, opened 2023-02-08
score feature added to fraction activity

Ayeshashaik759:
Here I have added the score feature to the fraction activity that is related to the issue #1032

### DIFF
--- a/activities/FractionBounce.activity/index.html
+++ b/activities/FractionBounce.activity/index.html
@@ -118,6 +118,8 @@
 						 v-bind:answer="answer" v-bind:correct-answers="correctAnswers" v-bind:mode="mode" />
 		</div>
 
+		<div id="Score">score:0/10</div>
+		
 		<sugar-activity ref="SugarActivity" v-on:initialized="initialized"></sugar-activity>
 		<sugar-journal ref="SugarJournal" v-on:journal-data-loaded="onJournalDataLoaded" v-on:journal-load-error="onJournalLoadError"></sugar-journal>
 		<sugar-localization ref="SugarL10n" v-on:localized="localized"></sugar-localization>
--- a/activities/FractionBounce.activity/index.html
+++ b/activities/FractionBounce.activity/index.html
@@ -118,7 +118,7 @@
 						 v-bind:answer="answer" v-bind:correct-answers="correctAnswers" v-bind:mode="mode" />
 		</div>
 
-		<div id="Score">score:0/10</div>
+		<div id="Score"></div>
 		
 		<sugar-activity ref="SugarActivity" v-on:initialized="initialized"></sugar-activity>
 		<sugar-journal ref="SugarJournal" v-on:journal-data-loaded="onJournalDataLoaded" v-on:journal-load-error="onJournalLoadError"></sugar-journal>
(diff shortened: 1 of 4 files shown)


activities/FractionBounce.activity/js/activity.js
```diff
@@ -6,6 +6,9 @@ requirejs.config({
 	}
 });
 
+var score = 0;
```
llaske, line 9: It's better to avoid global variables. You could use data variables instead.

activities/FractionBounce.activity/js/activity.js
```diff
@@ -295,11 +303,16 @@ let app = new Vue({
...
 				this.onTouchEnd();
 				let result = this.$refs.slopecanvas.checkAnswer();
+				count++;
 				if (result == this.answer) {
 					this.successSound.play();
+					score++;
 				} else {
 					this.failSound.play();
 				}
+				if (count < 30) {
```
llaske, line 313: Why 30? Not sure to understand.

activities/FractionBounce.activity/index.html
```diff
@@ -118,6 +118,8 @@
 						 v-bind:answer="answer" v-bind:correct-answers="correctAnswers" v-bind:mode="mode" />
 		</div>
 
+		<div id="Score">score:0/10</div>
```
llaske, line 121: The string should be localized.

activities/FractionBounce.activity/js/activity.js
```diff
@@ -295,11 +303,16 @@ let app = new Vue({
...
+					Score.innerHTML = "score: " + Math.round(score / 3) + "/10";
```
llaske, line 314: The string should be localized.

llaske (commented):
It's a good start.
BTW I think it's better to display the score only at the end of the game.
May be it could be center on the screen.

activities/FractionBounce.activity/index.html
Ayeshashaik759: Sure I will try to localize score. I have learned a little bit about it as part of learning how to create our own activity in vanillajs and vuejs.

activities/FractionBounce.activity/js/activity.js
Ayeshashaik759: Ok i will make the required changes

activities/FractionBounce.activity/js/activity.js
Ayeshashaik759: I tried printing count for each question in this method. But as this method is called trice for each question I got count 3 in console. So I have taken count<30 i.e till 10 questions  and when 10 questions are reached and then when we click on play button then as the new turn begins i changed score to 0/10

activities/FractionBounce.activity/js/activity.js
Ayeshashaik759: I will work on it for sure

Ayeshashaik759:
@llaske I tried to make the required changes. This the score is displayed after each turn please share your view on this
https://user-images.githubusercontent.com/101845170/218304739-756bbeb5-f37c-4362-b288-27f08861178b.mp4

llaske:
Good job. Thanks @Ayeshashaik759

PR #1194, opened 2023-02-25
Replace tutorial in SharedNotes activity

llaske:


llaske:
Fix https://github.com/llaske/sugarizer/issues/1090

PR #1199, opened 2023-02-26
Placeholder in search bar is now readable

abhishektiwari47:
Previously the the placeholder "Search in Neighborhood" was becoming unreadable in smaller screens.
Hence some changes were needed to fix this issues. Now the whole placeholder is visible without making any disturbance with rest of the UI

### DIFF
--- a/activities/QRCode.activity/css/activity.css
+++ b/activities/QRCode.activity/css/activity.css
@@ -1,5 +1,6 @@
 #main-toolbar #activity-button {
 	background-image: url(../activity/activity-icon.svg);
+	-webkit-appearance: none;
 }
 
 #main-toolbar #png-button {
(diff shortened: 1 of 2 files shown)


abhishektiwari47:
This might not be visible on a wider screen but you will find this problem in small screen. Just reduce the screen size to half.

llaske:
Do not send PR without an issue. We need to reproduce your issue before validating your PR.
Specifically it's not clear here what the issue is. What means unreadable ? Which screen size?

abhishektiwari47:
Thanks for this guidance. I will mention the issue soon.

abhishektiwari47:
Sir, can you please tell me the name of GSOC 23 mentors. Are you one of
them?

llaske:
Yes I am. The GSoC 2023 mentors list is available on https://github.com/sugarlabs/GSoC/blob/master/Ideas-2023.md

Issue #1201, opened 2023-02-27
Speak activity UI enhancement


Ydvaaman:
Hey @llaske . I have a suggestion for the Speak Activity. The ***separator*** separating the text box and the activity container has an incomplete length. So, what we can do is we can expand the length of the separator so it will look more symmetric than before.
For reference I have attached screenshots for the same, please look and review them and give me your thoughts. If you find it an issue then I would like you to assign this issue to me :)

## Before
![Screenshot (27)](https://user-images.githubusercontent.com/91108730/221446964-0acf9e29-813c-4673-a897-bf1821937133.png)<br>
## After
![Screenshot (28)](https://user-images.githubusercontent.com/91108730/221446755-d7413108-ca52-4d12-8b00-028599abd53a.png)

llaske:
Yes why not.

Ydvaaman:
hey @llaske  
I have raised a PR for the above  Speak activity UI enhancement please refer to my PR at https://github.com/llaske/sugarizer/pull/1204

llaske:
Fixed in https://github.com/llaske/sugarizer/pull/1210

sarvesh-kumar-code:
Hey @llaske As we are on speak activity you can see in the input box the arrow button is not aligned properly. 

#before#
![WhatsApp Image 2023-02-28 at 9 15 46 PM](https://user-images.githubusercontent.com/47060545/221904539-6e486d9a-9292-4fa4-81c0-e428425b8afc.jpeg)

#after#
![image](https://user-images.githubusercontent.com/47060545/221904936-6b17bd68-d001-4dde-952c-0c26f3a9626d.png)

Kindly share your thoughts on this.

rockharshitmaurya:
@llaske Actually this is a new issue: the alignment of the button on the text bar is not correct.
It is currently looking like this:
![WhatsApp Image 2023-03-01 at 2 47 35 AM](https://user-images.githubusercontent.com/46915044/221982901-85fd82cf-4fd2-4bca-8f78-e58803793a56.jpeg)

and it can be further improved like this :

![WhatsApp Image 2023-03-01 at 2 47 35 AM](https://user-images.githubusercontent.com/47060545/221904936-6b17bd68-d001-4dde-952c-0c26f3a9626d.png)

if you think this is an issue , we will be happy to fix this

llaske:
Hmmm. There is no issue on my side.

[...]

PR #1210, opened 2023-02-27
speak activity UI enhancement

Ydvaaman:
hey @llaske
I have updated the code for the Speak activity so that the separator separating the text box and the activity container spreads across the entire width of the screen. Refer to the  issue https://github.com/llaske/sugarizer/issues/1201

Screenshot (28)


llaske:
Okay. Thanks.

PR #1239, opened 2023-03-02
Replaced Bootstrap tour by IntroJs in Curriculam Activity

ABHINAV-JHA-27:
Migrated Curriculam Activity from Bootstrap Tour to IntroJs https://github.com/llaske/sugarizer/issues/1090


ABHINAV-JHA-27:
Hey @llaske , can you please have a look

llaske:
Two issues:

* [ ] The header has not the same height than in other activities.

image


* [ ] On the Trophy popup there is only one button.

image

ABHINAV-JHA-27:
Hey @llaske , I have solved the Padding issue But for the trophy issue, its happening because there is only 1 step for the introduction so it just shows next. should it have both??

UtkarshSiddhpura:
@llaske Yes, for one step it'll only show one button done/next
see: https://github.com/usablica/intro.js/issues/1101#issuecomment-746287995

llaske:
> But for the trophy issue, its happening because there is only 1 step for the introduction so it just shows next. should it have both??
It's better to show two buttons. @disha1202 found a work around [here](https://github.com/llaske/ExerciserReact/pull/143). May be you could do the same thing

llaske:
Fixed by https://github.com/llaske/sugarizer/pull/1286

Issue #1255, opened 2023-03-04
Markdown Activity: Mouse pointer UI issue on scrollbar


Ydvaaman:
## Hey @llaske
I have found the UI issue in the Markdown activity, when I try to point the mouse pointer on the scrollbar to scroll the slider the mouse pointer does not appear instead of the mouse pointer the text pointer appears on the scroll bar.
For reference I have attached a screenshot and video for the same, please look and review them and give me your thoughts.

![IMG20230304074408](https://user-images.githubusercontent.com/91108730/222870740-900ee82d-6f97-4573-8c5e-462ae50d2b3c.jpg)

https://user-images.githubusercontent.com/91108730/222870619-288ab0b4-9cf1-4da5-a10a-961ed8e10263.mp4

llaske:
That's right. Same in Write activity.

rockharshitmaurya:
> That's right. Same in Write activity.



> That's right. Same in Write activity.

hey @llaske i have fixed the issue in this PR(https://github.com/llaske/sugarizer/pull/1276) please review it

Ydvaaman:
Thank you @rockharshitmaurya 👍

Ydvaaman:
Hey @llaske
I have raised a PR for the above Markdown activity mouse cursor UI enhancement please refer to my PR at https://github.com/llaske/sugarizer/pull/1289

llaske:
Fixed in https://github.com/llaske/sugarizer/pull/1289

PR #1276, opened 2023-03-05
cursor issue fixed in Write.activity

rockharshitmaurya:
hey @llaske I have fixed the cursor issue as you suggested here (https://github.com/llaske/sugarizer/issues/1255#issuecomment-1454906559)

### DIFF
--- a/activities/Write.activity/css/activity.css
+++ b/activities/Write.activity/css/activity.css
@@ -1,5 +1,6 @@
 body {
   height: 100vh;
+  cursor: default;
 }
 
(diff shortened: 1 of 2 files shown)


llaske:
I'm not in favor of upgrading an external library.
If it's the only way to solve the issue, we could close the issue.

rockharshitmaurya:
hey, @llaske I have reverted those changes which I made in the external library and found another way to solve the issue, I have made a PR, please review it and give your thoughts

llaske:
Nice. Thanks.

Issue #1273, opened 2023-03-04
Ebook Activity : Want to map keys to the pagination


NischayGoyal1:
Hey @llaske , i want to add feature to Ebook activity , I want to map left and right arrow keyboard key to the paginations of ebooks, So that we can just change the pages by just pressing the keys
It will provide good user experience and feel

![image](https://user-images.githubusercontent.com/81116984/222932624-0f149241-1040-4227-803b-f2a6d0a35ecb.png)

llaske:
Sure. Why not.

NischayGoyal1:
Thanks ilaske , i am working on it

llaske:
Fixed in https://github.com/llaske/sugarizer/pull/1292

PR #1292, opened 2023-03-06
Mapped keyboard keys in EbookReader Activity

NischayGoyal1:
Hey @llaske , I have mapped the left and right keyboard keys to Pagination , Now we can change pages just by pressing keys
Please Review it
Issue : https://github.com/llaske/sugarizer/issues/1273

## Video

https://user-images.githubusercontent.com/81116984/223002252-aa62b63c-ca9a-4e97-b071-73cd85ff1dde.mp4


llaske:
Not a good idea to implement it like that. The listener is added to the whole document so there is an error when back to the library and pressing keys.

image

Plus the mounted method is call each time the ebook is displayed so the listener is added multiple times.

llaske:
It's not good, add a trace in your code, you will see that the listener is added each time a book is open.

NischayGoyal1:
#### I have implemented a fix by adding a "beforedestroy" hook. Specifically, I have added a removeEventListener function to the "beforedestroy" hook, which removes the event listeners whenever the Ebook component is destroyed. This ensures that the event listeners are only added once when the Ebook component is mounted and are removed when the component is destroyed, preventing the duplication of event listeners.
#### the reason why I added the event listeners in a separate JavaScript file was because the Ebook component was being mounted separately in a different file. As a result, there was no other way to add the event listeners without modifying that file.

I have made the required changes please review it

activities/EbookReader.activity/js/reader.js
```diff
@@ -11,6 +11,20 @@ var EbookReader = {
...
+			vm.previousPage();
+		  } else if (event.keyCode == 39) {
+			vm.nextPage();
+		  }
+		});
+	  },
+	  beforeDestroy: function() {
+		// Remove the event listener
+		document.removeEventListener("keydown", this.handleKeyDown);
+	  },
```
llaske, line 27: Look like indentation is wrong here.

NischayGoyal1: Fixed it

llaske:
Nice. Thanks.

PR #1300, opened 2023-03-07
stopwatchActivity: Activity dropdown UI Fix

VedantSharma11:
## SNIPPETS
---
### After:
Screenshot 2023-03-07 193854


llaske:
The `sugar-96dpi.css` file come from the Sugar-Web library and is share with all activities (see [here](https://github.com/llaske/sugarizer/blob/dev/docs/architecture.md)). It should not be updated in a single activity.

VedantSharma11:
okay @llaske i'll try some other way to fix it

llaske:
Nice. Thanks.

VedantSharma11:
Thanks @llaske

PR #1308, opened 2023-03-07
Refactor: Remove Unused Code for Improved Efficiency and Maintainability

rockharshitmaurya:
The CSS of the Flip and Food Chain activities, which were part of an old tutorial tour, has been refactored. This was done because we no longer require them since we are now using Intro JS for our tutorials.

Unused code can **slow down the performance**, and make it difficult to understand and **maintain the code**, Therefore, it's important to remove any unnecessary code to ensure that the software remains efficient and easy to maintain over time.


llaske:
Good point.

PR #1310, opened 2023-03-07
Curriculam Activity:Removal of Unused Bootstrap File from Codebase after Transition to Introjs Tour

VedantSharma11:
### As Curriculam Activity has been ported to Introjs Tour , so there was no use of the "bootstrap.min.css" file which was used in Bootstrap tour. So I have deleted it.


NischayGoyal1:
Hey @VedantSharma11 , You can also delete **jquery.min.js** , As it also of no use due to intro js

llaske:
You're right. Thanks.

PR #1312, opened 2023-03-07
Remove unused jQuery library from the project to reduce file size

rockharshitmaurya:
Hey @llaske, I Have Removed the unused jQuery library from some of the activities to reduce file size , I observed that in some of the activities of this project, there is a jQuery library located in the "lib" directory that remains unused **as it is not linked anywhere in the code.** This causes unnecessary weight in the codebase and removing it can help reduce the file size. 

Refactored Activities :
1. Curriculum.activity
2. Paint.activity
3. Vote.activity


llaske:
I guess you meant "Planet", not "Paint".
You're right. Thanks.

PR #1345, opened 2023-03-24
fixed bug on 3rd player join in chess.activity

### LINKED ISSUE #1047: Chess multiplayer bug when a third player join

UtkarshSiddhpura:
Fixes #1047

### DIFF
--- a/activities/Chess.activity/js/activity.js
+++ b/activities/Chess.activity/js/activity.js
@@ -407,7 +407,7 @@ var app = new Vue({
 
         }
       } else {
-        if (vm.opponentuser != null) {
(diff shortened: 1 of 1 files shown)


llaske:
Good job. Thanks.

PR #1357, opened 2023-05-17
📊 Implement new Sugarizer Chart activity

UtkarshSiddhpura:
### Feature To be Implemented:
-  [x] Activity UI
-  [x] Ability to input data to make different charts.
-  [x] Add/Remove data from chart.
-  [x] Swap Chart Fields.
-  [x] Config Palette to modify chart colors and axis labels.
-  [x] Font Palette to change font-family & size of labels & title.
-  [x] Horizontal , line & Pie charts.
-  [x] Full-screen Mode.
-  [x] Add activity Tutorial.
-  [x] Save in Journal.
-  [x] update activity description in locale.ini !
-  [x] Localization.
-  [x] Export Chart as image and CSV.
-  [x] Read data from Measure & Stopwatch activity !
-  [x] Read CSV file to make a Chart.(Support CSV in Journal & CSV Viewer)
-  [x] Share the activity !

#### Please review the completed features marked with checkmarks and provide your suggestions and feedback. Thank you!
---


llaske:
Very promising. Good job! :clap: 

My first remarks:

* The pie chart could be improved:
 * [ ] Axis and grid should not appear in this mode
 * [ ] Values should be in percent in this mode
 * [ ] Labels should be in around the pie
 * [ ] There must be a gradient of colors

image

See https://help.sugarlabs.org/chart.html for example.

* I'm not a big fan of the red bar on selected line. It seems that only the value column is selected. Could you suggest something else?

image

UtkarshSiddhpura:
> Very promising. Good job! clap
> 
> My first remarks:
> 
>     * The pie chart could be improved:
> 
>     * [ ]  Axis and grid should not appear in this mode
> 
>     * [ ]  Values should be in percent in this mode
> 
>     * [ ]  Labels should be in around the pie
> 
>     * [ ]  There must be a gradient of colors
> 
> 
> image
> 
> See https://help.sugarlabs.org/chart.html for example.
> 

Indeed it could be improved but as you can see in the features list the task about showing different chart type is WIP.
So let me implement it fully first :sweat_smile:

UtkarshSiddhpura:
>     * I'm not a big fan of the red bar on selected line. It seems that only the value column is selected. Could you suggest something else?
> 
> 
> image

Sure I'll suggest some better ones.

llaske:
> Sure I'll suggest some better. how is this?

Yes that's better

UtkarshSiddhpura:
>    * [ ]  Axis and grid should not appear in this mode 
>   * [ ]  Values should be in percent in this mode
>   *  [ ]  Labels should be in around the pie
>    * [ ]  There must be a gradient of colors
>    * I'm not a big fan of the red bar on selected line. It seems that only the value column is selected. Could you suggest something else?* 

I have taken your remarks and made the necessary changes. Please review the updated work and provide your feedback.

llaske:
Very good.

Regarding the gradient it seems a little light depending on the original color.

image

Did you retrieve the formula from the original activity?
Could you improve it?

activities/Chart.activity/NOTICE
```diff
@@ -0,0 +1,13 @@
+Copyright [yyyy] [name of copyright owner]
```
llaske, line 1: To be compliant with the license, please write your name and the year in this file like here https://github.com/llaske/sugarizer/blob/dev/activities/MindMath.activity/NOTICE#L1

UtkarshSiddhpura:
> Did you retrieve the formula from the original activity? Could you improve it?

No i didn't retrieve it from original activity, I'll improve it so it matches with the original activity.

llaske:
It's a good idea to use the activity title as chart title but it change when the view is change.

aqEpLUueF2

llaske:
I think it's better to not use quotes around values in the generated CSV file.

image

As mentioned [here](https://en.wikipedia.org/wiki/Comma-separated_values#Basic_rules) quotes can be ignore except in some specifics cases - coma in value for example.
On my side I'm more familiar with unquoted CSV files.

llaske:
I've seen three improvements: 
* [ ] In the toolbar the Import from Journal icon could be removed. It will be nice to add a vertical before the export palette.

image


* [ ] When opening and closing a CSV file, the timestamp of the CSV file seems to be more recent than the one of Chart activity. We could expect the opposite.

image

* [ ] When downloaded to the device the file has a .bin extension instead of a .csv

image

llaske:
Very cool :+1: 

Few remarks:

* [ ] Will be more consistent if the orange box will surround the whole column - like it surround the whole line in move lines feature.

Capture d’écran du 2023-06-18 22-27-16


* [ ] I understand the idea of the red background on title (when column values can't be converted to numeric) but it's not visible with the orange border. May be when the previous issue will be fixed, it will be okay.

16tk637M2t

* [ ] In sharing mode, I suggest to not share the fullscreen click. Like Tutorial and Stop, it's an user specific feature.

y7uRTPx7sG

* [ ] When opening a CSV file then changing its name, both the title of the Chart Activity and the title of the CSV is changed. The idea is to keep the same name for the CSV file.

S207XRXIVm

llaske:
Nice.

One more thing about sharing: the updated title is lost in sharing mode.
The title is changed but when the activity is reopened, the "old local" title is always used.
See example from the shared/sharing user side:

6JvL2WLu0i

n5s2eI1G5H

llaske:
Good.
It will be nice to give more explanation in tutorial  in CSV mode: only two first columns are keep, the way to move columns and the need to have a value in second column.

Issue #1303, opened 2023-03-07
Resizing textarea in activity palette looks ugly


kartikaysaxena:
For elaboration please have a look below

See how textarea on resizing falls out of div and also at the bottom it makes the other buttons go off the screen
<img width="1440" alt="Screenshot 2023-03-07 at 9 07 41 PM" src="https://user-images.githubusercontent.com/120778728/223472480-abdb44cb-0304-4302-b042-c5abe9c875e2.png">

My fix(How it should behave)
It should resize only in vertical direction like this
<img width="1440" alt="Screenshot 2023-03-07 at 9 08 19 PM" src="https://user-images.githubusercontent.com/120778728/223472811-06f80ba9-5f1f-4435-ac95-3ac8f2942f3e.png">

Shruti-codes236:
Can I work on this issue if this isn't resolved yet?

llaske:
The issue is the same for all activities.

llaske:
Fixed in https://github.com/llaske/sugarizer/pull/1522

PR #1361, opened 2023-05-25
Fixes the Text Area Resize 

Vishalk91-4:
Fixes issue #1303 

Hi @llaske
All the Activities section which had the text description box have been resized vertically, basically few codes have been implemented in all activity sections individually. 
I have also made the boxes resized up to a certain height so that it does not overflow.

After Modification

https://github.com/llaske/sugarizer/assets/116670999/3087060f-ea3a-4556-8574-d9ee1e6232d4


Vishalk91-4:
@llaske could you please review it.

llaske:
Changing `textarea` style globally is risky: it will apply to the object in all activity so what if there is another `textarea` inside. It will be better to apply change only to this one.

No need to do change in `sugar-200dpi.css` and `sugar.less`.

Vishalk91-4:
@llaske 
Do you need something like this, 
I'll add like description expand with it
textarea {
  resize: vertical;
  max-height: 70vh;
}
Should the things be added in css files outside or through every activity section

Vishalk91-4:
@llaske , the fix I found was to add the textarea in either the activity.css of each activity or in activity.pallete.html of each activity, will that work, just one line in each is to be added

llaske:
@Vishalk91-4 I think to target more precisely the `textarea` by prefixing hit by the full css path to it.
So even if there is another `textarea` in the activity, it will not match the css path.

PR #1369, opened 2023-06-17
Import device files

UtkarshSiddhpura:
re-using web FileReader api code to enable import/Read file from device File system on mobile devices. Please check on your side and give feedback.


lib/util.js
```diff
@@ -589,7 +589,7 @@ define(["l10n","sugar-web/datastore","FileSaver"], function (l10n, datastore, Fi
 	// Load a file
 	util.loadFile = function(file, callback) {
 		// Use FileReaer object in web
-		if (util.getClientType() == constant.webAppType || constant.noServerMode || (util.getClientType() == constant.appType && !util.platform.android && !util.platform.ios && !util.platform.electron)) {
```
llaske, line 592: You should not remove the `constant.noServerMode` test since it's not related to Android or iOS

llaske:
Nice. Thanks.

PR #1374, opened 2023-06-29
Add csv export for Stopwatch.activity.

UtkarshSiddhpura:


llaske:
I suggest to remove empty lines. So to remove counter without mark.
I suggest also to move the counter column at end and put in first column the mark number: 1 for the first mark, 2 for the second mark, ... The mark number value could be repeated if it comes from different counters.

image

UtkarshSiddhpura:
> The mark number value could be repeated if it comes from different counters.

like if mark is same for another counter then it should have same Mark number?

llaske:
Will be nice to update the tutorial to mention the new export button and explain how it works: export marks for each counter, etc...

Issue #1378, opened 2023-07-09
Replace WebL10n to i18next in all activities


llaske:
![Capture d’écran du 2023-07-09 18-18-55](https://github.com/llaske/sugarizer/assets/981841/08690e61-0dc4-4ad3-8d4e-232df8e853d4)

Currently localization in Sugarizer activities relies on [WebL10n localizaton framework](https://github.com/fabi1cazenave/webL10n).

This framework is outdated and no longer actively maintained. So in order to ensure continued support and modern localization capabilities, we decided  to migrate to the [i18next internationalization framework](https://www.i18next.com/).

This change require an update of all Sugarizer activities. The migration process, which depends on whether the activity is written in Vanilla JS or Vue.js, is described [here](https://github.com/llaske/sugarizer/blob/dev/docs/webL10n2i18n.md). Two activities have already been migrated [QRCode](https://github.com/llaske/sugarizer/pull/1371) and [Measure](https://github.com/llaske/sugarizer/pull/1371).

The objective of this issue is to migrate to i18next each activity:
* [x] Abacus
* [x] Abecedarium
* [x] Blockrain
* [x] Calculate
* [x] Calligra
* [x] Chart
* [x] ChatPrototype
* [x] Chess
* [x] Clock
* [x] ColorMyWorld
* [x] Constellation
* [x] Curriculum
* [x] DollarStreet
* [x] EbookReader
* [x] Falabracman
* [x] Flip
* [x] FoodChain
* [x] Fototoon
* [x] FractionBounce
* [x] GameOfLife
* [x] GetThingsDone
* [x] Gridpaint
* [x] Implode
* [x] LabyrinthJS
* [x] LastOneLoses
* [x] Markdown
* [x] MazeWeb
* [x] Measure
* [x] MediaViewer
* [x] Memorize
* [x] MindMath
* [x] Moon
* [x] Paint
* [x] PhysicsJS
* [x] Planets
* [x] Pomodoro
* [x] QRCode
* [x] Record
* [x] Reflection
* [x] SharedNotes
* [x] Speak
* [x] SprintMath
* [x] Stopwatch
* [x] Story
* [x] TamTamMicro
* [x] Tangram
* [x] TankOp
* [x] TurtleBlocksJS
* [x] VideoViewer
* [x] Vote
* [x] Write
* [x] XOEditor
* [x] XmasLights

Please make **one different PR for each activity**.

Before sending your PR, please ensure:
* There is no error in the browser console
[...]

PR #1382, opened 2023-08-06
Replaced WebL10n in Abacus Activity

Vishalk91-4:
Issue #1378 
Replaced WebL10n in Abacus Activity


llaske:
Seems to be a good start but the tutorial no longer work:

image

Vishalk91-4:
I'll look into this and change it

ShrutiSharma77:
I apologize if my approach caused any confusion. I'll work on adjusting the code to align with both migration guides and find a solution that doesn't involve an assignment operator.

activities/Abacus.activity/lib/l10n.js
llaske, line 1: The code of this file has been updated [here](https://github.com/llaske/sugarizer/blob/dev/lib/l10n.js).
Could you update your PR with the recent version?

llaske:
Fixed in https://github.com/llaske/sugarizer/pull/1445

PR #1384, opened 2023-08-08
Enhancement: Replacing webl10n with i18next in Chess Activity

ShrutiSharma77:
### Checklist
- [x] Tested the changes thoroughly across different languages and locales.
- [x] Verified that the Chess Activity functions as expected with the new `i18next` localization.

Screenshot from 2023-08-08 13-02-30


activities/Chess.activity/package.json
```diff
@@ -6,5 +6,8 @@
       "sugar-web": "github:sugarlabs/sugar-web/master",
       "domReady": "github:requirejs/domReady/2.0.1"
     }
+  },
+  "dependencies": {
+    "js-ini": "^1.6.0"
```
llaske, line 11: This dependencies is need one time (for migration) but should not stay here.

llaske:
Good start, thanks.

* The tutorial doesn't work in Spanish:

image

* There is an error in the console at launch:

image

* Will be better to disable i18next debug information - coming from [here](https://github.com/llaske/sugarizer/blob/dev/activities/Measure.activity/js/components/SugarL10n.js#L76). It could be a good idea to do another PR to update the the initial file.

image

activities/Chess.activity/package.json
ShrutiSharma77: Removed the dependency js.ini from package.json.

ShrutiSharma77:
@llaske, thank you for the review. I'll make sure to address the issues you mentioned.

ShrutiSharma77:
- [ ] Corrected the issues causing the tutorial not to work in Spanish.
Screenshot from 2023-08-09 17-36-42
- [ ] Resolved the error in the console at launch.
 Screenshot from 2023-08-09 17-14-05
- [ ] Disabled i18next debug information.

activities/Chess.activity/js/l10n.js
```diff
@@ -57,7 +57,7 @@ var Localization = {
 
 		// Activity initialization check
 		const SugarActivity = vm.$root.$children.find(function (child) {
-			return child.$options.name == 'SugarActivity';
+			return child.$options.name = 'SugarActivity';
```
llaske, line 60: I'm not comfortable with this fix. From an algorithmic POV, it has no sense to replace a test condition by an assignment!

More globally, because the Chess activity is a Vue.js activity but don't handle localization using `SugarL10n` component, I think the migration for this activity should be a mix of Vue.js migration guide and Vanilla.js migration guide.
It seems that you've just replaced the existing `l10n.js` file by the one provided in the Vanilla.js migration guide and it's not a good idea here.

I think the content of the file should be adapted here.

ShrutiSharma77: I apologize if my approach caused any confusion. I'll work on adjusting the code to align with both migration guides and find a solution that doesn't involve an assignment operator.

ShrutiSharma77:
@llaske, I have updated the function and removed the assignment operator in the initialization of SugarActivity. Please review the changes and suggest any further improvements.

llaske:
Good job. Thanks.

I've merged it manually [here](https://github.com/llaske/sugarizer/commit/2c7c727111fc31e20091d8cc6fea0e15e269aa96) because you've sent a PR on master branch.

Please, next time send your PR only to dev branch. See [here](https://github.com/llaske/sugarizer/blob/master/CONTRIBUTING.md).

ShrutiSharma77:
Thank you for taking care of the merge manually. I apologize for the misunderstanding in sending the pull request to the wrong branch.

In the future, I'll make sure to send pull requests only to the "dev" branch.

PR #1387, opened 2023-08-12
Enhancement:Replaced webL10n with i18next in Clock activity

shruti2522:
Issue #1378 
- [x] Replaced instances of the `webL10n` library with `i18next`.
- [x] Tested the Clock activity with different language settings.
Screenshot from 2023-08-12 19-38-19


activities/Clock.activity/lib/l10n.js
llaske, line 1: The code of this file has been updated [here](https://github.com/llaske/sugarizer/blob/dev/lib/l10n.js).
Could you update your PR with the recent version?

llaske:
There is an error at load time:

image

llaske:
Fixed in https://github.com/llaske/sugarizer/pull/1405

PR #1389, opened 2023-08-12
Fix tutorial launch on firstscreen.js and improve startTutorial.

UtkarshSiddhpura:
The below code to set language and trigger event should be only for if block
https://github.com/llaske/sugarizer/blob/27d8ed7e5c31ab72f64ce04d46e0a99c4df684d1/lib/l10n.js#L39C5-L41C29
```js
if (locales !== null) { // here ... }
```

### DIFF
--- a/js/firstscreen.js
+++ b/js/firstscreen.js
@@ -94,15 +94,11 @@ enyo.kind({
 
 		// At first launch, display tutorial
 		var that = this;
-		var interval = window.setInterval(function() {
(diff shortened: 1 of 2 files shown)


vinayaknayar (approved):
Changes in l10n seems to be correct.

llaske:
Good job @UtkarshSiddhpura.

@vinayaknayar I think the fix to `l10n.js` should be replicated to other activities using it.

PR #1393, opened 2023-08-14
Remove unnecessary code from l10n.js

UtkarshSiddhpura:
Chess activity has unnecessary code after migration to l18next.


llaske:
You're right, this code come from SugarL10n component and is useless here.
Mentioning https://github.com/llaske/sugarizer/issues/1378 for reference.

PR #1397, opened 2023-08-23
Replaced WebL10n to i18next in Calculate Activity

Roshan-Pawar:
Issue #1378
This PR migrates the Calculate activity to i18next from webL10n library.


llaske:
Good start but some strings are not localized (clear button, placeholder, tooltips)

**Before**
Capture d’écran du 2023-08-26 21-50-37

**After**
Capture d’écran du 2023-08-26 21-51-15

Roshan-Pawar:
Hey @llaske I've tried to debug this issue, but I haven't been able to resolve it. Could you please help me understand what might be causing this issue and how I can fix it?

llaske:
Buttons and placeholder are localized here https://github.com/llaske/sugarizer/blob/dev/activities/Calculate.activity/js/calculate-app.js#L258. You should debug there to see what happens.
Tooltips are automatically localized using the `data-l10n-id` property https://github.com/llaske/sugarizer/blob/dev/activities/Calculate.activity/index.html#L24 coming from webL10n. I don't think there is a such feature in i18next so you have to localize manually these items. Why not in the `transateGui` function above?

llaske:
Fixed in https://github.com/llaske/sugarizer/pull/1458

PR #1398, opened 2023-08-28
Changed EbookReader from webL10n to i18next

Vishalk91-4:
The code and activity are properly working, required files have been changed


activities/EbookReader.activity/lib/sugar-web/l10n.js
llaske, line 1: Why you've added two times this file? I don't see a good reason to duplicate it.

llaske:
Fixed in https://github.com/llaske/sugarizer/pull/1531

Issue #1356, opened 2023-05-06
New enhanced UI for Chat Activity


vinayaknayar:
As mentioned earlier, Chat activity UI needed to be rethought. Here is the Figma Link to the design approved by @llaske 
Figma Link: https://www.figma.com/file/A4RDcgCuTfC4JxQkec02KH/Sugarizer-Chat-Activity?type=design&node-id=0%3A1&t=XkDWZuwypF1illlz-1
Image: 
![image](https://user-images.githubusercontent.com/98237998/236647017-568a11f9-256f-4c7b-b8c3-d42fe0de2da5.png)

samyakshah3008:
Hey @llaske , I believe this issue is resolved. can we close this? Since it is good to have open issues which actually needs some work to be done. 😄 
If there is some work required, can you highlight it?

llaske:
@samyakshah3008 you're right.
Fixed in https://github.com/llaske/sugarizer/pull/1399

samyakshah3008:
Cool ✅

PR #1399, opened 2023-08-28
Rewrite Chat activity in Vue.

UtkarshSiddhpura:
The new UI of the Chat activity is shown here #1356
I've done some changes in UI (open for feedback).


llaske:
Very cool :+1: 

Few suggestions:

* [ ] When several users are typing their name are mixed
* [ ] I think the typing detection could be simplified: an user is typing while there is some text in its input

ziJEzXTj5q

* [ ] When the user is not connected, the input box should be grayed/disabled with a message inside suggesting to user to share the activity. It's specially important when the activity is reopen from the Journal

llaske:
Nice !
My remarks:

* [ ] Doesn't support multiple typing. Only one user is mentioned

Capture d’écran du 2023-09-16 10-58-19

* [ ] Typing mention disappears when you have the focus on the typing area or if you start typing

V9bVd98mYw

* [ ] It will be nice to add a clear button in the toolbar to clean the history (use the "standard" clean icon from other activities)

llaske:
Thanks @UtkarshSiddhpura . I think it's almost ready to release.
Some remarks:

* [ ] There is a vuejs warning

image

* [ ] It seems that localization doesn't work

image

UtkarshSiddhpura:
>     * [ ]  There is a vuejs warning

I don't see any warning can you please check again.

llaske:
Few remarks:

* [ ] Seems that the UserJoin / UserLeft localization strings are missing. I guess it's now here because you've reintegrated humane popup. May be this notification is useless - except in the case below - since there is an explicit message in the conversation.

Capture d’écran du 2023-10-18 22-01-24

* [ ] When an user left and the user is reading previous messages, the popup "New message from X" is displayed. This popup is not pertinent.

Capture d’écran du 2023-10-18 22-03-45

llaske:
Nice. Thanks.

Issue #787, opened 2020-04-22
XSS injections on Labyrinth JS and Paint Activity


lp1dev:
Hi!
Going through the code again I noticed two XSS injections on sugarizer.
The first one is on labyrinthJS, the text you input can be interpreted with a payload such as (watch out, whitespaces are important).

`</><img src=h onerror=alert(JSON.stringify(localStorage)) />`

The second one concerns PaintJS and is even more interesting as I think it could be triggered through a shared drawing on someone else's device : 
When you add a text area to your canvas you can escape JavaScript with a payload such as :

`<img src=h onerror=alert(JSON.stringify(localStorage)) />`

I'm not 100% sure about the best way to solve this for sugarizer, but I can write a PR if this solution suits you : 
I noticed you already have an input sanitization function (https://github.com/llaske/sugarizer/pull/231/files) which doesn't seems to be shared between activities, maybe it would be interesting to make it global so it can be used anywhere? 

It seems to me that the removeTags function is safe enough, I tested it with examples from Jason Haddix's XSS payloads (https://raw.githubusercontent.com/danielmiessler/SecLists/master/Fuzzing/XSS/XSS-Jhaddix.txt) and custom payloads, it didn't fail to sanitize my inputs!

Cheers,

llaske:
You're right. Thanks. 
At first, I suggest to repeat the same fix. There's no so much activities with input then rendering.

llaske:
Fixed in https://github.com/llaske/sugarizer/pull/1402

PR #1402, opened 2023-08-29
Replace innerHTML with innerText to avoid XSS injections in labyrinth…

UtkarshSiddhpura:
… and paint activity.

It resolves the issues discussed here #787


llaske:
Good. Thanks.

PR #1405, opened 2023-09-22
Changed code from WebL10n to l10n in Clock Activity

Vishalk91-4:
Fixes issue #1378 

Replaced Code in Clock Activity to l10n
All Tutorials are working nicely, no error in console


llaske:
Nice work. Thanks.

Issue #1404, opened 2023-08-31
Foodchain Activity, Score gets added up when activity is reopen


Vishalk91-4:
Hi @llaske, in the Foodchain Activity, 

I found that, after restarting the game, the score of my previous game also adds up to it
![Screenshot 2023-08-31 221032](https://github.com/llaske/sugarizer/assets/116670999/08e40329-fabb-4c51-b87a-a0eabd257dea)
![Screenshot 2023-08-31 221051](https://github.com/llaske/sugarizer/assets/116670999/7d2f0887-a297-4f66-bfb4-2e203cc7d4bf)

Images overlap, and the game cannot be started after a snake has eaten, at the initial point i.e the position of snake should change after one life is gone, otherwise 2 lifes are gone at same time, since there is a delay of revival
![Screenshot 2023-08-30 071712](https://github.com/llaske/sugarizer/assets/116670999/50c29754-4af4-4079-8902-0c82a214075c)
![Screenshot 2023-08-31 214337](https://github.com/llaske/sugarizer/assets/116670999/e7c55660-ec19-4e4c-ad57-9dcc7e09b25c),

llaske:
That's true for the score, image overlap is already declared here https://github.com/llaske/sugarizer/issues/1191

Vishalk91-4:
@llaske, can I work on it to fix this issue

llaske:
@Vishalk91-4 sure

gauravsingh94:
@llaske  I have  resolved the reported issue. Please take a moment to review the changes and provide your feedback.

llaske:
I've reviewed the full activity.
It appears that the score is kept from one game to the next in the whole activity (Learn, Build and Play).
It seems that it score is more to show the investment of the player in the game than to compete with other players.
So we don't need to change the current behavior.
Thanks for your help @Vishalk91-4  and @gauravsingh94

PR #1406, opened 2023-09-22
Fixed the error in score in FoodChain Activity

Vishalk91-4:
Fixes issue #1404 

Made the score to zero in FoodChain Activity, after clicking the home button once a game is finished, so that it doesn't get added up
No error in console

### DIFF
--- a/activities/FoodChain.activity/playgame.js
+++ b/activities/FoodChain.activity/playgame.js
@@ -699,6 +699,8 @@ enyo.kind({
 		this.$.timer.stop();
 		this.$.timerMonster.stop();
 		FoodChain.goHome();
+		FoodChain.context.score = 0;
+		this.$.score.setContent(String("0000"+FoodChain.context.score).slice(-4));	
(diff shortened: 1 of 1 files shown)


llaske:
It's a partial fix. It works when the user go back to home but it doesn't works when the user stop the activity during the game.

Vishalk91-4:
@llaske, I mean if the user pauses during the game and plays the score shouldn't be set to reset there
Other than that, after pausing if the user clicks onto home icon, then it reset's the score

https://github.com/llaske/sugarizer/assets/116670999/d7574a0d-a1ef-4f58-bc5d-ac7b1a64208b

llaske:
> @llaske, I mean if the user pauses during the game and plays the score shouldn't be set to reset there
> Other than that, after pausing if the user clicks onto home icon, then it reset's the score

Yes you're right but the score should be reset when the activity is close then reopen.

Ko8ZM6y65s

llaske:
Now useless, see https://github.com/llaske/sugarizer/issues/1404#issuecomment-1811316149

PR #1408, opened 2023-09-26
Changed Code to i18next in Falabracman

Vishalk91-4:
Changed the code from webL10n to l10n in Falabracman Activity
No Error in console


llaske:
Thanks

PR #1413, opened 2023-10-24
Update Etoys Activity to support saving of Etoys projects

codefrau:


llaske:
Thanks for this contribution @codefrau :+1: 
Good to see you're still working on Etoys.
Time has passed since we met in Brussels :sweat_smile:

PR #1416, opened 2023-11-23
Tweniee: code from WebL10n to l10n in SharedNotes

Tweniee:
#1378
Replaced WebL10n in SharedNotes


llaske:
Few remarks:

* [ ] Default content text is not localized

image

* [ ]  Tooltips are not localized 

Capture d’écran du 2023-11-23 21-40-32

* [ ] Tutorial is not localized 

image

Tweniee:
> * Tooltips are not localized

I have seen all the merged PR they are also using tooltip statically. so please confirm 

image
image

Tweniee:
I get it now @llaske, can you please have a look now.
thanks for your time
image

activities/SharedNotes.activity/js/activity.js
```diff
@@ -711,16 +711,13 @@ define(["sugar-web/activity/activity", "sugar-web/datastore", "notepalette", "zo
 					tutorial.start(tutorial.tourInit);
 				}, 500);
 			}
+				var defaultLanguage = (typeof chrome != 'undefined' && chrome.app && chrome.app.runtime) ? chrome.i18n.getUILanguage() : navigator.language;
+				var language = environment.user ? environment.user.language : defaultLanguage;
+        		l10n.init(language);
```
llaske, line 716: Could you fix the indentation? Should be align and use TAB

llaske (changes requested):
Nice. It works now.
Could you just fix the indentation issue?

activities/SharedNotes.activity/js/activity.js
Tweniee: Sure I'll fix it give me 10-15 min

Tweniee: image
it was looking good in my VS code but as per your suggestion added the tab

Vishalk91-4:
@Tweniee,  I think it's important for maintenance reason, that's why @llaske was asking you to remove those ig


image
And now it's properly intended as is seen

PR #1422, opened 2023-12-07
Migrated Calculate Activity to l10n

Vishalk91-4:
Fixes Calculate Activity issue #1378

The Tutorials shows undefined message
image

@llaske, could you check out what's the bug here


llaske:
A first issue I can see is that the l10n initialization is missing:

```
env.getEnvironment(function(err, environment) {
			l10n.init(language);
		});
```

It should be somewhere in the activity.

llaske:
Fixed in https://github.com/llaske/sugarizer/pull/1458

PR #1426, opened 2023-12-14
Fixes#1419:Message_when_namebox_of_user_is_empty

kaurjasleen240305:
I have used same warning component of first screen which is used in case 1 screen for showing ErrorRemoteLoading just used another ErrorUsernameEmpty message in locale.in


llaske:
It's not an issue.

PR #1427, opened 2023-12-15
Fix po formatting

chimosky:
Noticed this while trying to upload activity to weblate.

@llaske kindly review.


llaske:
@chimosky PO files are useless since localization of TurtleJS activity is not handle in Sugarizer

PR #1435, opened 2023-12-22
Replaced WebL10n with i18next in FractionBounce activity

VishnuVardhanBR:
Fixes #1378 
Followed the guide, no errors in console, locales work.

FractionBounce_Migration


llaske:
Nice. Thanks.

UtkarshSiddhpura:
Why FractionBounce isn't checked in the migration list?

llaske:
You're right, it should be. I've checked it.

PR #1436, opened 2023-12-22
Replaced WebL10n with i18next in Story activity

VishnuVardhanBR:
Fixes #1378 
Followed the guide, no errors in console, locales work.

Story_Migration

Can we ignore the above warnings?


llaske:
Nice. Thanks.

PR #1445, opened 2023-12-28
Replaced WebL10n with i18next in Abacus activity

VishnuVardhanBR:
Fixes #1378 

Had to remove an eventListener to make it work after referring to qrcode.activity migration.

Abacus_Migration


llaske:
Nice. Thanks.

PR #1447, opened 2023-12-30
Updated Style for Planet Info

HitanshuPrasad:
Updated Style for Planet Info for better user experience. #1448 

Updated Style : 

Screenshot from 2023-12-30 21-02-01
Screenshot from 2023-12-30 21-01-39

### DIFF
--- a/activities/Planets.activity/js/activity.js
+++ b/activities/Planets.activity/js/activity.js
@@ -334,14 +334,14 @@ define(["sugar-web/activity/activity", "sugar-web/env", "sugar-web/datastore", "
 					planetInfo.appendChild(information);
 				}
 
-				document.getElementById("Name").innerHTML = '<p>' + l10n.get("PlanetName") + '</br>' + l10n.get(name) + '</p>';
-				document.getElementById("Type").innerHTML = '<p>' + l10n.get("PlanetType") + '</br>' + l10n.get(type.replace(/\s+/g, '')) + '</p>';
-				document.getElementById("Year").innerHTML = '<p>' + l10n.get("YearLength") + '</br>' + l10n.get("EarthDays", {year:year}) + '</p>';
-				document.getElementById("Mass").innerHTML = '<p>' + l10n.get("Mass") + '</br>' + mass + '</p>';
-				document.getElementById("Temperature").innerHTML = '<p>' + l10n.get("SurfaceTemperature") + '</br>' + temperature + '</p>';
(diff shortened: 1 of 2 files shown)


activities/Planets.activity/js/activity.js
```diff
@@ -334,14 +334,14 @@ define(["sugar-web/activity/activity", "sugar-web/env", "sugar-web/datastore", "
...
-				document.getElementById("SunDistance").innerHTML = '<p>' + l10n.get("SunDistance") + '</br>' + sunDistance + '</p>';
+				document.getElementById("Name").innerHTML = '<p>' + l10n.get("PlanetName") + '</p><p>' + l10n.get(name) + '</p>';
+				document.getElementById("Type").innerHTML = '<p>' + l10n.get("PlanetType") + '</p><p>' + l10n.get(type.replace(/\s+/g, '')) + '</p>';
+				document.getElementById("Year").innerHTML = '<p>' + l10n.get("YearLength") + '</p><p>' + l10n.get("EarthDays", {year:year}) + '</p>';
+				document.getElementById("Mass").innerHTML = '<p>' + l10n.get("Mass") + '</p><p>' + mass + '</p>';
+				document.getElementById("Temperature").innerHTML = '<p>' + l10n.get("SurfaceTemperature") + '</p><p>' + temperature + '</p>';
+				document.getElementById("Moons").innerHTML = '<p>' + l10n.get("NumberOfMoons") + '</p><p>' + moons + '</p>';
+				document.getElementById("Radius").innerHTML = '<p>' + l10n.get("PlanetRadius") + '</p><p>' + radius + '</p>';
+				document.getElementById("SunDistance").innerHTML = '<p>' + l10n.get("SunDistance") + '</p><p>' + sunDistance + '</p>';
```
samyakshah3008, line 344: Can't we have a `div` ? any particular reason for `p`? because it will be then paragraph inside paragraph. Parent element is already paragraph. How about going with simple `div`? thoughts?

llaske:
Hmmm. Will be better to have bold on label only, not on value

HitanshuPrasad:
Updated the style for only labels to be bold.

activities/Planets.activity/js/activity.js
HitanshuPrasad: That would be more organized way but will make no difference the activity. I had same doubt when I was working on it.

samyakshah3008: Correct, output won't be affected here, but if it improves the code why not prefer it? 
@llaske your thoughts here?

llaske:
Thanks.

PR #1452, opened 2024-01-01
Replaced WebL10n with i18next in XOEditor activity

VishnuVardhanBR:
Fixes #1378 

XOEditor_Migration

Could you check @llaske


llaske:
Nice. Thanks.

PR #1455, opened 2024-01-03
Change hover effect on popup items

### LINKED ISSUE #1441: Separate hover effect for icon and text in settings popup
The icon and the text currently have separate hover animations. They should function as a unified element, with the hover effect applying to the entire list item as a whole.

image

VishnuVardhanBR:
Fixes #1441
The icon itself has a separate hover animation

Any feedback @llaske?


llaske:
Nice fix :+1: 
BTW could we avoid the double highlight on icon?

Capture d’écran du 2024-01-03 22-01-21

VishnuVardhanBR:
We could not include the "icon" class (which has this hover effect) when we are using the Icon component in a popup, since we already are using the item-icon class. 

What do you think?

llaske:
> What do you think?

Update your PR, I will have a look.

VishnuVardhanBR:
Added a prop to handle it, Could you check this? @llaske

llaske:
@VishnuVardhanBR smart fix :clap: 

Two remarks:
* I'm not a big fan of the property name `includeIconClass` , I think `disableHoverEffect` will be better.
* With this fix the popup header item lost fully its hover effect effect, it will be nice to fix the header item too.

Fgnrw108oI

VishnuVardhanBR:
Made changed, could you check this @llaske ?

llaske:
The idea was to highlight the whole header to be consistent with items.

VishnuVardhanBR:
> The idea was to highlight the whole header to be consistent with items.

I misunderstood, made changes now, can you please check @llaske ?

llaske:
Good. Thanks

PR #1458, opened 2024-01-05
Replaced WebL10n with i18next in Calculate activity

VishnuVardhanBR:
Fixes #1378 

Screenshot


llaske:
Finally I don't think it's a good idea to remove `data-l10n-id` feature. 
It's possible to keep a similar way to localize HTML components in i18next. See https://github.com/llaske/sugarizer-doc/issues/42#issuecomment-1777916993

VishnuVardhanBR:
Could you take a look at [4edd835](https://github.com/llaske/sugarizer/pull/1458/commits/4edd835e75935b94216a36b63e37bb54903a0ce4) @llaske 

One more doubt, should the localization for "calcul", "label", "clear" be in translateGui() or should it be changed to use data-i18next?

llaske:
That's fine. Thanks.

PR #1464, opened 2024-01-07
Replaced WebL10n to i18next in GameOfLife

Vishalk91-4:
Migrated Game of Life activity to i18next

image

@llaske, could you please review this PR


llaske:
The activity no longer work and there is an error in the console.

image

llaske:
Fixed in https://github.com/llaske/sugarizer/pull/1481

PR #1472, opened 2024-01-09
# 1378 Replaced WebL10n with i18next in Chart activity

Rohan20-10:
Replaced WebL10n with i18next in Chart activity

PR #1475, opened 2024-01-09
# 1378 Replaced WebL10n to i18next in Maze Web Activity

Rohan20-10:
Replaced WebL10n to i18next in Maze Web Activity
Screenshot 2024-01-09 202833


Vishalk91-4 (changes requested):
@Rohan20-10, actually when you first made changes for the first PR, I think #1472, you made them in mastere branch of yours, so that change get's added in every different branch you create out of master

PR #1484, opened 2024-01-13
#1378 Replaced WebL10n to i18next in Planets Activity

Rohan20-10:
@llaske pls review the pr
attaching ss of the results
#  Replaced WebL10n to i18next in Planets Activity
Screenshot 2024-01-13 181608


llaske:
Good. Thanks.

PR #1489, opened 2024-01-15
#1378 Replaced WebL10n to i18next in Write Activity

Rohan20-10:
@llaske pls review this pr
attaching ss of the results
# Replaced WebL10n to i18next in Write Activity
Screenshot 2024-01-15 192918


llaske:
Nice. Thanks

PR #1486, opened 2024-01-14
#1378 Replaced WebL10n to i18next in Reflection Activity

Rohan20-10:
@llaske pls review this pr
attaching ss of the results
# Replaced WebL10n to i18next in Reflection Activity
Screenshot 2024-01-14 200442

PR #1490, opened 2024-01-15
Replaced WebL10n with i18next in Reflection activity

VishnuVardhanBR:
Fixes #1378 
No errors in console, added locales.

Could you check @llaske ?
Screenshot


llaske:
Already fixed in https://github.com/llaske/sugarizer/pull/1486

VishnuVardhanBR:
My bad, I haven't checked before making the pull request,
but upon checking #1486, I've noticed that the tooltips aren't localized, can you confirm @llaske?

PR #1492, opened 2024-01-16
Replaced WebL10n with i18next in Paint activity

VishnuVardhanBR:
Fixes #1378 

@llaske could you check?

Screenshot


VishnuVardhanBR:
You're right, my bad.

Rohan20-10:
> @VishnuVardhanBR, I saw @Rohan20-10 raise the PR for the same activity just a day earlier So, I think you should just check before making a PR
> 
> So, Let's contribute but not on conflicting them with each other's issue

Thanks @Vishalk91-4 for your concern. Even I will surely follow that advice.

Rohan20-10:
> You're right, my bad.

np @VishnuVardhanBR  it happens. 
Thanks for your co-operation.

PR #1494, opened 2024-01-16
Replaced WebL10n with i18next in Moon activity

VishnuVardhanBR:
Fixes #1378 
No errors in console, locales work. 

Screenshot


llaske:
po files should also be removed.

VishnuVardhanBR:
They have been removed, am I mistaken?

llaske:
You're right thanks.

PR #1498, opened 2024-01-17
Replaced WebL10n with i18next in Stopwatch activity

VishnuVardhanBR:
Fixes #1378 for Stopwatch activity.

Screenshot


llaske:
Nice. Thanks

soham2312:
@VishnuVardhanBR Hello I want to know that axios.min.js and i18nextmin.js and even l10.js will remain same in every case
am I correct please if I am wrong please correct me

VishnuVardhanBR:
@soham2312 Yes, axios.min.js and i18next.min.js will remain the same, l10n.js may require some changes in some activities but is mostly the same. 

You can check the [guide](https://github.com/llaske/sugarizer/blob/dev/docs/webL10n2i18n.md) here if you haven't already.

PR #1517, opened 2024-02-07
fix #1516 sugar_history get removed when user delete its account

### LINKED ISSUE #1516: [v2] Remove user from history on delete account
## Issue:
By deleting user-account from privacy settings (Delete server account), the user is not removed from local storage. Hence visible on firstscreen.

## Fix:
use `removeUser` method of sugarizer history module to remove user from local storage.

riteshsinghh-coder:
Fix #1516 
sugar_history get removed when user delete its account so user doesnot show on first page

https://github.com/llaske/sugarizer/assets/93989204/e93aea43-5307-4c34-a3d7-233a4cb1a3aa


llaske:
Nice. It fix the issue but:

* [ ] I'm not a big fan of the fix, it will be better to avoid the exception than catching it
* [ ] Please respect the code indentation (TAB instead of SPACE)

image

UtkarshSiddhpura:
@llaske Also do you think that we should somehow store the user already fetched on mount rather then re-fetching the user.
https://github.com/llaske/sugarizer/blob/9e5c7c45283d21a1361a07e38938fa0c9ab39b2b/v2/js/screens/settings-myprivacy.js#L106
https://github.com/llaske/sugarizer/blob/9e5c7c45283d21a1361a07e38938fa0c9ab39b2b/v2/js/screens/settings-myprivacy.js#L130

llaske:
Nice. Thanks.
That's fine to reload the user @UtkarshSiddhpura

PR #1526, opened 2024-02-14
fix: Update tests to accommodate changes in v2 and test utils API

UtkarshSiddhpura:


llaske:
Nice start but there is still a failed test.

image

UtkarshSiddhpura:
> Nice start but there is still a failed test.
> 
> 

All tests are passing on my side can you locate which test is failing? It's probably 1 of the four.

llaske:
It's in password component

image

UtkarshSiddhpura:
> Nice start but there is still a failed test.
> 
Resolved

llaske:
That's nice but is it possible to remove all `console.warn`?

image

UtkarshSiddhpura:
I guess all the warnings are related to props datatypes (for e.g icon `x` , `y` accepts Number but all the implemented-screens/components-using-icon are passing it a string)
So should i change the prop data-type or everywhere it's passing it as string?

llaske:
@UtkarshSiddhpura I think that it's okay for `x`, `y` or `color` to be a number but there is no reason to force `id` to be a number.

UtkarshSiddhpura:
@llaske I think we should setup something so before pushing or before merging we run all the test cases.

UtkarshSiddhpura:
> Not sure to understand. What do you want to setup? What prevents to fix warnings?

Something like a pre-push hook that only pushes if all the test are passed

llaske:
Good job. Thanks.

> Something like a pre-push hook that only pushes if all the test are passed

Got it. You mean action to launch unit testing before commit? I think it's more something to put in Github actions (like for Sugarizer Server [here](https://github.com/llaske/sugarizer-server/blob/dev/.github/workflows/testingWorkflow.yaml)).

PR #1527, opened 2024-02-14
fix: long pressing keys and selection in password input box

### LINKED ISSUE #1515: [v2] password.js Component Issues
### Issues found
1. **Long pressing keys doesn't work**: we could ignore other keys but being not able to long-press `backspace` could be annoying to users.
2. **Ctrl keys combination**: e.g `ctrl+a`, `ctrl+backspace`.

### Possible Fixes
1. Handle all long presses or only backspace key.
2. Remove text highlights of input selection. Handle `ctrl+backspace` by checking `e.target.value === ""`.

UtkarshSiddhpura:
#### Fixes #1515 
- This fix tries to make password.js component to behave like the password component in v1.

### DIFF
--- a/v2/js/components/password.js
+++ b/v2/js/components/password.js
@@ -11,7 +11,7 @@ const Password ={
 					<div class="password-input">
 						<input 
 							class="password-value" id="text" ref="password"
-							v-on:keyup="keyEntered" name="password"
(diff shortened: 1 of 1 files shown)


llaske:
Smart idea :+1:

PR #1533, opened 2024-02-17
Replaced WebL10n with i18next in Markdown activity

VishnuVardhanBR:
Fixes #1378 
Locales work, no errors.


llaske:
In case of fallback, the sample text is not localized.
What is it:

image

What it should be:

image

VishnuVardhanBR:
Unable to reproduce, any specific steps?

Screenshot

llaske:
> Unable to reproduce, any specific steps?

Hmmm. Guess you're running Sugarizer App locally (using Electron or file:///), I've tested on a local Sugarizer server.
It can make a difference because there is a delay loading files.

VishnuVardhanBR:
I'm unable to reproduce on the server. I tried throttling the cpu and network using devtools, but couldn't see the issue, is there any another way?

Screenshot 2024-02-20 at 1 22 34 AM

VishnuVardhanBR:
> I guess the problem is that [this line](https://github.com/llaske/sugarizer/blob/dev/activities/Markdown.activity/js/activity.js#L12) could be executed before locale initialization. 

Most likely. I can fix it but I just can't reproduce it to test before committing, please take a look below.

https://github.com/llaske/sugarizer/assets/64434937/4572747b-7462-42fb-8807-b00d4c007143

llaske:
Strange. It happens each time on my side.

wE9tB0e4a7

VishnuVardhanBR:
I made some changes but i'm unable to validate whether it works or not, what do I do?

Vishalk91-4:
@llaske 

I tried in my system as well, it's showing the same thing, perhaps a localization issue
Screenshot 2024-02-25 183233

soham2312:
@llaske please review the Indentation changes

Issue #1536, opened 2024-02-21
Player without face and color


hackorlyf:
![PlayerNotSpawned](https://github.com/llaske/sugarizer/assets/130314506/4caf7624-869c-4a5c-a110-8d9b245f1462)

hey @llaske there are times I faced this minor bug on maze web activity where the player initially is faceless and of a grey color until and unless we give it an input. This is at total random and I had to restart twice from start to re-create it. I want to work on it.

llaske:
Can't reproduce it on my side. Try to find the sequence to reproduce it.

hackorlyf:
Okay I'll look into it.

hackorlyf:
@llaske just keep restarting the maze web activity by the button provided in the activity itself and play upto 7-8 level. It shows at the start state so don't give any input until the whole maze is loaded. You'll be able to reproduce it somewhere in between restarting 3-4 times.

llaske:
You're right, I've reproduced it

![image](https://github.com/llaske/sugarizer/assets/981841/bf049613-619d-47ae-b46c-69cdca7597b4)

hackorlyf:
I wanna contribute in helping you fix this bug

llaske:
You're welcome to send a PR

hackorlyf:
any idea what part of code might be causing this bug?

llaske:
nope

sonisumit7904:
There might be a refress action being performed after every input, because of which after input the face is visible, but maybe sometimes at the start of the game, that refresh action doesn't work, 
You could use this as a starting point for searching the bug.

hackorlyf:
Thank you so much @sonisumit7904 Got stuck sometime ago but I'll look further into it.

sonisumit7904:
@hackorlyf Happy to help!

llaske:
Fixed in https://github.com/llaske/sugarizer/pull/1541

PR #1541, opened 2024-02-29
fix: Mazeweb.activity sometimes having player without face.

UtkarshSiddhpura:
- The Unexpected behaviour mentioned in issue #1536 occurs randomly it's not related to restarting the activity or anything.

- The behaviour is due to calling `runLevel` function several times without clearing the `dirtyCells` array/state.
- if we call `runLevel` again and `dirtyCells` still have some cells left to draw from previous level, it'll try to draw it in another level.

- I also reproduced the unexpected behaviour by setting `maze.goalPoint` to same as `maze.startPoint` (which was in actual code happening randomly) and then with `setTimeout` changing the `goalPoint` again.

https://github.com/llaske/sugarizer/assets/90092555/5efe0905-6a90-4b62-9f4c-c1737e7852c8

### DIFF
--- a/activities/MazeWeb.activity/js/activity.js
+++ b/activities/MazeWeb.activity/js/activity.js
@@ -37,9 +37,7 @@ define(["sugar-web/activity/activity","tween","rAF","activity/directions","sugar
             }
 
             // Load from datastore
-            if (!environment.objectId) {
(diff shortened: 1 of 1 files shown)


llaske:
Good job. Thanks.

Issue #1546, opened 2024-03-06
[v2] Language Settings: En cannot be selected


Vishalk91-4:
@llaske
In v2, the language en, present as popup-title cannot be selected on changing the language

https://github.com/llaske/sugarizer/assets/116670999/c36a1d4b-4474-4900-9293-8992c4d94e15

llaske:
You're right.

d1vyanshu-kumar:
@Vishalk91-4 I am interested in solving this issue, can I work on this issue?

llaske:
> @Vishalk91-4 I am interested in solving this issue, can I work on this issue?

No need to ask. Feel free to send a PR to a fix, see https://github.com/llaske/sugarizer/blob/dev/CONTRIBUTING.md

llaske:
Fixed in https://github.com/llaske/sugarizer/pull/1548

PR #1548, opened 2024-03-08
Fixed selection of popup-title in language settings

Vishalk91-4:
@llaske 

Fixes en language issue #1546 
Also added time lapse as the popup was getting removed instantaneously when cursor was removed

https://github.com/llaske/sugarizer/assets/116670999/915df224-c9bc-4204-8a1a-ec493208b42c

Coud you please review it

### DIFF
--- a/js/journal.js
+++ b/js/journal.js
@@ -386,7 +386,7 @@ enyo.kind({
 		}
 		this.getToolbar().setMultiselectToolbarVisibility(false);
 	},
-      
+
 	removeSelected: function() {
 		this.dialogAction = constant.journalRemove;
 		this.$.warningbox.setTitle(l10n.get("Erase"));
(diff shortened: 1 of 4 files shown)


llaske:
Not a big fan of this fix: if it's not found it means that it's English.
I think that there is an error in the current implementation: English should be the title.
I suggest a fix by updating only one line, the one to activate the title:

image

Vishalk91-4:
@llaske 
Yes that seems better
I thought in other way to set this as default, but yes the one you mentioned works

llaske:
Okay.
Please remove the timer part since it's not needed.

llaske:
Please, change only the updated file.

Vishalk91-4:
> Please, change only the updated file.

Yes, Sorry, some local changes got into it
Fixed it now

llaske:
Nice. Thanks.

Issue #1553, opened 2024-03-08
In Tank Operation the defeat music persists even after returning to a previous screen.


d1vyanshu-kumar:
Restart Button: Adding a “Restart” button directly on the game over the screen would be convenient for players. This way, they won’t need to navigate back and then click the start button. It streamlines the process and makes restarting quicker.

The defeat music continues even after returning to the previous screen.
@llaske does this issue look good?  can I start working on this?

Activity name: Tank-operation.

https://github.com/llaske/sugarizer/assets/154233802/a6bc0588-00f1-4e81-b7ae-0a74f1883536

llaske:
> Restart Button: Adding a “Restart” button directly on the game over the screen would be convenient for players. 

I'm not in favor of changing the game play.

> The defeat music continues even after returning to the previous screen.

You're right, it's a bug.

llaske:
Fixed in https://github.com/llaske/sugarizer/pull/1556

PR #1556, opened 2024-03-11
Fix: defeat music stops correctly during navigation.

d1vyanshu-kumar:
The defeat music in the Tank Operation activity persists even after returning to a previous screen. This PR resolves the issue by ensuring that the defeat music stops correctly during navigation.
here is how I solve this issue:
by using the if statement i.e if the game ends && the screen is also changed then the sound will pause 
Activity name: Tank-operation.
Issue: #1553 
@llaske please review this PR.

### DIFF
--- a/activities/TankOp.activity/play.js
+++ b/activities/TankOp.activity/play.js
@@ -330,11 +330,7 @@ enyo.kind({
 
 			// Back to app
 			app.renderInto(document.getElementById("board"));
-		}
-		if(this.endOfGame){
-			if(app.renderInto(document.getElementById("board"))){
-				sound.pause();
-			}
+			sound.pause();
 		}
 
 		// Compute direction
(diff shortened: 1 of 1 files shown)


d1vyanshu-kumar:
@llaske please review this PR.

activities/TankOp.activity/play.js
```diff
@@ -57,6 +57,7 @@ enyo.kind({
 	create: function() {
 		this.inherited(arguments);
 		play = this;
+		pause = this;
```
llaske, line 60: What's the objective of this line?

d1vyanshu-kumar: @llaske I am sorry about that it was just a mistake I removed it and pushed it again.

d1vyanshu-kumar:
@llaske please review it now i think it looks good now.

llaske:
Why do you duplicate the same test condition?

image

llaske:
Not sure to understand your explanation.
Simpler is better: do not duplicate the if statement and add the `pause` call in the existing one.

d1vyanshu-kumar:
@llaske kindly review this now i think it looks good now.

d1vyanshu-kumar:
@llaske please review this pr.

Issue #1551, opened 2024-03-08
Ability to resize images in Fototoon


VishnuVardhanBR:
Images are set to fit the entire page, would be nice if we can adjust the size.

Nirvanjha2004:
Is this issue closed or can i work on it  @llaske?

llaske:
It's still open

AdityaKrSingh26:
Hey @llaske please have a look at my pr for this issue

AdityaKrSingh26:
Hey @llaske i have added the fix you mentioned, Please have a look at it and let me know if there is any other improvment i can make.

Pr : #1698

llaske:
Fixed in https://github.com/llaske/sugarizer/pull/1916

PR #1562, opened 2024-03-14
Issue/#1551

hharshas:
resizeimg1
Earlier, the image was displayed via backgroundImage api, so its only way to resize it was to resize '#mainCanvas' div ~ entire page, that will be no good. 
I tried to use existing Globe 'TYPE_Rectangle' by adding bitmap (easel js) to it, but many other component are "TYPE_RECTANGLE" and it is hassle to make each component behave same after these changes (This is my final result but bitmap fills image in parent element). 
Final solution that i can understand is creating a unique Globe type "TYPE_IMAGE" and working with it, that may takes time. If someone is willing to help me with this then i will continue. #1551


llaske:
Please send your PR to the dev branch since it's not related to Sugarizer v2.

PR #1566, opened 2024-03-15
 Blockrain ~ Game can start before pressing play button 

### LINKED ISSUE #1563: Blockrain ~ Game can start before pressing play button
If pressed any keys or clicked any buttons, game starts automatically and also shows "game over" at the end, even when play button is on the top from starting.
Also down arrow button don't help much as we have click multiple times unlike down arrow key. So when button clicked down it should work the same.

hharshas:
Fixed #1563 (now play button removes after pressing any key and game starts).


llaske:
Nice. Thanks.

PR #1561, opened 2024-03-14
FIXED : Display a message when there is no activity matching the search in listview

hharshas:

PR #1573, opened 2024-03-25
 Show message when there is no activity matching the search (listview)  #1561 

hharshas:
@llaske #1561 [TODO.md](https://github.com/llaske/sugarizer/blob/feature/v2/v2/TODO.md)


llaske:
It's nice but the icon and the message appears during 1 second before loading the activity.
Would be nice to avoid it.

By1rJlW7er

hharshas:
@llaske review the changes.

llaske:
That's nice but because Sugarizer is use by children and some of them can't read, it's always better to avoid text.
I suggest to replace it by the `spinner-light.gif` already in the repo.

hharshas:
@llaske  there is no image folder in v2, so I've called the gif from outside, but there should be one isn't it ?

llaske:
Yes

hharshas:
@llaske 
so can I copy one to v2 ?

llaske:
Yes copy it instead of referencing the original, the v1 directory is supposed to be removed at end

hharshas:
@llaske You can check the changes now.

llaske:
Nice but please add only the `spinner-light.gif` from now

hharshas:
@llaske  yeah adding other assets will be make it messy for now !!!!

llaske:
Good job. Thanks.

PR #1579, opened 2024-03-29
Added functionality to undo and redo using the keyboard in the Paint Activity

SamarthBagga:
Added the functionality to undo by pressing ctrl and 'z' on the keyboard and redo pressing ctrl and 'y'. I believe this can be done for all the activites supporting undo and redo buttons as it an intuitive way to do these actions and is common everywhere.
What do you think @llaske ?

[screen-capture (2).webm](https://github.com/llaske/sugarizer/assets/69239707/d71c3830-143a-470b-a577-76de21092c07)

### DIFF
--- a/activities/Paint.activity/js/buttons/redo-button.js
+++ b/activities/Paint.activity/js/buttons/redo-button.js
@@ -6,6 +6,11 @@ define([], function () {
   function initGui() {
     var redoButton = document.getElementById('redo-button');
     PaintApp.elements.redoButton = redoButton;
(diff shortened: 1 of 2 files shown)


llaske:
Shortcut are too complex for children. Plus, most f the time Sugarizer is use on tablets.
Do not send a PR before an issue was created and there is comment on it.

SamarthBagga:
> Shortcut are too complex for children. Plus, most f the time Sugarizer is use on tablets.
> Do not send a PR before an issue was created and there is comment on it.

Okay I'll close this one

PR #1580, opened 2024-03-30
Fixes #1453

### LINKED ISSUE #1453: Message Updation in Fraction Activity
Screenshot 2024-01-03 141754

After the game over message, when we click on play and the game starts, 
Then Pausing it befor the first turn, results in showing Game over again instead we could write, Bouncing ball and the fraction on the first turn

@llaske, your views

SamarthBagga:
Fixes #1453 
Updated the help text if someone clicks on play then pause after one game is over.

### DIFF
--- a/activities/FractionBounce.activity/js/activity.js
+++ b/activities/FractionBounce.activity/js/activity.js
@@ -102,12 +102,6 @@ let app = new Vue({
 
 	computed: {
 		helpText: function () {
-			console.log(this.paused)
-			console.log(this.bounceCount)
-			console.log(Score.innerHTML)
-
-				console.log(Score.innerHTML.length)
-			
 			if (this.answer == -1) {
 				return this.l10n.stringHelpClickToStart;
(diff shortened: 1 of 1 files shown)


SamarthBagga:
@llaske , could you please review this PR ?

Vishalk91-4:
https://github.com/llaske/sugarizer/assets/116670999/0b850155-3f66-4add-8aa0-14368c9fa544

@SamarthBagga, I don't think this fixed the issue, it didn't work in my device
Maybe @llaske could check on his device and verify the same

SamarthBagga:
@Vishalk91-4, Thanks for pointing this out! This problem was arising due to the fact that the Score inner html length is zero at the start of the game when no rounds have been played but when the user has played a round and gotten a score out of ten for that then the string has a length but is equal to " ". I have made the necessary edits for the same, you can try it again.

Vishalk91-4:
> @Vishalk91-4, Thanks for pointing this out! This problem was arising due to the fact that the Score inner html length is zero at the start of the game when no rounds have been played but when the user has played a round and gotten a score out of ten for that then the string has a length but is equal to " ". I have made the necessary edits for the same, you can try it again.

Works for me now

PR #1586, opened 2024-04-06
[v2] Moved getJournal to mounting

Vishalk91-4:
Moved Journal from loading everytime in drawing to Mounting


llaske:
Sounds like a good idea to move it there.
I thought that there was a drawback to do that but it seems to work.

PR #1590, opened 2024-04-07
[v2] Removing unwished empty section in activities

### LINKED ISSUE #1588: [v2] Activities poup has unwished empty section
When there is no instance of the activity in the Journal, the activity popup has unwished empty section.

Capture d’écran du 2024-04-07 10-59-18

Should be:

Capture d’écran du 2024-04-07 11-00-39

Vishalk91-4:
Fixes #1588 

Removed the empty section of activities

image


llaske:
Nice. Thanks.

Issue #1600, opened 2024-04-18
Gear ~ unexpected ui of Play toolbar icon.


hharshas:
After pressing play toolbar icon, when we press other toolbar icons, the play action stops ,but the play svg on the play toolbar icon remains instead of switching to the pause svg.
![Screenshot from 2024-04-18 21-51-09](https://github.com/llaske/sugarizer/assets/125907752/a40b0a4f-a2aa-4598-b7ef-398c645ae963)

llaske:
You're right, it's related to https://github.com/llaske/sugarizer/pull/1506

llaske:
Fixed in https://github.com/llaske/sugarizer/pull/1601

PR #1506, opened 2024-01-20
Add pause button in Gears Activity

VishnuVardhanBR:
Fixes #1503 
Added a pause button in gears activity. 
Any feedback? @llaske

PR #1601, opened 2024-04-18
Gear activity play button improvement

hharshas:
#1600 @llaske did the changes to #1506


llaske:
Good idea to factorize UISwitch :+1: 
Thanks.

Issue #1603, opened 2024-04-19
Missing flag handling in ColorMyWorld


llaske:
When there is a missing flag for a country in ColorMyWorld, the popup display undefined before the country name.
It will be better to display nothing in this case.

![2t7nCztPC7](https://github.com/llaske/sugarizer/assets/981841/0034c2d8-d530-4776-ab16-f128f3033670)

vraj-712:
But i think you want to remove that undefined part, right ? not whole pop up box

vraj-712:
Can you please how you want that?? because i am unable to understand your requirements

llaske:
> Can you please describe how you want that?? because i am unable to understand your requirements

The fix suggested by @hharshas is good but is not enough. It solve the case where there is no existing emoji matching the country name in the activity.
But like I mentioned [here](https://github.com/llaske/sugarizer/pull/1604#issuecomment-2068066132), there is an existing emoji for St Vincent Grenadines. The problem is that the name mentioned in the `flag` constant doesn't match the name of the country so just changing the value to `Saint_Vincent_and_the_Grenadines` will make the flag appears.
So it should be nice to fix the value and double-check values for other countries.

piyushmishraa:
hi i would like to work on this issue.Can you assign it to me

yashveeeeer:
@llaske i have made some changes could you review it and i also have some doubts.

llaske:
@yashveeeeer thanks for your contribution.

BTW your PR is not a the right place. It should be done on this repo. See https://github.com/llaske/sugarizer/blob/dev/CONTRIBUTING.md for more.

Plus, as I mentioned [here](https://github.com/llaske/sugarizer/pull/1643) it will be nice to check other missing flags.

yashveeeeer:
Thanks alot @llaske I'll take care of the original repo and I'll check that problem you mentioned.

llaske:
Fixed in https://github.com/llaske/sugarizer/commit/461795ebeb5171f80e707f6b559ae0a42f86be3f

Sushant0124:
@llaske Is the Issue still there

llaske:
No, it's fixed

PR #1604, opened 2024-04-19
improvement in "Missing flag handling in ColorMyWorld" issue

hharshas:
Added a condition to display name only if flag is not defined. #1603 @llaske

### DIFF
--- a/activities/ColorMyWorld.activity/lib/map.js
+++ b/activities/ColorMyWorld.activity/lib/map.js
@@ -19,6 +19,7 @@ define(["activity/ol","print","util","colormyworld","humane","flag","l10n"],
 		});
 
 		window.map.on('click',function(evt){
+			console.log("hello");
 			if(colormyworld.mode==COLORING){
 				var FOUND=false;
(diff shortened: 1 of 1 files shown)


llaske:
Sounds nice.
BTW I've just checked and the matching Emoji exist: https://emojipedia.org/flag-st-vincent-grenadines 
It's probably a name mismatch with the one declared in `flag.js`: https://github.com/llaske/sugarizer/blob/dev/activities/ColorMyWorld.activity/lib/flag.js#L221
Should be interesting to fix it and to ensure there is no mismatch in the flag list.

> [Fun with flags](https://bigbangtheory.fandom.com/wiki/Sheldon_Cooper_Presents:_Fun_with_Flags) :laughing:

llaske:
@hharshas do you still work on this PR?

hharshas:
Naah @llaske I am busy rn, you can assign it to others.

piyushmishraa:
hi i would like to work on this issue ...can you assign it to me please ?

Issue #838, opened 2020-07-29
Undo button for Gears activity


hr1sh1-coding:
Drawing lines (or any other figure that's not closed) results in the gears (through which the lines pass) getting deleted. Its highly  likely one might delete by mistake, in which case having an undo button would be a good thing.

Also, it's not possible to draw very small gears (I'm not sure whether this is a bug or a feature).

llaske:
Hmmm. I'm not sure it's so easy to delete a figure by mistake because you should draw the line exactly over the figure to delete. Because it's not allowed to draw a figure over another figure, I don't think you could do it without intention.

Yes it's not possible to draw very small gears else it will not be possible to join it with others.

hr1sh1-coding:
When one tries to draw gears very close to other gears, there might be overlap, which causes deletion.

llaske:
Okay you're right. It could help in some cases. 
A Redo button should be added too.

hharshas:
@llaske should undo button and redo button be placed among toolbar icons ?

hharshas:
@llaske should undo button also be helpful with location, I mean is it only to monitor the creation and deletion of gears / chains or also to the earlier location from where it is dragged ?

llaske:
@hharshas will be nice on both.

Devmoni:
@llaske Could you please elaborate on what you mean by 'level issue'? (mentioned in : #1609 )
 I'd like to understand it better.

llaske:
Fixed in https://github.com/llaske/sugarizer/pull/1812

PR #1609, opened 2024-04-21
Added undo and redo in gear activity

hharshas:
#838 @llaske added the feature of undo and redo around deletion and addition of gears/chains, which will now enable you to track its location as well.

### DIFF
--- a/activities/Gears.activity/css/activity.css
+++ b/activities/Gears.activity/css/activity.css
@@ -35,6 +35,14 @@ body {
   background-image: url(../icons/edit-clear.svg);
 }
 
+#main-toolbar #undo-button {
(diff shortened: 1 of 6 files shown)


llaske:
Nice.

I've got two issues:

* [ ] Icons Undo/Redo don't appear
* [ ] Gears are not always recreated at the same level (see the orange gear on the top of the magenta one)

rSJhQOQwhp

llaske:
Icons are now visible. BTW it will be nice to add a vertical bar before the undo icon.

The level issue is still there. If it's really complex to fix, we could debate about it.

When reopening an activity and trying to undo, the whole screen become blank.

0VdthB9gJP

Issue #1607, opened 2024-04-20
Alignment issue in Abacus


VishnuVardhanBR:
The alignment can be fixed for the input as it is confusing. 

<img width="475" alt="Screenshot" src="https://github.com/llaske/sugarizer/assets/64434937/5df3bd49-b88a-4a14-923a-4719d6c61d01">

llaske:
Yes, that's right.

vraj-712:
@llaske @VishnuVardhanBR Can please review my changes.
Here i attached screenshot of changes.
![Screenshot 2024-04-23 214808](https://github.com/llaske/sugarizer/assets/139670583/50593d18-b725-42ba-a46a-9436bed4fa93)

llaske:
Fixed in https://github.com/llaske/sugarizer/pull/1612

er-reddyashok:
Hi is this issue still open

vraj-712:
Hi! 👋 No, this issue is no longer open. Thanks for checking in!

PR #1612, opened 2024-04-23
Fix alignment issue. Wrap label and input tag in div and apply css to it.

vraj-712:
## Changes
✅ Wrap label and input tag in div. then apply flex property to it and center items vertically and give some margin padding to it.
❌ Remove Margin from input tags
 so basically it appear as vertical menu.


llaske:
Related to https://github.com/llaske/sugarizer/issues/1607

llaske:
That's better. Thanks.

Note: mention the issue in the body of the PR, not in title.

PR #1613, opened 2024-04-25
Fixed bug that show undefined where country flag is not unavailable.

### LINKED ISSUE #1603: Missing flag handling in ColorMyWorld

vraj-712:
Fixed #1603

Just Add Small if Else Condition to check whether the flag is available or not. if available then display it otherwise do nothing only show country name.


llaske:
That's pretty the same fix than the one proposed on https://github.com/llaske/sugarizer/pull/1604

PR #1622, opened 2024-06-01
Improve the localization handling using i18next-vue

UtkarshSiddhpura:
I've built i18next-vue lib from the source and had to remove the usage of es6 import/export in that file. Instead, I tried using global variables and the Vue object provided by vue.js lib, which allows us to use i18next-vue in our app.


UtkarshSiddhpura:
Added i18next.js module
Updated the app to use this new module

llaske:
Good job @UtkarshSiddhpura , it's far better :+1:

PR #1627, opened 2024-06-07
Add Prompt component to show Display warning on logging off.

UtkarshSiddhpura:


llaske:
That's better but the header bar height is too high.
Try to align the height with the dialog box header bar height.

Capture d’écran du 2024-06-08 22-25-54

UtkarshSiddhpura:
[x ] Include unit testing of the component
[x] It will more Sugar UI compliant to have a dialog with a header
[x] Icon and header text are too big.
[x] Try to align the height with the dialog box header bar height.

PR #1632, opened 2024-06-10
Loading anim while getting response from server

UtkarshSiddhpura:
https://github.com/llaske/sugarizer/assets/90092555/eda054a8-77e2-43fc-926d-cbe22f0aedc5


llaske:
It's a good start but the spinner should be added also in first screen when login on the app.

6QynMeo021

Look for `spinner.setShowing` in v1 https://github.com/llaske/sugarizer/blob/dev/js/firstscreen.js#L272

UtkarshSiddhpura:
> It's a good start but the spinner should be added also in first screen when login on the app.
> Look for `spinner.setShowing` in v1 https://github.com/llaske/sugarizer/blob/dev/js/firstscreen.js#L272

Done 👍

PR #1634, opened 2024-06-17
Fixes #1466: Bug in Calligra Activity when letters deleted

### LINKED ISSUE #1466: Bug in Calligra Activity when letters deleted
In the Calligra Activity,
When some of the alphabets get deleted, and we switch to words section then some of the deleted alphabets are visible and some are not visible in the word being drawn

https://github.com/llaske/sugarizer/assets/116670999/60261be8-d86e-4a0d-8566-60eb01a5d911

@llaske , could you please look after it

mohit3500:
In the Calligra Activity, when some of the letters get deleted by user, and when we switch to word section then some of the deleted letters are not visible.

When the user delete some letters, the letters get splice from defaultTemplate in the database.js into js folder from  Calligra.activity folder, therefore some letters not drawn in the word section.

If we add duplicate template of all the letters in the defaultTemplate file in the database.js while giving visibility false to them so that they do not show into letter section. The duplicate letters are used in case some letters got deleted from letter section.

Before deleting letters
image

After deleting some letters
image
.
Word image after deleting some letters
image

The local storage data must be cleaned before using because it contains older defaultTemplate into it.


llaske:
Hmmm. These "invisible" letters are not the same than the original ones. They have more or less path to handle linked letters.
Duplicate the whole letters could works but it's less than optimal.
Why not use a property "deleted" for deleted letters but not really delete them?

llaske:
Sounds good but I suggest to not initialized the `delete` property to avoid update of `database.js` file.

Plus, please make your PR on dev branch (see https://github.com/llaske/sugarizer/blob/dev/CONTRIBUTING.md).

PR #1637, opened 2024-06-22
Use i18next-vue in sandbox app

UtkarshSiddhpura:


llaske:
The popup doesn't display at the right place.

ZkQr08XkbL

I think the popup test [here](https://github.com/llaske/sugarizer/blob/feature/v2/v2/js/screens/sandbox.js#L215) should be changed.

UtkarshSiddhpura:
> The popup doesn't display at the right place.
> 

- The component popup is never made to work with containers that have Y scrolling.
- As it detect the mouse position on the window, but if we then set the popup at that mouse position it'll be incorrect as we have to take the ` window.srollY (the amount of scroling done in pixels)` into the account.
- Even if we take the *window scrollY* into the account for e.g ` popup.show(x, y + window.scrollY) `
    -  The line here [adjust for out of screen](https://github.com/llaske/sugarizer/blob/feature/v2/v2/js/components/popup.js#L111) will try to adjust the popup (as we're passing `y+window.scrollY`), which is not needed as the actual y is still within the screen.

UtkarshSiddhpura:
The Fix is 
- Use `position: absolute` for a popup which is positioning itself relative to an element of height 100vh. Just show in above commit

PR #1641, opened 2024-06-24
Fix i18next not exporting language.

### LINKED ISSUE #1639: [v2] Localization failed in activities in file:/// mode
In v2 in file mode, the localization failed in activities.
It works in connected mode.

Hxa7K6PVr1

UtkarshSiddhpura:
After some debugging i found that the user.js module needs to access the language from i18next while signing up the user, which the module i18next is not exporting
Here https://github.com/llaske/sugarizer/blob/feature/v2/v2/js/modules/user.js#L100

Fixes #1639


llaske:
Nice. Thanks.

PR #1642, opened 2024-06-26
Fixes[#1466]- Bug in Calligra Activity when letters deleted

mohit3500:
Fixed by changing Remove Item section in activity.js and changing visibleImages in template.js by adding deleted and visible properties into them.


llaske:
Your PR is still on the master branch. I've merged it manually here https://github.com/llaske/sugarizer/commit/58551c1b0ad89c5caae3b0c5ff71b1ef5b42e90e
Thanks.

PR #1643, opened 2024-07-02
Fixed undefined emoji 

piyushmishraa:
fixed the flag emoji by changing name of country to correct one https://github.com/llaske/sugarizer/issues/1603

https://github.com/llaske/sugarizer/assets/116213566/45d41d77-f12c-415e-9683-14d66410a0f6

### DIFF
--- a/activities/ColorMyWorld.activity/lib/flag.js
+++ b/activities/ColorMyWorld.activity/lib/flag.js
@@ -218,7 +218,7 @@ define(function (require) {
         Burkina_Faso: "🇧🇫",
         Uruguay: "🇺🇾",
         Uzbekistan: "🇺🇿",
-        Saint_Vincent: "🇻🇨",
+        Saint_Vincent_and_the_Grenadines: "🇻🇨",
         Venezuela: "🇻🇪",
         United_States_Virgin_Islands: "🇻🇮",
         Namibia: "🇳🇦",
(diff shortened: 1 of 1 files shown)


llaske:
Nice. But should be cool to ensure there are no other issues on flags.
In short: ensure that all countries names in `geojson` files match a country in `flag.js` file.
Could be automatized.
What do you think?

piyushmishraa:
That sounds nice..let me work on it .

piyushmishraa:
hi @llaske ..so i was getting familiar with the code base and i noticed that in map.js,target name and tool_display was being matched and taken care of .i still didnt get why we got this error after this.However, I am still trying to figure out a way so that countries in geojson and flag.js match.if you have something you can guide me.                                                                                                                                                 other thing was the emoji for French Southern Territories was missing so i took care of that.Rest of the flags are alright .

llaske:
My idea is to extract the list of existing countries in the app and comparing it to flag list.
For example, one way to do that could be to:
- Write a script to extract all NAMES in `world.geojson`, `europe.geojson`, ... and put it in a CSV file order by name (and deduplicated). 
- Extract country names in `flag.js` and put it in a CSV order by name (and with _ replaced by white space).
- Compare both CSV files to see where there is missing or misspelled flags.

@piyushmishraa make sense?

piyushmishraa:
Hmm makes sense👍, getting back to work ✌🏼

llaske:
@piyushmishraa do you still work on it?

llaske:
Fixed in https://github.com/llaske/sugarizer/commit/461795ebeb5171f80e707f6b559ae0a42f86be3f

PR #1647, opened 2024-07-17
Fix previous button not working on the select server loginscreen.

UtkarshSiddhpura:


UtkarshSiddhpura:
#### But now there is another issue now the previous button works

1. Add a Server Address
2. Go Back
3. Select New User
4. Now this new user will be automatically be logged in on the server Address.

#### Also if the server Address is Not Valid It Stucks On Loading

llaske:
Another issue: the Done button is not localized

image

PR #1648, opened 2024-07-18
Journal view

UtkarshSiddhpura:


llaske:
I know it's a WIP but two remarks:

* [ ] Miss the screen when no entries match filter condition (here in v1)

image

* [ ] Entries are duplicated when they are open from the remote Journal

4TlWK42f9w

PR #1652, opened 2024-08-06
[GSOC] Human Body Activity

SamarthBagga:


llaske:
Nice start :+1: 

Few remarks:
* [ ] Add a popup when clicking on a part in paint mode to give the name the part - like in Color my world activity

Capture d’écran du 2024-08-10 12-19-18

* [ ] Do not maintain the popup displayed in doctor mode. You could use the same temporary popup than the one in paint mode (see above) and display: the name of the part to find, wrong message repeating the part to find, correct message

image

* [ ] The toolbar should handle three mutually exclusive modes: paint (default), doctor, visit (where an animation zoom randomly on parts - see below)

3St7CVX28a

* [ ] Left and Right are reversed, it should be considered from the body point of view

* [ ] Please use TAB indentation in your source code

llaske:
The project continue on https://github.com/llaske/sugarizer/tree/feature/humanbody/activities

PR #1655, opened 2024-08-19
FIXED -> Vertical bar are broken in Chart/Gears activities on Windows 

### LINKED ISSUE #1646: Vertical bar are broken in Chart/Gears activities on Windows
On Chrome Window, the vertical bar used in activities toolbar to split group of icons is broken.

It look like:

image

and should be:

image

Issue is both on Chart and Gears activity

MdSaifAliMolla:
Fix  #1646

### DIFF
--- a/activities/Chart.activity/index.html
+++ b/activities/Chart.activity/index.html
@@ -33,19 +33,19 @@
 					v-if="SugarPresence"
 				></sugar-toolitem>
 
-				<hr style="margin-bottom:0.2rem;"/>
+				<hr style="margin-bottom:4px;"/>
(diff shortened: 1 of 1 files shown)


llaske:
The fix totally broke the UI on other platforms.

image

Because Sugarizer works on any platform, any fix should preserve the compatibility with every supported platform.

MdSaifAliMolla:
@llaske Sorry to bother  but which platform is it and how can I test if my pr is compatible with other platforms.
I have checked responsiveness using devtools and it worked fine.

MdSaifAliMolla:
@llaske  Updated the pull request. I think the problem should be solved now .

llaske:
> @llaske Updated the pull request. I think the problem should be solved now .

it's not.

image

did you try on your side?

PR #1657, opened 2024-09-02
Solving the problem of Missing flag handling in ColorMyWorld #1603

yashveeeeer:
Hi @llaske, as you had mentioned earlier that there could be other missing flags, I have added all the flags of all official countries  which were missing but there were some like Somaliland, the Unicode Consortium do not provides the flag (emoji code), so emoji is not available on the internet I was unable to add them, but the error is also not there, when we click on a non flag country it only shows the name of the country.
There was problem in "Timor-Leste", due to the hyphen, the code was not able to replace the hyphen part to underscore due to which it had the issue, i have resolved it too. Kindly review it and please let me know if i could make it more better.


yashveeeeer:
@llaske could you please review it?

llaske:
Could you please make the PR on dev branch, see https://github.com/llaske/sugarizer/blob/dev/CONTRIBUTING.md

PR #1659, opened 2024-09-13
Missing flag handling in ColorMyWorld #1603

yashveeeeer:
Hi @llaske, i hope now i am at right place, as you had mentioned earlier that there could be other missing flags, I have added all the flags of all official countries which were missing but there were some like Somaliland, the Unicode Consortium do not provides the flag (emoji code), so emoji is not available on the internet I was unable to add them, but the error is also not there, when we click on a non flag country it only shows the name of the country.
There was problem in "Timor-Leste", due to the hyphen, the code was not able to replace the hyphen part to underscore due to which it had the issue, i have resolved it too. Kindly review it and please let me know if i could make it more better.


yashveeeeer:
@llaske , could you please review it, I am pretty sure I am at right place and I have also followed the contribution guidelines as you had mentioned.

llaske:
@yashveeeeer good job. It's nice that you've fixed flags for other countries 👍 
BTW I'm not a big fan of your test to fix the "Timor-Leste" case.
I suggest instead to replace the hyphen by a blank in originals geojson files.

yashveeeeer:
@llaske, thank you for the review.
Well, initially I also thought, why am I doing a separate case for a country there should be another way, but that time i was just so excited to send a pull request 😂(as i am a newcomer), i will try to do as you said.

yashveeeeer:
@llaske I have made the changes in the geojson files and removed the separate case for "Timor-Leste" and its working perfectly.

llaske:
Nice. Thanks.

yashveeeeer:
@llaske , i was wondering if you could tell me any issue on which i can work upon and learn something new because i have find that most of issues has already been fixed.

Issue #1662, opened 2024-10-03
Postion of Sugar Icon in Popup on homescreen is toward bottom right.


AdityaKrSingh26:
The sugar icon in popup on home screen has alignment issue and is shifted toward bottom right.
![Screenshot 2024-09-04 000113](https://github.com/user-attachments/assets/260e7b44-f40d-4d20-bc8a-df782d57290f)

AdityaKrSingh26:
Hey @llaske have a look into this and please let me know if there are any changes that is needed to be made

llaske:
Nice. Thanks.
I've merged it manually in https://github.com/llaske/sugarizer/commit/7c88bc11bfc1ec7669e7d16b2098da6c65790d25 because you've sent a PR on master branch.
Please next time send PR only to dev branch. See https://github.com/llaske/sugarizer/blob/dev/CONTRIBUTING.md

AdityaKrSingh26:
sure. i will take a note of this

llaske:
Should not be close until released in an official version

PR #1663, opened 2024-10-03
Fix: Sugar Icon alignment in homescreen popup (issue: #1662)

AdityaKrSingh26:
The sugar icon in popup on home screen had alignment issue (margin from top and bottom was a bit too more). Fixed the values of x and y in popup.js.
Before : 
Screenshot 2024-09-04 000113
After : 
Screenshot 2024-10-03 235709

Fix for #1662


llaske:
Fixed in https://github.com/llaske/sugarizer/commit/7c88bc11bfc1ec7669e7d16b2098da6c65790d25

llaske:
Your fix is not related to the same issue. Please open a new issue.

Issue #1661, opened 2024-09-28
unable to play the recorded video in the Record activity


yashveeeeer:
Hi @llaske, while using the Sugarizer I have found that I am not able to play the recorded video in the Record  activity of the Sugarizer.
Could you also check on your side if the problem is same.
I have started working on the same.

llaske:
@yashveeeeer could you detail your testing environment.

yashveeeeer:
Hardware:
System Model: Victus by HP Gaming Laptop
System Type:  x64-based PC
Processor(s): 1 Processor(s) Installed.
                     [01]: Intel64 Family 6 Model 154 Stepping 3 GenuineIntel ~2300 
Windows Environment
OS Version: 10.0.22631 N/A Build 22631

Software:
Nodejs v20.11.1
electron@31.3.1
NPM 10.2.4

if you want other info just let me know.
and also i want to mention one thing when tried the problem on the deployed version of sugarizer on internet using the other laptop of asus vivobook(i used microsoft edge and chrome ) the video was working fine, then i also tried on the third laptop which was also asus vivobook it was not working well and i tried to check sugarizer on my microsoft edge and chrome the playing is not working well even though the version of the microsoft browser is same as of the other one where it was working well.

yashveeeeer:
@llaske could you correct me if I have provided wrong info, I'm pretty sure you might be busy.

yashveeeeer:
Okay, so i guess i have made little mistake, so while solving the error and testing, i use electron app (so i think its an desktop version, because initially when i had forked the repo i had install the electron and when i run the sugarizer locally on desktop using the code, in my task manager the electron app shows working and while doing it end task the sugarizer also get closed) but for the comparison i used the web version of it, so the error might be for the desktop version.
Also i want to mention that  in capture-helper.js file in the video recording code section which uses Recordrtc when i add the mimeType the application works well it shows the recorded video properly on my local electron app.

llaske:
> Okay, so i guess i have made little mistake, so while solving the error and testing, i use electron app (so i think its an desktop version,

Yes, it's the desktop version.

> when i add the mimeType the application works well it shows the recorded video properly on my local electron app

Could you detail exactly what you've done? It could be a fix to the issue.

yashveeeeer:
Okay, so in the capture-helper.js file (path of the file [activities\Record.activity\js\capture-helper.js]) the code here is trying to record the audio , video and capturing picture and also the display part of it.
[...]

Issue #1203, opened 2023-02-27
Recording video doesn't work on Windows


VedantSharma11:
Hello @llaske 
I wanted to bring to your attention an issue I found while using the Record Activity. Specifically, the VideoRecord feature does not seem to be working properly. When I click on the record button it starts recording the video for 5 seconds but when I try to preview it the video player does not work properly and it is not able to play the video, the photo and audio-record feature works properly. I have tested this on several different devices and it breaks only on windows devices, so I believe the issue is with the code itself.

I wanted to report this issue to you so that you could share your thoughts on it.
if you find it as an issue I would like you to assign it to me, thank you.

https://user-images.githubusercontent.com/96445392/221550097-e92797df-ba18-40a7-81e2-e851713714a2.mp4

llaske:
@VedantSharma11 which browser are you using? Is there any message in the console?

VedantSharma11:
@llaske i am not running it on the browser i am running it locally on my system .

VedantSharma11:
@llaske i am not running it on the browser i am running it locally on my system and it runs as a system application, and the above issue exists there.

llaske:
How did you install it? How do you run it?

VedantSharma11:
I forked and cloned the sugarizer repository on my local machine and then i installed the packages using "npm install" and to run it locally i used "npm start"

llaske:
Okay run it with debug option and tell me if there is error message on the devtools console.

     npm start -- --sdebug --window

VedantSharma11:
It gives some warnings on the devtools console. I am attaching the snippet for the same

![Screenshot 2023-02-28 030359](https://user-images.githubusercontent.com/96445392/221690975-c9ec1b20-10d1-47dc-83fd-fcca6a384691.png)

llaske:
Okay. Look like a security issue.
Could you install Sugarizer using the Windows installer instead (see [here](https://sugarizer.org/#apps)) to check if it fix the issue?

VedantSharma11:
@llaske I installed sugarizer using the windows installer and i tried the record activity but the issue still persists

Vishalk91-4:
@llaske, is this issue open to work upon

llaske:
@Vishalk91-4 sure.

llaske:
Fixed in https://github.com/llaske/sugarizer/pull/1664

PR #1664, opened 2024-10-05
unable to play the recorded video in the record activilty

yashveeeeer:
fixing issue #1203  and #1661

### DIFF
--- a/activities/Record.activity/js/capture-helper.js
+++ b/activities/Record.activity/js/capture-helper.js
@@ -532,7 +532,7 @@ define(["activity/recordrtc", "sugar-web/activity/activity", "sugar-web/datastor
                 navigator.mediaDevices.getUserMedia({video: true}).then(function (mediaStream) {
                     var recordRTC = RecordRTC(mediaStream, {
                         type: 'video',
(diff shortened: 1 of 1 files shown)


yashveeeeer:
@llaske would you please review it.

llaske:
Smart fix. Thanks.
Did you had opportunity on others operating system than Windows? 
As you know, Sugarizer is multi-platform so the fix should work on every platform.
I've read that VP8 has broader compatibility, especially in older devices and software. May be it could be better to use VP8 than VP9.

yashveeeeer:
Thanks @llaske, Yeah I will try to see the fix on different OS using Virtual Machine.
and  you are right about VP8 has broader compatibility on older devices than VP9, I just read that , I will change it to that after checking the fix on different OS.

yashveeeeer:
kindly ignore that force push and i have changed VP9 to VP8, I have also check the fix on Linux.
actually the video player is working fine in Linux, the problem is only with the windows environment, when I added the fix the video was still working fine with Linux.

llaske:
Thanks a lot for fixing this very old bug.

yashveeeeer:
My pleasure, I learned a lot, from this bug.

PR #1673, opened 2024-11-21
Fix: vertical bar broken in chart/gear acitivity (issue: #1646)

AdityaKrSingh26:
Fixed the broken vertical line in Chart/Gears activities, ensuring its applicable on all plaforms (issue: #1646)

Steps:
- Divided the toolbar in two parts left and right, used` dislay: flex` and `justify-content: space-between` to move both the part to ends of both side of screen. and used `dislay: flex` and `align-item:centre` to move hr to correct position.
- Earlier the negative margin was causing the cross platform issue for hr. Removed the negative margin.

### Chrome:
image
image

### Firefox:
image
image

### Electron.js app:
image
image

### DIFF
--- a/activities/Chart.activity/index.html
+++ b/activities/Chart.activity/index.html
@@ -21,7 +21,6 @@
 	<body oncontextmenu="return false;">
 		<div id="app" :class="{fullscreen: isFullscreen}">
 			<sugar-toolbar ref="SugarToolbar">
-				<div id="toolbar-left">
 				<sugar-toolitem ref="activityPal" id="activity-button"></sugar-toolitem>
 				<sugar-toolitem
 					ref="sharedPal"
@@ -34,19 +33,19 @@
 					v-if="SugarPresence"
 				></sugar-toolitem>
 
-				<hr />
+				<div class="seprator-line"></div>
 				<sugar-toolitem id="add-button" :title="l10n.stringAddValue" @click="addData"></sugar-toolitem>
 				<sugar-toolitem id="remove-button" :title="l10n.stringRemoveValue" @click="removeData"></sugar-toolitem>
 
-				<hr />
+				<div class="seprator-line"></div>
 				<template v-for="(string, key) in CHART_TYPES" :key="key">
 					<sugar-toolitem @click="setChartType(key)" :id="`${key}-chart-button`" :class="{active: pref.chartType==key}" :title="l10n[string]"></sugar-toolitem>
 				</template>
 				<sugar-toolitem id="csv-button" :class="{active: pref.chartType == 'csvMode'}"  @click="setChartType('csvMode')" :title=l10n.stringViewCsv></sugar-toolitem>
-				<hr />
+				<div class="seprator-line"></div>
 				<sugar-toolitem id="config-button" @click="toggleConfig" class="invoker" :class="{active: configActive}" :title="l10n.stringConfigs"></sugar-toolitem>
 				<sugar-toolitem id="text-palette-button" @click="toggleTextPal" class="invoker" :class="{active: textPalActive}" :title="l10n.stringConfigs"></sugar-toolitem>
-				<hr />
+				<div class="seprator-line"></div>
 				<sugar-toolitem 
 					id="export-button" 
 					ref="exportPal"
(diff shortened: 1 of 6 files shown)


activities/Chart.activity/index.html
```diff
@@ -21,45 +21,49 @@
...
+					<sugar-toolitem
+						ref="sharedPal"
+						id="network-button"
+						:title="l10n.stringNetwork"
+						palette-file="sugar-web/graphics/presencepalette"
+						palette-class="PresencePalette"
+						palette-event="shared"
+						@shared="SugarPresence.onShared"
+						v-if="SugarPresence"
+					></sugar-toolitem>
```
llaske, line 35: Do not change indentation.

activities/Chart.activity/index.html
```diff
@@ -21,45 +21,49 @@
...
-					v-bind:title="l10n.stringExportSettings" 
-					palette-file="lib/js/palettes/exportpalette.js"
-					palette-class="ExportPalette"
-					palette-event="export-file"
-					@export-file="exportFile"
-					:title="l10n.stringExportSettings"
-				></sugar-toolitem>
+					<hr />
+					<sugar-toolitem id="add-button" :title="l10n.stringAddValue" @click="addData"></sugar-toolitem>
+					<sugar-toolitem id="remove-button" :title="l10n.stringRemoveValue" @click="removeData"></sugar-toolitem>
```
llaske, line 39: Do not change indentation.

activities/Chart.activity/lib/sugar-web/graphics/css/sugar-96dpi.css
```diff
@@ -26,16 +26,23 @@ a {
...
+  display: flex;
+  align-items: center;
+}
 /* Toolbar separator */
 .toolbar hr {
   display: inline-block;
   height: 33px;
   margin: 2.5px;
-  margin-bottom: -17px;
+  /* margin-bottom: -17px; */
```
llaske, line 45: Do not let commented code.

llaske (changes requested):
I guess it's related to your IDE settings but do not change file indentation.
Only lines updated should be in the PR.

AdityaKrSingh26:
Ok  i will make the changes you mentioned. Thanks

AdityaKrSingh26:
Hey @llaske i have removed the comments and indentation in my latest commit please have a look

llaske:
It works.
You're not supposed to change the library `sugar-96dpi.css` because it's a shared library copied in every activity.
BTW it could be acceptable if it's a minor change. But using `flex` in the toolbar a major change that will impact all activities. Plus split it in left/right has no logical reason. 
Could you find a way to limit the impact to activity css only or to the `hr` element only?

AdityaKrSingh26:
I looked around a bit and found out that the issue with <hr/ > lines not being consistent was `margin-bottom : -17px` the negative margin is causing cross platform issue as different browser interpret negative margin differently (please correct me if i am wrong). While` display : flex `property is consistent on all platforms.

I split right/left because on applying display flex the style class of pull-right was over written so to move the icons in group in 2 sides of screen i used `justify-content : space-between`

AdityaKrSingh26:
One thing that i can think of is that wrapping <hr/> tag with a <div> tag and apply flex property on parent <div> of <hr/>
Is this fine ?

llaske:
I understand your fix but I would like a fix where only the CSS `.toolbar hr` need to be updated.
Upgrading HTML and use a `flex` property will change too many files.

AdityaKrSingh26:
Hey @llaske .... i looked and tried out diffrent css properties for restricting the changes to `<hr/>` tag only but it seems pretty impossible without changing some properties of the parent tag.

As you mentioned we are not supposed to modify the `sugar-96dpi.css` one thing we can do is add the necessary changes in `acitivity.css` of the specific required files

llaske:
@AdityaKrSingh26 that's fine if you can limit change to `activity.css`

AdityaKrSingh26:
@llaske please have a look at the latest commit.

llaske:
It works but I'm not a big fan of using flex because it change the look and the behavior of the toolbar.
Why not replace the HR object by a DIV with a vertical border? It will avoid changing the toolbar itself.

AdityaKrSingh26:
> It works but I'm not a big fan of using flex because it change the look and the behavior of the toolbar. Why not replace the HR object by a DIV with a vertical border? It will avoid changing the toolbar itself.

Ok i will make the changes

AdityaKrSingh26:
> It works but I'm not a big fan of using flex because it change the look and the behavior of the toolbar. Why not replace the HR object by a DIV with a vertical border? It will avoid changing the toolbar itself.

Done. Please have a look

AdityaKrSingh26:
@llaske  Done :+1: 
Firefox:
Screenshot from 2024-12-18 13-57-16


Chrome:
Screenshot from 2024-12-18 13-57-01


Electron:
Screenshot from 2024-12-18 13-56-34

llaske:
Work also on Safari.
Good job, thanks.

AdityaKrSingh26:
> Work also on Safari. Good job, thanks.

Finally. :sob:

Issue #1674, opened 2024-12-10
Listview Popup increasing page length and opening downwards for lower icons.


AdityaKrSingh26:
In list the popup on hovering for lower icons goes out of current screen height and is not fully visible and for bottom icon increases the page length. As shown in the screen recording.

[Screencast from 2024-12-10 23-34-22.webm](https://github.com/user-attachments/assets/ec06435b-f7d5-4677-9488-499fed3d6a20)

AdityaKrSingh26:
Hey @llaske could you please have look at the issue and pr

llaske:
Fixed in https://github.com/llaske/sugarizer/pull/1675

PR #1675, opened 2024-12-10
Fix: Listview popup increasing page length (issue : #1674)

AdityaKrSingh26:
The pr is a fix to issue : #1674. Added line in `js/popup.js` preventing the popup in listview from increasing the page length for bottom icons and going out of page height for lower icons on screen.

Demonstration : 
[Screencast from 2024-12-10 23-42-58.webm](https://github.com/user-attachments/assets/7b50ae2d-1bf7-4647-a410-b5fcf80d509b)


llaske:
Nice. Thanks.

PR #1679, opened 2024-12-19
[+]Chess Game Fix

### LINKED ISSUE #1678: Unusual Draw in Chess Activity
The engine declares draw, even though there is no draw there 
Which can be seen in the Game Recording Provided Below:

https://github.com/user-attachments/assets/8d3ecd92-ef3d-4cb1-9059-25a81159455a

Note :
This is only the one position in the game , there are many positions which can give the result as draw

MostlyKIGuess:
## Fixes #1678  .
-  here's the preview of me running the same game:
https://github.com/user-attachments/assets/c0f7a553-f4d7-4714-9e28-a9beac432df5

## Changes Made:
- Worked on correcting the conditions as per the README of https://github.com/douglasbagnall/p4wn/tree/master .
- The previous conditions was checking perhaps the possibility of the draw and not the draw actually. So it was not needed.
- Essentially it was implementing what was already implemented in the engine, check
 `P4_MOVE_FLAG_DRAW = 64       a draw is available` from the README.
 
## Further Improvements
- The engine is kind of old, I have added comments to understand what conditions is doing what operations.
- We can use the engine's p4_move2string function to easy our approach in the game. ( I shall perhaps create an issue on this and then work on this). 
- Further tests from the original library can be added if using the same engine.

### DIFF
--- a/activities/Chess.activity/js/chessGame.js
+++ b/activities/Chess.activity/js/chessGame.js
@@ -391,18 +391,24 @@ var ChessGame = {
       var start_time = Date.now();
       var possibleMove = this.state.findmove(this.level);
       var delta = Date.now() - start_time;
+
+      // console.log(`Random Move: ${possibleMove}, Delta: ${delta}`);
+
       // game over
-      if (possibleMove.length === 0) {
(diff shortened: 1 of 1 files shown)


MostlyKIGuess:
I accidentally made it into main branch. Apologies

PR #1680, opened 2024-12-19
Chess Game Fix

### LINKED ISSUE #1678: Unusual Draw in Chess Activity

MostlyKIGuess:
## Fixes #1678  .
-  here's the preview of me running the same game:

https://github.com/user-attachments/assets/96f9fdbd-c8db-4f73-afde-d84e622390ee

## Changes Made:
- Worked on correcting the conditions as per the README of https://github.com/douglasbagnall/p4wn/tree/master .
- The previous conditions was checking perhaps the possibility of the draw and not the draw actually. So it was not needed.
- Essentially it was implementing what was already implemented in the engine, check
 `P4_MOVE_FLAG_DRAW = 64       a draw is available` from the README.
 
## Further Improvements

- The engine is kind of old, I have added comments to understand what conditions is doing what operations. And it just repeats move so once the kid learns to beat one level he can just repeat the moves to do the same.
- We can use the engine's p4_move2string function to easy our approach in the game. ( I shall perhaps create an issue on this and then work on this). 
- Further tests from the original library can be added if using the same engine.


activities/Chess.activity/js/chessGame.js
```diff
@@ -391,18 +391,24 @@ var ChessGame = {
...
       // game over
-      if (possibleMove.length === 0) {
+      if (possibleMove.length === 0 && (move.flags === 7 || move.flags === 15 || move.flags === 23 || move.flags === 39)) {
+        // console.log("Game won by computer");
         this.game_won = true;
         return;
       }
-      if (1 / possibleMove[2] == 1 / this.state.stalemate_scores[this.state.to_play]) {
+  
+      if (possibleMove.length === 0 && (move.flags === 5 || move.flags === 13 || move.flags === 21 || move.flags === 37 || move.flags === 64)) {
```
llaske, line 404: Not sure to understand this test. Could you explain?

activities/Chess.activity/js/chessGame.js
```diff
@@ -391,18 +391,24 @@ var ChessGame = {
       var start_time = Date.now();
       var possibleMove = this.state.findmove(this.level);
       var delta = Date.now() - start_time;
+
+      // console.log(`Random Move: ${possibleMove}, Delta: ${delta}`);
+
       // game over
-      if (possibleMove.length === 0) {
+      if (possibleMove.length === 0 && (move.flags === 7 || move.flags === 15 || move.flags === 23 || move.flags === 39)) {
```
llaske, line 398: It seems that instead of testing each value you could check `move.flags & 7`

activities/Chess.activity/js/chessGame.js
```diff
@@ -391,18 +391,24 @@ var ChessGame = {
...
+  
+      if (possibleMove.length === 0 && (move.flags === 5 || move.flags === 13 || move.flags === 21 || move.flags === 37 || move.flags === 64)) {
+        // console.log("Match draw due to stalemate");
         this.game_draw = true;
         return;
       }
       var move = this.state.move(possibleMove[0], possibleMove[1]);
 
-      if (!(move.flags & (1 << 0))) {
+      if (!(move.flags & (1))) {
```
llaske, line 411: This line do the same test than before

activities/Chess.activity/js/chessGame.js
```diff
@@ -414,27 +420,31 @@ var ChessGame = {
...
+          // console.log("Draws");
+          //  critical situation which cause the error
+          this.game_draw = true;
+        
           return;
         }
       }
 
-      if (move.flags & (1 << 1) && move.flags & (1 << 2)) {
+      if (move.flags & (2) && move.flags & (4)) {
```
llaske, line 442: This line do the same test than before

activities/Chess.activity/js/chessGame.js
```diff
@@ -414,27 +420,31 @@ var ChessGame = {
...
         this.game_lost = true;
 
-      } else if (move.flags & (1 << 1) && !(move.flags & (1 << 2))) {
+      } else if (move.flags & (2) && !(move.flags & (4))) {
```
llaske, line 445: This line do the same test than before

activities/Chess.activity/js/chessGame.js
```diff
@@ -414,27 +420,31 @@ var ChessGame = {
...
         this.game_check = true;
-      } else if ((!(move.flags & (1 << 1)) && move.flags & (1 << 2)) || move.flags & (1 << 6)) {
+      } else if ((!(move.flags & (2)) && move.flags & (4)) || move.flags & (64)) {
```
llaske, line 447: This line do the same test than before

activities/Chess.activity/js/chessGame.js
```diff
@@ -458,18 +468,21 @@ var ChessGame = {
...
+        // console.log(`Move: ${source}-${target}, Flags: ${move.flags}`);
         // illegal move
         if (move.flags === 0) return 'snapback';
 
-        if (move.flags === 7) {
+        if (move.flags === 7 || move.flags === 15 || move.flags === 23 || move.flags === 39) {
+          console.log("Game won by player");
           this.game_won = true;
-        } else if ((!(move.flags & (1 << 1)) && move.flags & (1 << 2)) || move.flags & (1 << 6)) {
+        } else if ( (!(move.flags & 2 )) && ((move.flags & 4 ) || move.flags & (64))) {
```
llaske, line 478: This line do the same test than before

activities/Chess.activity/js/chessGame.js
```diff
@@ -458,18 +468,21 @@ var ChessGame = {
...
+          // console.log("Match draw due to stalemate by player");
           this.game_draw = true;
         } else {
           this.game_check = false;
         }
 
-        if (move.flags & (1 << 1) && !(move.flags & (1 << 2))) {
+        if (move.flags & (2) && !(move.flags & (4))) {
```
llaske, line 485: This line do the same test than before

activities/Chess.activity/js/chessGame.js
```diff
@@ -414,27 +420,31 @@ var ChessGame = {
...
         }
-        if (possibleMove[2] == this.state.stalemate_scores[this.state.to_play]) {
-          this.game_draw = true;
-          return;
-        }
+        // if (possibleMove[2] === this.state.stalemate_scores[this.state.to_play]) {
+          // console.log("Match draw due to stalemate proper");
+        //   this.game_draw = true;
+        //   return;
+        // }
```
llaske, line 431: Do not keep code into comment

llaske (changes requested):

activities/Chess.activity/js/chessGame.js
```diff
@@ -391,18 +391,24 @@ var ChessGame = {
       var start_time = Date.now();
       var possibleMove = this.state.findmove(this.level);
       var delta = Date.now() - start_time;
+
+      // console.log(`Random Move: ${possibleMove}, Delta: ${delta}`);
+
```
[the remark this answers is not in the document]
MostlyKIGuess: Alright sir I will remove it

activities/Chess.activity/js/chessGame.js
MostlyKIGuess: Yeah no, so basically I changed this but the previous issue was due to the test like this , so for example a valid move has a flag 1, and check has 2 , and if it's mate/statemate it's 4, now the flag we get is the OR of all of them. So keeping 
move.flag & 7 will give 1 on a valid move as well which is incorrect giving false positives for draw. + 8 is for capture, and all the cases are covered here, but I did re write the code a bit more after this too.

activities/Chess.activity/js/chessGame.js
MostlyKIGuess: I removed this line of code, it was not needed is what I realised later

activities/Chess.activity/js/chessGame.js
MostlyKIGuess: I kept numericals because from the library it's easy to read via numbers because they have assigned number to certain things in chess. I could change that if you would say

activities/Chess.activity/js/chessGame.js
MostlyKIGuess: it's not supposed to be condition 1 not true AND condition two OR condition three true. It's supposed to be Condition one not true AND ( condition two or condition 3 is true).

activities/Chess.activity/js/chessGame.js
llaske: Yes please. Don't change what doesn't need changing.

activities/Chess.activity/js/chessGame.js
llaske: Ok got it.

activities/Chess.activity/js/chessGame.js
MostlyKIGuess: I have changed it as it was

MostlyKIGuess: And made the appropriate logic changes to solve the issue.

llaske:
Nice. Thanks.

Issue #1191, opened 2023-02-24
In the activity of food chain, the picture of frog shows incomplete


ybwell:
In the activity of the food chain, the picture of the frog during customs clearance sometimes shows incomplete, and the frequency of occurrence is not low.
![111](https://user-images.githubusercontent.com/87561656/221172305-85c598a6-a8cb-4662-9331-be2d21dc7cac.png)
![222](https://user-images.githubusercontent.com/87561656/221172311-2d922c11-8eff-48b2-827e-795a459ecabe.png)

llaske:
Yes, that's true

llaske:
@abhishektiwari47 there is no need to ask permission to work on an issue. See https://github.com/llaske/sugarizer/blob/dev/CONTRIBUTING.md

abhishektiwari47:
Yes,I got it. That could be my previous question....

Vishalk91-4:
@llaske, I found that the image overlap occurs after we have hit a stone, frog is dead but the background with it gets shadowed over

https://github.com/llaske/sugarizer/assets/116670999/648114fe-ecfb-44af-bc5f-3c549a23ee1a

llaske:
> @llaske, I found that the image overlap occurs after we have hit a stone, frog is dead but the background with it gets shadowed over

Interesting. I guess it's related to drawing in canvas.

Vishalk91-4:
> > @llaske, I found that the image overlap occurs after we have hit a stone, frog is dead but the background with it gets shadowed over
> 
> Interesting. I guess it's related to drawing in canvas.

And another thing than can we reduce the padding present at top and bottom in the frog canvas
![image](https://github.com/llaske/sugarizer/assets/116670999/277fcc0a-b14a-4ca7-80eb-a26a6aba184e)
This extra spaces cause the frog to get hit with the stone even when it is at some distance from it

llaske:
Fixed in https://github.com/llaske/sugarizer/pull/1681

PR #1681, opened 2024-12-22
Picture of frog and rocks is sometimes incomplete in Foodchain activity 

AdityaKrSingh26:
### Fix to issue : #1191 

**Issues:**
- Picutre of frog is sometimes incomplete when overlapping with other images.
- On collision of frog with rock erases some parts of the rock making the image of rock as incomplete.

**Fix:**
Whenever the frog collides with a rocks they rocks are erased from canvas and redrawn.
Whenever the frog and flies overlap frog is redrawn.


llaske:
Nice. Thanks.

Issue #1691, opened 2024-12-29
Pressing the start button in ColorMyWorld activity immediately shows "Congratulations! You finished!" 


AdityaKrSingh26:
In ColorMyWorld activity when you hit the start button immediately when the activity loads it shows the popup of "Congratulations! You finished!" whereas if you wait for some seconds it works normally.
![Screenshot from 2024-12-29 12-53-07](https://github.com/user-attachments/assets/d8913d6f-653b-4af4-b162-8a8f42c14300)

https://github.com/user-attachments/assets/f6231e78-5376-4060-a7a0-c970272bec70

AdityaKrSingh26:
@llaske please have a look at this

llaske:
Fix on https://github.com/llaske/sugarizer/pull/1692

PR #1692, opened 2024-12-29
Add a spinner at start of color my world activity 

AdityaKrSingh26:
**Fix to issue** : #1691

**Issue** : Features take 2 seconds to load and if start button is pressed before that it shows a popup -> "Congratulations! You finished!" 

**Fix** : Add a loader screen for 2 seconds so that the activity loads completely in meantime 

[Screencast from 2024-12-29 13-21-03.webm](https://github.com/user-attachments/assets/c3216116-55e5-4891-ba0f-d90f0968ef4d)


llaske:
Disable the whole screen give a bad experience to users.
Plus the spinner is not the one use in Sugarizer.
Hard coding 2 seconds wait is not a good practice: it could be more or less depending on the device and the network connection.

The better solution will be to detect when the initialization is done and disable the button until it happens.

AdityaKrSingh26:
> Disable the whole screen give a bad experience to users. Plus the spinner is not the one use in Sugarizer. Hard coding 2 seconds wait is not a good practice: it could be more or less depending on the device and the network connection.
> 
> The better solution will be to detect when the initialization is done and disable the button until it happens.

okay i will work on this and fix it

AdityaKrSingh26:
Is this good enough ? 
Screenshot from 2024-12-30 00-08-04

llaske:
Not sure to understand.
You didn't fix the issue, you've just added a spinner.
The idea is to forbid to click on play before initialization is done.

AdityaKrSingh26:
> Not sure to understand. You didn't fix the issue, you've just added a spinner. The idea is to forbid to click on play before initialization is done.

I'm sorry i forgot to push some changes. i have pushed the changes where i am preventing any action on click of run button while activity is loading.


https://github.com/user-attachments/assets/f7f2031c-54f5-4e2a-9ebf-3a5e04f8fdec

llaske:
Nice. Thanks.

Issue #1689, opened 2024-12-29
Fullscreen Button Hides Toolbar Instead of Maximizing to Fullscreen


Devmoni:
### Description:
When the fullscreen button is clicked, it hides the toolbar. The expected behavior is for the app to enter fullscreen mode (using the entire device screen), but the toolbar should remain visible. The fullscreen button should be replaced by the unfullscreen button.

### Expected Behavior:
- The fullscreen button should maximize the app to fullscreen, keeping the toolbar visible.
- The fullscreen button should be replaced by the unfullscreen button.
- When the unfullscreen button is clicked, the app should exit fullscreen and the fullscreen button should reappear.
### Current Behavior:
- The fullscreen button hides the toolbar, which is not the expected behavior.

Devmoni:
As I have just implemented this functionality for a single activity to test the result, I think it needs to be implemented separately for each activity to ensure consistency across the app.
@llaske

llaske:
No. The expected behavior is to hide the toolbar and use all the space need.
It's different from the full screen mode of the browser.
Most of Sugarizer users are using a tablet where the app use already the whole screen. So implementing your full screen mode has no sense for them.

PR #1693, opened 2024-12-29
Fix fullscreen button behavior and toolbar visibility

Devmoni:
Fix : #1689  

### Description:
This PR addresses the issue with the fullscreen button behavior. The fullscreen mode is now implemented to maximize the app without hiding the toolbar. The fullscreen button is replaced by the unfullscreen button when activated.

### Changes:
- Implemented fullscreen functionality while keeping the toolbar visible.
- Added proper toggling of fullscreen and unfullscreen buttons.
### Note:
I have implemented this functionality for the Calculate activity to demonstrate the current behavior. Once approved, this can be extended to other activities for consistency across the app.

### DIFF
--- a/activities/Calculate.activity/css/activity.css
+++ b/activities/Calculate.activity/css/activity.css
@@ -34,22 +34,32 @@
   background-image: url(../icons/digits-6.svg);
   display: none;
 }
-#fullscreen-button {
-  background-image: url(../icons/view-fullscreen.svg);
-}
-#unfullscreen-button {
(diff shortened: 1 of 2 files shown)


llaske:
It's not a bug.
Please do not send a PR before there is a related issue approved. See https://github.com/llaske/sugarizer/blob/master/CONTRIBUTING.md

Devmoni:
Understood, thank you for pointing that out. 
I apologize for any oversight and will make sure there’s an approved issue before submitting a PR in the future.

PR #1698, opened 2025-01-07
Resize images in fototoon

AdityaKrSingh26:
Fix : #1551

Earlier the images were set as background image to the whole canvas as a result they were not resizable. 
In this fix i have made the images as a sperate globe which is resizable.

After fix:
https://github.com/user-attachments/assets/defadbed-dcb0-4ec7-90b4-6d807a7bf45e

### DIFF
--- a/activities/Fototoon.activity/js/activity.js
+++ b/activities/Fototoon.activity/js/activity.js
@@ -109,7 +109,7 @@ define(["sugar-web/activity/activity","sugar-web/datastore","sugar-web/env","tex
 						element.src = text;
 					}
 					element.onload = function () {
-						toonModel.addImage(element.src);
+                        toonModel.addGlobe(toon.TYPE_IMAGE,element.src)
 					};
 				});
 			}, { mimetype: 'image/png' }, { mimetype: 'image/jpeg' }, { activity: 'org.olpcfrance.PaintActivity'});
--- a/activities/Fototoon.activity/js/toon.js
+++ b/activities/Fototoon.activity/js/toon.js
@@ -41,6 +41,8 @@ define(["easel","sugar-web/datastore","sugar-web/env","l10n","humane"], function
     var TYPE_CLOUD = 'CLOUD';
     var TYPE_EXCLAMATION = 'EXCLAMATION';
     var TYPE_RECTANGLE = 'RECTANGLE';
+    var TYPE_IMAGE = 'IMAGE';
+    var imageUrl = null;
 
     // TYPE_WHISPER is saved as TYPE_GLOBE and mode MODE_WHISPER
     var TYPE_WHISPER = 'WHISPER';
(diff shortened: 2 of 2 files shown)


AdityaKrSingh26:
Hey @llaske please have a look at this

llaske:
It's nice but it broke the way of using the activity.
When the image icon is clicked it suppose to add a new page in the comic.
Now it's no longer possible to add page.

AdityaKrSingh26:
> It's nice but it broke the way of using the activity. When the image icon is clicked it suppose to add a new page in the comic. Now it's no longer possible to add page.

okay ..... i was not aware of that feature.
We can do 2 things:
1. either add a button to add new page 
2. or add a new page when next page button is clicked if the current page is the last page.

llaske:
The better option is the third.

AdityaKrSingh26:
> The better option is the third.

Okay i will work on this 
Just one thing i would like to make sure is that going with option 3 we will have only one image per page.

llaske:
> with option 3 we will have only one image per page.

yes, that's right

AdityaKrSingh26:
Hey @llaske i still trying to fix the issue  but i am facing issue while adding a new page when adding a new image could you please guide me with something.

AdityaKrSingh26:
@llaske please have a look at my latest commit. I have added the code for resizing the image and when new image is added a new page is created

Preview:

https://github.com/user-attachments/assets/3c3f16a5-16e1-41e7-9ff5-5cce8e62e11a

llaske:
Good job.
But when opening an existing Fototoon activity with the new version, it shrink images inside.

image

become:

image

AdityaKrSingh26:
My older commit was not keeping the aspect ratio of image when it was first loaded. I have updated it so that the original aspect ratio of image is respected.

@llaske please have a look at my latest commit.

llaske:
I've tested your latest PR, there is a memory leak issue.
I've created a three redimensioned images project with some globes.
After navigating several times from a page to another, images are not displayed or displayed at the wrong place.
It's not happens on the original activity.

[Capture vidéo du 2025-03-30 11-48-53.webm](https://github.com/user-attachments/assets/af3ff89c-85d8-4be5-80a5-971e090a0508)

AdityaKrSingh26:
Got it ..... I'll try to fix this and update the pr

AdityaKrSingh26:
Hey @llaske please have a look at my latest commit. I have added :
- A cleanup method to Clean up previous resources first and ensure when new comic box is loaded previous one is cleaned. 
- Created a completion handler to process any pending changes.

https://github.com/user-attachments/assets/7d96fc7d-d97c-47dd-9512-4b15a2d5f8b3

llaske:
Hmmm it fix the issue but it's much less fluid, the user has to wait 1 second to switch between pages :disappointed:

AdityaKrSingh26:
> Hmmm it fix the issue but it's much less fluid, the user has to wait 1 second to switch between pages 😞

I am having difficulty understanding why this is happening ........... but I'll give it a try

Rejah-Rabeeul:
Hi @lionellaske, I noticed this PR is still in Draft status and hasn't been updated in a while.

I am interested in this feature. If the original author is no longer working on it, may I do this task?

llaske:
@Rejah-Rabeeul yes you can

llaske:
Fixed in https://github.com/llaske/sugarizer/pull/1916

Issue #1695, opened 2024-12-30
Sample values in Chart activity are not initialized


llaske:
At first launch, the Chart activity use sports name to initialize values.
When launch from the app in a default English language, values are not initialized (EgLabelX).

![image](https://github.com/user-attachments/assets/3cbfd603-2301-4db9-bdc4-b7b669adf05f)

Full step to reproduce:

[Capture vidéo du 2024-12-30 18-02-31.webm](https://github.com/user-attachments/assets/022ec1f6-b234-4b92-a762-b3e164e671ff)

AdityaKrSingh26:
The issue is begin caused because the the dictionary is not fully loaded in `SugarL10.js` when the acitivity starts. One thing we can do is add a loader. If it is fine to add a loader @llaske I will raise a pr for this or should i look for other alternatives ?

SaniyaBorkar:
@AdityaKrSingh26 have you tried this loader implementation? If yes, can you just test does it work in firefox? I tried a delay but it isn't working with firefox. If you got the same issue with firefox ig this implementation wouldn't work. I have an alternative fix.

kumar-piyush12:
Hi @llaske , I have started working on this issue. Will update you asap

kumar-piyush12:
Hi @llaske, I have fixed this issue. I am new in Open Source Contribution. I humbly request you to accept my Pull Request.
Thank you
![1](https://github.com/user-attachments/assets/7ccb22c0-f3a8-4f51-8b74-b3d8bfce312e)

SaniyaBorkar:
@llaske Please have a look at this

llaske:
Fixed in https://github.com/llaske/sugarizer/pull/1703

PR #1703, opened 2025-01-12
Initialized sample values in chart activity

SaniyaBorkar:
Fix to issue: #1695 

Issue: The problem occurs because the sample values are not localized if the onJournalNewInstance function is triggered before the localization process. As a result, the function returns default values (e.g., EgLabelX), which are not in the localized format.

Fix: To resolve this issue, the onJournalNewInstance function is triggered after the localization process. This ensures that all sample values are correctly localized before they are used, preventing the default values (e.g., EgLabelX) from being used.

https://github.com/user-attachments/assets/1c0a9bc5-76db-4159-9c37-a041be2b9a78

### DIFF
--- a/activities/Chart.activity/js/activity.js
+++ b/activities/Chart.activity/js/activity.js
@@ -22,6 +22,8 @@ const app = new Vue({
 		"csv-view": CsvView,
 	},
 	data: {
+		isLocalized: false,
+		pendingNewInstance: false,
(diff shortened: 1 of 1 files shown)


llaske:
Nice. Thanks.

SaniyaBorkar:
Happy to contribute!

Issue #1727, opened 2025-02-08
Bug: X and Y axis labels are not Interchanged when switching between Vertical and Horizontal Bar Charts


AliHassan245:
![Image](https://github.com/user-attachments/assets/575c4342-f111-4d04-b0d9-fc723e815ed9)
![Image](https://github.com/user-attachments/assets/23618a9a-cfc6-451c-83d4-3f1064cce86d)

llaske:
That's true, it make sense.

llaske:
Fixed in https://github.com/llaske/sugarizer/commit/c3baee4433ecf4f5bf88fbb8bb6ace4ede8c9c0b

PR #1729, opened 2025-02-08
X and Y axis labels interchanged when switching between Vertical and …

AliHassan245:
X and Y axis labels are interchanged when switching between Vertical and Horizontal Bar Charts 
Issue Number: #1727


llaske:
Nice. Thanks.

Issue #1739, opened 2025-02-16
Bug: BlockRain Activity Game Over Logic Fails When Side Columns Are Full


AliHassan245:
- When the game side columns reach the top, the game does not end. Instead, the game-over state only triggers once the middle column is filled. This behavior is non-standard compared to typical Tetris-like games, which usually end as soon as any column reaches the top boundary.

![Image](https://github.com/user-attachments/assets/e8505e63-b36e-488e-80cd-94e3332dad67)

![Image](https://github.com/user-attachments/assets/b0e8fff6-ebc8-4ca6-8193-0d6c2ecc7ef4)

- Sugarizer also states that it follows standard Tetris rules, where the game should end when any column reaches the top boundary.

![Image](https://github.com/user-attachments/assets/cd3d233f-8831-4fc6-b24d-b46e258313e5)

llaske:
Hmmm. That's true but it comes from the original library.
I doubt it could be solved.

llaske:
It's not a competition we should all be happy to improve open source software.
Fixed in https://github.com/llaske/sugarizer/pull/1741

akhileshdasari2004:
> * Hey [@akhileshdasari2004](https://github.com/akhileshdasari2004) [@AdityaKrSingh26](https://github.com/AdityaKrSingh26) ,thanks for your contribution. I want to point out that that I had already resolved this issue and was waiting for the maintainer [@llaske](https://github.com/llaske) to officially tag it as a bug before submitting my PR. because as per our guidelines (https://github.com/llaske/sugarizer/blob/master/CONTRIBUTING.md), PRs should only be made to approved issues. In the future, it would be helpful if we double-check whether someone has already started work on an issue to avoid duplicate efforts.
> * [@llaske](https://github.com/llaske), as I mentioned earlier, I had already resolved this issue and was waiting for you to officially tag it as a bug before submitting my PR. Please let me know if I should proceed with the PR.
> 
>  blockrain3.mp4

Hey @AliHassan245, got it! Appreciate the clarification.

PR #1741, opened 2025-02-17
Fix issue in Blockrain activiyt game over logic

AdityaKrSingh26:
Fixes : #1739

Fix :  When placing a block Check if block would be placed at or above the top.

After fix demo:

https://github.com/user-attachments/assets/92ad9e49-2802-450c-bdc2-ee8c5358a20c


AdityaKrSingh26:
Hey @llaske please have a look at this

llaske:
Nice. Thanks.

PR #1745, opened 2025-02-18
fixed When the game side columns reach the top, the game does not end

akhileshdasari2004:
Issue Fixed:
Previously, the game-over condition only triggered when the middle column was filled, allowing side columns to exceed the top without ending the game. This was non-standard for Tetris-like games, which should end as soon as any column reaches the top boundary.

Resolution:
Now, the game-over state triggers immediately when any column reaches the top, ensuring proper gameplay mechanics. 🚀

imageimage
image

@llaske I have fixed the indentations. I updated the codebase where changes were needed and ensured the indentations align with the existing code style.

### DIFF
--- a/activities/Blockrain.activity/js/blockrain.jquery.src.js
+++ b/activities/Blockrain.activity/js/blockrain.jquery.src.js
@@ -2,8 +2,7 @@
 
   "use strict";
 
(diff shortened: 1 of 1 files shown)


llaske:
Fixed in https://github.com/llaske/sugarizer/pull/1741

PR #1758, opened 2025-02-27
Fix: Using SVG flags to bring compatibility to Windows

### LINKED ISSUE #1757: Country flag doesn't display in ColorMyWorld on Windows
On Windows countries flags don't appear in popup when there is a click on the country.
The country code is displayed instead.

Image

Works on other OS.

t-aswath:
**Fixed #1757**  

As per the solution proposed in the issue, a `flags` folder has been added containing all flag images in SVG format from [Flagpedia](https://flagpedia.net/download).  

The `flag.js` file has been updated to serve as a mapping between country names and their respective country codes.  

Additionally, the `humane.log` function, which supports rendering HTML, now includes an `` tag. It retrieves the corresponding SVG file name from `flag.js` and renders it in the same format as before: `flag <space> country`.

Total size of the flags svg = ~ 3.7 MB


llaske:
Nice.
Will be better to have a spacing between the flag and the country name.

t-aswath:
Sure sir,
i have left a space `+" "+ ` between the flag and the country name. do you want me to increase it

t-aswath:
> @t-aswath One suggestion I have is to avoid hardcoding the width and height of the country flag. Instead, try adjusting its size relative to the country text for a more aesthetically pleasing look

sure

llaske:
Adding space has no effect in HTML most of the time. You should add two `&nbsp;` instead.
Also the flag should also be replaced in the popup in Tour mode.

image

t-aswath:
> Adding space has no effect in HTML most of the time. You should add two `&nbsp;` instead.
> Also the flag should also be replaced in the popup in Tour mode.
> 
> image
> 

Ya forgot about that part `&nbsp;`

I will fix that plus I will work on the tour part by today

t-aswath:
I have fixed the issues mentioned in the latest commit

llaske:
Nice. Thanks @t-aswath

PR #1759, opened 2025-02-27
Country flag doesn't display in ColorMyWorld on Windows

### LINKED ISSUE #1757: Country flag doesn't display in ColorMyWorld on Windows

AliHassan245:
Fix #1757


llaske:
You didn't respect the rule "Images should be host in the activity. Sugarizer must be able to run without Internet access". 
The @t-aswath PR is better.

PR #1762, opened 2025-03-01
Bug Fixed : Jappy Activity initial render

### LINKED ISSUE #1738: Bug in Jappy activity editor
https://github.com/user-attachments/assets/e213c41b-921b-4add-b710-7d3c1b53130f

The editor doesn't display text when the user types text in the initial render.

SnehalSrivastava27:
Gone through the bug and referred CodeMirror API to solve the issue 
The issue was in proper using of CoderMirror functions
It was using swapDoc instead of setValue , therefore the code was rendering/mirroring on screen when we click on the tab 

Fixes #1738

### DIFF
--- a/activities/Jappy.activity/js/codeeditor.js
+++ b/activities/Jappy.activity/js/codeeditor.js
@@ -1562,7 +1562,7 @@ function init() {
             }
             if (len(window.files) > 0) {
                 tag.title = list(parsed_data)[0];
-                editor.swapDoc((ρσ_expr_temp = window.files)[ρσ_bound_index(tag.title, ρσ_expr_temp)]);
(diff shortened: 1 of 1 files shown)


SnehalSrivastava27:
@Vishvsalvi and @llaske please have a look would love to hear your feedback and improvement if any

Vishvsalvi:
If it fixes the issue it should be merged, I would suggest you to put up a demo video next time for the fixes

SnehalSrivastava27:
> If it fixes the issue it should be merged, I would suggest you to put up a demo video next time for the fixes

Thank you 

https://github.com/user-attachments/assets/642ddc88-dd6b-484a-9770-26c8a6d8f913

llaske:
Nice. Thanks

PR #1767, opened 2025-03-12
Floss fund estabhlishment PR

MostlyKIGuess:
- This also will be needed to done in master branch as that's the directly public branch.
In order for this to validate our schema, we would need to have this structure .
https://github.com/sugarlabs/www/pull/679
As present in the guidelines:
https://floss.fund/funding-manifest/

### DIFF
--- a/.well-known/funding-manifest-urls
+++ b/.well-known/funding-manifest-urls
@@ -1,0 +1,1 @@
+https://www.sugarlabs.org/funding.json


llaske:
Hmmm, we should talk about that. Sugarizer has its one funding tool mentioned on https://sugarizer.org/#websites.

pikurasa:
> Hmmm, we should talk about that. Sugarizer has its one funding tool mentioned on https://sugarizer.org/#websites.

Ideally, Sugar Labs as an org would fundraise for its projects and those funds would be used to pay project expenses. I believe we can raise more funds, and get more work done, if we work together in a coordinated way.

pikurasa:
When I spoke to our accountant about this, for example, she said:

> Similarly, shifting donation streams to Sugar Labs would give contributors and donors a clearer understanding of where their money is going and help avoid confusion around personal vs. organizational funding.
> Once the assets and related expenses are moved under Sugar Labs, it will be easier to establish policies for ongoing costs, equipment needs (like Lionel’s MacBook), and any compensation for project leaders. 
> This approach would also help Sugar Labs present a more unified financial picture for both internal decision-making and external reporting, in my opinion.

pikurasa:
@llaske please let me know what your plan is for this.

If I don't hear back from you tomorrow, I'll just remove the funding link (from [this file](https://github.com/sugarlabs/www/blob/master/funding.json)) for seeking funds through Floss Fund. It can be reintroduced at a later time, but I cannot submit the application until all the projects listed validate.

Also, it does occur to me that we could just add it directly to sugarlabs/sugarizer (downstream, in this case), but that seems unwise to me.

Again, I think we should all cooperate towards finding funding. I believe this is the way to success, that we will achieve more working together.

llaske:
@pikurasa okay I've merged the PR

PR #1769, opened 2025-03-13
Enhancing story narrating experience by introducing audio in story books

### LINKED ISSUE #1666: Enhancing story narrating experience by introducing audio in story books
Add Read-Aloud Functionality for Storybooks

Description: It would be highly beneficial to introduce a "read-aloud" feature that allows children to listen to storybooks being narrated in a natural, human-like voice. This would enhance the user experience, especially for younger audiences who may prefer or require auditory learning. The goal is to provide a more immersive, humanized storytelling experience.

Proposed Implementation:

Text-to-Speech Integration:

We can leverage a Text-to-Speech (TTS) API to convert the story text into speech. These services offer natural-sounding voices that can mimic human narration.

Voice Customization:
Users can choose between different voice types (e.g., gender, accent) to cater to different preferences and languages.
User Interface:

Add a "Read Aloud" button on the storybook interface, which, when clicked, triggers the TTS engine to start narrating the story.
Include basic playback controls (play, pause, stop) for better user control.

Performance Considerations:
Caching the audio output for frequently accessed books to minimize API calls and improve performance.
Ensuring the feature runs smoothly across all supported platforms.
Benefits:

Enhances accessibility for children with visual impairments or reading difficulties.
Provides a comforting, human-like storytelling experience.
Supports auditory learners and adds a layer of engagement to the story-reading process.

AliHassan245:
Fix #1666

### DIFF
--- a/activities/EbookReader.activity/css/activity.css
+++ b/activities/EbookReader.activity/css/activity.css
@@ -359,3 +359,13 @@ body {
 	border: none;
 	outline: none;
 }
+
+#read-play {
+	background-image: url("../icons/read-play.svg");
+	background-size: contain;
+	background-repeat: no-repeat;
+	width: 47px;
+	height: 47px;
+	border: none;
+	outline: none;
+  }
--- a/activities/EbookReader.activity/icons/read-play.svg
+++ b/activities/EbookReader.activity/icons/read-play.svg
@@ -1,0 +1,23 @@
+<svg width="55" height="55" xmlns="http://www.w3.org/2000/svg">
+ <metadata id="metadata18">image/svg+xml</metadata>
+ <g>
+  <title>background</title>
+  <rect fill="none" id="canvas_background" height="514" width="514" y="-1" x="-1"/>
+ </g>
+ <g>
+  <title>Layer 1</title>
+  <g fill="none" stroke="#ffffff" stroke-width="3.194291" stroke-linejoin="round" stroke-miterlimit="4" display="block" id="speaker-100" transform="matrix(1.0957048,0,0,1.0957048,1.0436598,-2.2757779)">
+   <g fill="none" stroke="#ffffff" stroke-width="3.194291" stroke-linejoin="round" stroke-miterlimit="4" id="g4">
+    <g fill="none" stroke="#ffffff" stroke-width="3.194291" stroke-linejoin="round" stroke-miterlimit="4" id="g6">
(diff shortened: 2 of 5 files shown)


llaske:
Good work 👍 
Smart idea to use [SpeechSynthesisUtterance](https://developer.mozilla.org/en-US/docs/Web/API/SpeechSynthesisUtterance) that seems widely supported.
We'll be nice to transform the button into a two state start/stop button to be able to avoid restarting from the beginning.
Also the ink color of the button should be white instead of black to be visible on the toolbar.

llaske:
I've tested on Chromium, it doesn't work. Could you check on your side?

AliHassan245:
@llaske, SirI usually test on Chromium, and everything is working perfectly on my side. Could you please provide more details about the issue or any specific errors you're seeing? That might help me pinpoint any discrepancies between our environments.

[Ebook.webm](https://github.com/user-attachments/assets/7442df3e-ada3-4200-8f3e-ea1f9465a40a)

llaske:
I've got the trace telling that the speech start but there is no sound.

image

Same issue on Chrome on the same PC (Ubuntu 24.04.2 LTS) thought it works on Firefox.

One other thing: it seems that you retrieve the full text and play it. So when clicking on the pause button, it pause only at the end of the text which it's not a good user experience.

nithinshiju-dev:
@llaske I’d like to help continue this PR since the original author seems inactive.
Pausing and resuming from the exact word isn’t possible with the native SpeechSynthesisUtterance API because it doesn’t expose the current spoken offset, even when the text is chunked.
I’m also not able to run it properly in Chromium. Please let me know what changes or improvements you’d like to see added, and if there are any specific requirements for running/testing this via a web server.
Screenshot 2025-12-17 184057

llaske:
@nithinshiju-dev it probably means that this feature was a good idea but that there is no solution to implement it.

nithinshiju-dev:
@llaske Yes, I’ll check if there’s any alternative.

nithinshiju-dev:
Hi @llaske,
I wasn’t able to add my commits directly to this PR due to branch limitations, so I opened a new PR continuing this work.
In the new PR, I improved pause/resume behavior by keeping the full page text in memory and tracking spoken progress via `SpeechSynthesisUtterance.onboundary`, then recreating the utterance from the remaining substring on resume.
Link to the new PR: https://github.com/llaske/sugarizer/pull/1842

llaske:
> I wasn’t able to add my commits directly to this PR due to branch limitations, so I opened a new PR continuing this work. In the new PR

Okay but you should not include in the PR files not related to the fix: all files in directory `lib/sugar-web`

PR #1780, opened 2025-03-22
Incorrect Total Dice Value Subtraction After Shaking 

### LINKED ISSUE #1774: Incorrect Total Dice Value Subtraction After Shaking
When dice (shapes) are shaken before removal, the total dice value becomes miscalculated resulting in a negative total.

Steps to Reproduce:
1. Create more than one dice in the activity.
2. Shake the die.
3. Remove one or more dice from the board when dice is moving(sliding).

https://github.com/user-attachments/assets/73a93cb1-11ae-4685-9183-5382a0f79a50

AliHassan245:
Fix #1774

### DIFF
--- a/activities/3DVolume.activity/js/activity.js
+++ b/activities/3DVolume.activity/js/activity.js
@@ -807,207 +807,88 @@ define([
 				}
 			}
 		}
+            
+		function getScoreForShape(diceEntry, ifRemove) {
+			
+			let shapeType = diceEntry[2];
+			let mesh = diceEntry[0];
+			
+			if (shapeType === "cube") {
+				return getCubeScore(scoresObject, mesh, ifRemove);
+			} else if (shapeType === "deca") {
+				return getDecaScore(scoresObject, mesh, ifRemove);
+			} else if (shapeType === "dodeca") {
+				return getDodecaScore(scoresObject, mesh, ifRemove);
+			} else if (shapeType === "icosa") {
+				return getIcosaScore(scoresObject, mesh, ifRemove);
+			} else if (shapeType === "octa") {
+				return getOctaScore(scoresObject, mesh, ifRemove);
+			} else if (shapeType === "tetra") {
+				return getTetraScore(scoresObject, mesh, ifRemove);
+			}
+			
+			return 0;
+		}
+		
+		function removeShape(diceEntry) {
+			let mesh = diceEntry[0];
+			let body = diceEntry[1];
+			let isNumbered = diceEntry[3]; 
+		
+			if (isNumbered) {
+				
+				let scoreToSubtract = getScoreForShape(diceEntry, true);
+				scoresObject.presentScore -= scoreToSubtract;
+		
+				let scoresArray = scoresObject.lastRoll.split(" + ");
+				let indexToRemove = scoresArray.indexOf(scoreToSubtract.toString());
+				if (indexToRemove !== -1) {
+					scoresArray.splice(indexToRemove, 1);
+				}
+				scoresObject.lastRoll = scoresArray.join(" + ");
+				updateElements();
+			}
+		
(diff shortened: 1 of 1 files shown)


llaske:
@SamarthBagga could you review the PR?

SamarthBagga:
Hi @AliHassan245 ,
Could you please explain your fix ?

AliHassan245:
@SamarthBagga, Here is the explanation:

**Intial Modification:**

- Adding a Physics Engine Sleep State Check in onRemoveClick

Problem: After being shaken, when we remove the shapes (dice volumes) the score total become negative. This is because when shapes are still in motion, their orientation is unstable and can cause misidentifaction of upward face, leading to incorrect value subtracted from score total. Hence resulting in negative score total.

Solution: To fix this, I integrated a motion check into the removal process by using the **physics engine’s sleep state**. In the **onRemoveClick** handler, I added a loop that checks every shape’s physics body to ensure that each one is in the **SLEEPING state**. Removal is allowed to proceed only if all shapes are at rest. This guarantees that the shape being removed has settled into a stable orientation, so the upward face is identified correctly and the proper score is subtracted, resulting in an accurate score total.

**Additional Modification:**

- Creating a Helper Function (getScoreForShape)

Problem: The original code directly called different score functions (e.g., getCubeScore, getDecaScore, etc.) in multiple places, which made it hard to maintain consistency, especially when adapting to the new sleep-check logic.

Solution: The helper function, getScoreForShape, takes a dice entry (which includes the shape type) and a removal flag. It then selects and calls the appropriate score function for that shape. This centralizes the score calculation logic for all shapes (cube, decahedron, dodecahedron, icosahedron, octahedron, and tetrahedron), ensuring consistency, simplifying maintenance, and reducing code duplication.


- Creating a Unified Removal Function (removeShape)

Problem: The original removal logic was split between an old remove() function and parts of the onRemoveClick handler. With the added sleep-check, our removal logic needed to also update the score display (both total score and the individual face values string) in a consistent way.

Solution: I created a new unified removal function, removeShape, that encapsulates all steps of removal:
It checks whether the shape is a numbered (dice) volume.
If it is numbered, it calls getScoreForShape to get the score to subtract, updates the global score, and then manually removes the corresponding substring from the lastRoll display string.
It then removes the shape’s physics body from the world, its mesh from the scene, and deletes the entry from diceArray.

 Here is a video demonstration of the modification:

https://github.com/user-attachments/assets/d916e86a-b7e8-4069-9075-30483553201e

Looking forward to your feedback on my updates!

SamarthBagga:
@AliHassan245
There might be an easier way to do this. The physics body has a property called sleepState. You can put a check in the remove function to check if the body is asleep or not before subtracting the score. Try to do this and tell me if it works. Also don't remove the comments from the code.

AliHassan245:
@SamarthBagga, The remove function approach, while initially appealing, tends to introduce issues with initialization order and fragments the removal logic. This makes it more challenging to maintain and debug. In contrast, using onRemoveClick with helper functions(getScoreForShape & removeShape) centralizes the logic and ensures all dependencies are properly initialized before use, resulting in a more reliable and maintainable solution.

SamarthBagga:
@AliHassan245 , 
Your fix pauses the functionality which isnt ideal. Check the fix in this - [https://github.com/SamarthBagga/sugarizer/tree/fixSubtraction](url). Try to implement it, test it and tell me what problems you face. Ignore all the other changes, it is some linting issue with my prettier. Just check the remove function. There might be some bugs you can fix them, but the fixes shouldnt pause the functionality at all.

SamarthBagga:
@AliHassan245 , did you manage to do it ?

llaske:
Fixed in https://github.com/llaske/sugarizer/pull/1787

PR #1785, opened 2025-04-18
Fix: Incorrect Total Dice Value Subtraction

AdityaKrSingh26:
Fix : #1774 

Issue : 
the activity would:
- subtract the original score values instead of the new values shown on the dice after shaking
- resulting in incorrect scores, sometimes negative values

Fix : 
- Added proper synchronization to ensure dice scores are accurately calculated before removal
- score calculation function called to update the current values
- correct score values are subtracted from the total values

Demo:

https://github.com/user-attachments/assets/46e24b5b-1104-4585-9f3e-c9b2b606ddac

### DIFF
--- a/activities/3DVolume.activity/js/activity.js
+++ b/activities/3DVolume.activity/js/activity.js
@@ -670,13 +670,23 @@ define([
 
 		// Adds all the numbers on top for the numbered volumes.
 		function updateElements() {
-			lastRollElement.textContent =
-				scoresObject.lastRoll.substring(
-					0,
-					scoresObject.lastRoll.length - 2
-				) +
-				"= " +
-				scoresObject.presentScore;
+			if (!scoresObject.lastRoll || scoresObject.lastRoll === "") {
+				lastRollElement.textContent = "";
+				return;
+			}
+			
+			let displayString = scoresObject.lastRoll;
+			// Trim trailing " + " if present
+			if (displayString.endsWith(" + ")) {
+				displayString = displayString.substring(0, displayString.length - 3);
+			}
+			
+			// Only show score if we have valid dice scores
+			if (displayString) {
+				lastRollElement.textContent = displayString + " = " + scoresObject.presentScore;
+			} else {
+				lastRollElement.textContent = "";
+			}
 		}
 
 		const renderer = new THREE.WebGLRenderer({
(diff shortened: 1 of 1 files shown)


AdityaKrSingh26:
Hey @llaske @SamarthBagga please have a look at this

SamarthBagga:
This is not right, look at the original issue and try to replicate that situation. That issue still persists. Have a look at my answer in #1785 and try to implement that.

AdityaKrSingh26:
> This is not right, look at the original issue and try to replicate that situation. That issue still persists. Have a look at my answer in #1785 and try to implement that.

The link you mentioned in : https://github.com/llaske/sugarizer/pull/1780#issuecomment-2760556863 is not working

SamarthBagga:
here is the working link - [link](https://github.com/SamarthBagga/sugarizer/tree/fixSubtraction)

AdityaKrSingh26:
> here is the working link - [link](https://github.com/SamarthBagga/sugarizer/tree/fixSubtraction)

I will go through this 👍

AdityaKrSingh26:
Hey @SamarthBagga @llaske have a look at my updates pr : 

**Fix:** The key change is adding a `shouldSubtractScore` variable that only allows score subtraction when the dice are not moving. This prevents the scores from being subtracted when dice are still in motion, fixing the negative value issue.

**Demo Video :** 

https://github.com/user-attachments/assets/72178f04-4807-43b5-9b3b-fd4f883ae738

SamarthBagga:
In your fix, when you try to remove a volume that isn’t moving while others are, the scores don’t get subtracted. You need to add a check inside each volume to see if it’s moving before subtracting the scores.

AdityaKrSingh26:
> In your fix, when you try to remove a volume that isn’t moving while others are, the scores don’t get subtracted. You need to add a check inside each volume to see if it’s moving before subtracting the scores.

Got it

AdityaKrSingh26:
Hey @SamarthBagga have a look at my latest commit.

Fixes : 
- added helper functions to check individual dice states (`isDiceSleeping()` and `hasAnyScoredDice()`) so the code now only subtracts scores from dice that are both numbered and stationary.
- The fix updates the display logic to completely hide the score when no scored dice remain on the board. preventing unnecessary "= 0" display.

SamarthBagga:
@AdityaKrSingh26 , this is not correct. Try this situation - put two numbered volumes on the board and wait for them to sleep, then put another numbered volume and try to remove one of the sleeping numbered volumes. You will see what is wrong.

llaske:
Fixed in https://github.com/llaske/sugarizer/pull/1787

PR #1787, opened 2025-04-29
fixed dice subtraction issue after shaking 

rahulshendre:
Updated the `remove` function to handle score subtraction when clicking dice. Added two changes in the click-based removal part:
- Set score = 0 to avoid errors if the die type is weird.
- Added sleepState != 0 check so only moving dice (not stable ones) subtract scores when clicked. 

Why sleepState != 0?
We want scores to update only for moving dice when players click them, which is part of the design. Tested it and it works as expected.

Testing
- Clicked moving dice: score updates, `lastRoll` looks good.
- Clicked stable dice: no score change, as intended.
- Tested multiplayer: scores and dice sync fine across players.
- Tried edge cases (I was very rough with the testing, changing the board multiple times, using 20+ volumes, etc): no issues.

I forked it from
https://github.com/SamarthBagga/sugarizer/tree/fixSubtraction
@SamarthBagga had already fixed the issues, I did thorough testing and sent a pr for the same


rahulshendre:
@SamarthBagga, your updated function was working correctly, I tried it thoroughly with multiple corner cases that could arise
Please review the pr.

SamarthBagga:
Why are there so many changes? Do not copy my activity.js from that branch. My prettier config was different and the linting has caused a lot of changes in the file. Only make the relevant changes.

rahulshendre:
fixed it, please check the commit 3 
[fixed the persisiting formatting issue](https://github.com/llaske/sugarizer/pull/1787/commits/93ae1ca9f0deb0f574f67a65bf1c2a861b6fce94)

SamarthBagga:
There are still some unwanted changes.

rahulshendre:
will fix them

SamarthBagga:
There is an issue in this. If we remove all the still volumes whose totals have been calculated and there are still some moving volumes on the board we get a "=0" . Try to update the code so that even that is not shown.

rahulshendre:
i'll look into it

rahulshendre:
1st video shows the recreation of the issue you mentioned

https://github.com/user-attachments/assets/d40d45df-b9aa-4486-bbf5-cb41d67ff15b

The second one shows the fixed version of the same

https://github.com/user-attachments/assets/c650f082-1da4-441c-ab56-0b63c230291e


Before asking you for a code review, I'll explain the simple fix I did-

Added a condition to check if scoresObject.lastRoll is empty
If it's empty, we simply set the display to an empty string ("")
Only if there are actual scores to show (when scoresObject.lastRoll is not empty), we display them with the "= " and score

a simple if-else implementation.
what are your thoughts on this ?

SamarthBagga:
@rahulshendre @llaske this seems fine. although i am not sure why the funding manifest url has been changed.

llaske:
Nice. Thanks @rahulshendre

PR #1788, opened 2025-05-06
fix journal not coloring sometimes

UtkarshSiddhpura:


llaske:
Good job @UtkarshSiddhpura 
Thanks.

PR #1794, opened 2025-05-14
Remove redundant code from feature/humanbody

AdityaKrSingh26:
Fix implemented:

Duplicate loader.load : 
- identical loader.load blocks for loading the skeleton model. One is inside the `env.getEnvironment` callback when and the other is inside the `onNetworkDataReceived `function.
- Solution: Create a single function to load all the model and call it from both places.

Duplicate `env.getEnvironment` calls
- Solution: Call it once and store the environment.

Repetitive Camera Zoom Functions
- The `zoomInFunction`, `zoomOutFunction`, `zoomEqual` Function, and `zoomToFunction` all have very similar structures
- Solution: Create a single zoom function that takes the zoom type and target FOV as parameters

### DIFF
--- a/activities/HumanBody.activity/js/activity.js
+++ b/activities/HumanBody.activity/js/activity.js
@@ -64,7 +64,7 @@ define([
             // Load from datastore
             if (!environment.objectId) {
                 console.log("New instance");
-                loadModel({
+				loadModel({
 					modelPath: "models/skeleton/skeleton.gltf",
 					name: "skeleton",
 					position: { x: 0, y: -5, z: 0 },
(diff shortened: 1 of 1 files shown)


AdityaKrSingh26:
@llaske @SamarthBagga have a look at this

SamarthBagga:
@AdityaKrSingh26 , this seems fine. You can also try to do a similar thing with the Dice Activity.

AdityaKrSingh26:
@llaske @SamarthBagga please have a look at the latest commit

llaske:
There is still indentation issues.
See below:

image

No dots (space) should be use for indentation, only arrows (TAB).

AdityaKrSingh26:
Done. @llaske please have a look once again

llaske:
Nice work. Thanks.

PR #1796, opened 2025-06-02
ADD:  Human Body Model, Organ Model and Layer Toggling mechanism

AdityaKrSingh26:
- Added Layer switching in toolbar and separtion lines
- Changed Background from #000000 (Black) to #1a1a1a
- Changed Mouse Click function for better mesh selection using screen-space testing
- Added Human Body and Human organ system model


llaske:
Good job :+1: 

My remarks:
* [x] Body should be the default model. 
* [x] The model selection palette don't keep track of the selected model (should be selected in the palette). Plus the palette icon change when clicking on it. 
* [x] When switching from a model to another then go back, painted parts should be kept. 
* [x] The zoom in limit should be increase: the user should be able to go closer 
 
On the body model: 
* [x] Legs parts are linked: clicking on a side paint both side, it should not 
* [x] Foots are very dark comparing to other parts 
* [x] Ankle are black 

On the organs model: 
* [x] The organs model seems very huge, the browser is slow when displaying it. It's probably related to its size (40Mb compared to 2Mb for other models). 
* [x] Mouth seems too low comparing to brain and eyes. More globally organs heigh seems not aligned together

llaske:
Few more remarks:

* [x] When the activity is reopen the model palette has not the right selection

[Capture vidéo du 2025-06-17 22-12-55.webm](https://github.com/user-attachments/assets/47eb028d-a519-4023-b4cc-90661f772cf5)

 * [x] The current zoom level and the current camera position should be keep in the context. So when reopening the activity the same view is displayed

AdityaKrSingh26:
> Few more remarks:
> 
> * [ ]  When the activity is reopen the model palette has not the right selection
> 
>  Capture.video.du.2025-06-17.22-12-55.webm 
> * [ ]  The current zoom level and the current camera position should be keep in the context. So when reopening the activity the same view is displayed

@llaske i have fixed them have a look

PR #1799, opened 2025-07-05
[feature/Stickman-activity] 

AdityaKrSingh26:
- [x]  Basic UI Added
image


llaske:
Nice :+1:

AdityaKrSingh26:
- [x]  Speed Controller Add
- [x] Play/Pause Control
- [x] Template select controller Add
- [x] Frame Add
- [x] Frame Preview
- [x] Frame Remove
- [x] Export as video
- [x] Onion Skining

image

llaske:
* [x] The stickman is ugly: long neck, wide shoulders, poorly proportioned ... Take inspiration of stickman [here](https://pivotanimator.net/Download.php) and [here](https://stickfigure-recorder.web.app/).

* [x] It should be possible to move it on the board without changing its parts positions.

[Capture vidéo du 2025-07-10 18-57-36.webm](https://github.com/user-attachments/assets/932fd094-fdfc-4ff4-8b66-c0922134c2d9)

* [x] Changing position of a joint should not impact the whole stickman, only member. See how Pivot animator work https://www.youtube.com/watch?v=srU8ER3UKrA

* [x] Frame height is too small. It should be increase. Could be 1/3 of the activity window.

* [x] Plus button override the frame window so it's impossible to remove the latest frames

image

llaske:
That's better.

* [x] The frame height should still be increase. The minus button is still too small and I think we should add margins around frames
* [x] The stickman should be centered at start and it should be bigger to fill a large part of the space
* [x] The current frame image should be updated when the stickman parts are moved. It's currently updated only when a new frame is created

activities/Stickman.activity/js/activity.js
```diff
@@ -268,7 +268,7 @@ define([
 
 			// Create message
 			const message = document.createElement('p');
-			message.textContent = `Are you sure you want to remove Stickman #${stickmanId}?`;
+			message.textContent = 'Are you sure you want to remove the Stickman ?';
```
llaske: Should be localized

activities/Stickman.activity/js/activity.js
```diff
@@ -373,13 +373,9 @@ define([
 
 			if (stickmen.length <= 1) {
 				minusButton.disabled = true;
-				minusButton.style.opacity = '0.5';
-				minusButton.style.cursor = 'not-allowed';
 				minusButton.title = 'Cannot remove the last stickman';
```
llaske: Should be localized

activities/Stickman.activity/js/activity.js
```diff
@@ -373,13 +373,9 @@ define([
...
 			} else {
 				minusButton.disabled = false;
-				minusButton.style.opacity = '1';
-				minusButton.style.cursor = 'pointer';
 				minusButton.title = 'Remove stickman';
```
llaske: Should be localized

activities/Stickman.activity/js/activity.js
AdityaKrSingh26: Fixed

activities/Stickman.activity/js/activity.js
AdityaKrSingh26: Fixed

activities/Stickman.activity/js/activity.js
AdityaKrSingh26: Fixed

PR #1802, opened 2025-07-19
[feature/human-body] Model Palette Fixes

AdityaKrSingh26:
- [x] Removed unnecessary `console.log()`
- [x] ADD: model change prevention for non-host user but allow model change in paint mode
- [x] Fixed : User A paint a part of the skeleton, user B paint a part of the body. Both are on a different model but when they will switch the model, they will view the painted parts of the other user. User A switching to body will see part colored by B. User B switching to skeleton will see part colored by A.


AdityaKrSingh26:
- [x] Doctor mode is now infinite with no "Game Over"
- [x] Before starting shared tour and shared doctor mode the model is synced to make sure all users remain on same model

AdityaKrSingh26:
- [x] ADD : Save as image
image
image

PR #1805, opened 2025-08-21
Fix #1803: Activity icons lose colors when navigating in restricted mode (Ring View)

### LINKED ISSUE #1803: [v2] Icons colors are lost when switching from spiral to ring view
When switching from Spiral to Ring on the home screen, icons colors are lost and an error occurs in the console log.

https://github.com/user-attachments/assets/66006d9c-b1e7-4c07-82a1-b0f16db8ae24

rahularora27:
fixes #1803 

## Problem
In ring view on small screens, when clicking the "more activities" icon (three dots) to navigate through the restricted activity list, all activity icons would lose their colors. The colors would only return after navigating back to the home screen and clicking an activity.

This issue occurred specifically in ring view, which is activated on smaller screens when there isn't enough space to display all activities in the full spiral layout.

## Root Cause
 The `showNextRestrictedList()` and `showPreviousRestrictedList()` functions were updating `restrictedModeInfo.activities` to show different sets of activities, but **were not calling `computePopup()`** to apply colors to the newly displayed activity icons.

The `computePopup()` function is responsible for applying icon colors based on journal entries, and without this call, the new activity icons would render with their default colors (when the DOM is re-rendered) instead of the personalized colors from previous usage.

## Solution
Added `computePopup()` calls wrapped in `this.$nextTick()` to both the functions. The `$nextTick()` wrapper ensures that:

1. Vue completes the DOM re-rendering after `restrictedModeInfo.activities` changes
2. Vue refs point to the correct new DOM elements
3. `computePopup()` can successfully access and apply colors to the newly rendered activity
  icons

### DIFF
--- a/js/screens/homescreen.js
+++ b/js/screens/homescreen.js
@@ -18,6 +18,7 @@ const HomeScreen = {
 											:ref="'activity'+activity.id"
 											class="home-icon"
 											:svgfile="activity.directory + '/' + activity.icon"
+											:color="activityColors[activity.id]"
 											:x="restrictedModeInfo.positions != undefined ? restrictedModeInfo.positions[index].x : (activityPositions[index] ? activityPositions[index].x : 0)"
 											:y="restrictedModeInfo.positions != undefined ? restrictedModeInfo.positions[index].y : (activityPositions[index] ? activityPositions[index].y : 0)"
 											isNative="true"
@@ -112,6 +113,7 @@ const HomeScreen = {
 			favactivities: [],
 			activities: [],
 			activityPositions: [],
+			activityColors: {},
 			popup: null, // singular popup data
 			username: null,
 			buddycolor: null,
@@ -297,11 +299,7 @@ const HomeScreen = {
 				if (popup.itemList && popup.itemList.length >= 1) {
 					popup.icon.color = iconColor;
 					const iconRef = this.$refs["activity" + popup.id][0];
-					if (iconRef) {
-						iconRef.wait().then(() => {
-							iconRef.colorData = iconColor;
-						})
-					}
+					this.activityColors[activity.id] = iconColor;
 				}
 				popupData[activity.id] = popup;
 			});
@@ -559,17 +557,11 @@ const HomeScreen = {
 
 				this.restrictedModeInfo.start = newStart;
 				this.restrictedModeInfo.activities = activities.slice(this.restrictedModeInfo.start, this.restrictedModeInfo.start + this.restrictedModeInfo.count)
-				this.$nextTick(() => {
-					this.computePopup();
-				});
 				return;
 
 			}
 			this.restrictedModeInfo.start = newStart;
 			this.restrictedModeInfo.activities = activities.slice(this.restrictedModeInfo.start, this.restrictedModeInfo.start + this.restrictedModeInfo.count - 2)
-			this.$nextTick(() => {
-				this.computePopup();
-			});
 		},
 
 		showPreviousRestrictedList() {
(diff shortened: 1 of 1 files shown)


llaske:
@UtkarshSiddhpura could you review the PR?

UtkarshSiddhpura:
@rahularora27 It works but i think it could be improved..

Computing the whole popup data on any change in RestrictedList is redundant.. as it already gets computed for all activities inside `draw` method.

Seems like this part only colors the icons currently rendered on the screen, since $refs only gives access to elements that exist in the DOM.
https://github.com/llaske/sugarizer/blob/dev/js/screens/homescreen.js#L299C1-L306C1

That's why we also have an `if` here as ref could be null for activity icons not in view.
https://github.com/llaske/sugarizer/blob/dev/js/screens/homescreen.js#L300

A solution you could try is to have a state value for the icon color.. so it renders with default color and on change of that state value it'll change the icon color e.g `:color= "activity.color"` in this part of the code
https://github.com/llaske/sugarizer/blob/dev/js/screens/homescreen.js#L15C1-L28C1

Try it and ping me here.. if you face any issues with this approach.

UtkarshSiddhpura (changes requested):
Try the approach mentioned in https://github.com/llaske/sugarizer/pull/1805#issuecomment-3216825279

rahularora27:
@UtkarshSiddhpura Thanks for the feedback. I have implemented the reactive approach as suggested.

I initially tried adding colors directly to activity objects with activity.color, but Vue was not detecting the changes properly (reactivity issue). While this.$set() should have forced reactivity, it caused the activity icons to disappear entirely (bug).

So I have added an activityColors object instead, since there was already a different object for activityPositions. This works well - the colors are displayed correctly when navigating and we avoid computing the entire popup data on every change (navigation).

UtkarshSiddhpura (approved):
@llaske LGTM... you can check on your side.

llaske:
Doesn't seems to work.
Does it work on your side?

[Capture vidéo du 2025-08-28 21-40-23.webm](https://github.com/user-attachments/assets/9b1b45aa-870c-46b4-bb21-ce42d2fbd105)

rahularora27:
Works well for me. Could you explain how to reproduce this?

Uploading Screen Recording 2025-08-29 at 11.38.06 AM.mov…

rahularora27:
Works well for me. Could you explain how to reproduce this?

https://github.com/user-attachments/assets/b8bb1c6b-c79b-44a2-b818-00f9cf19624f

llaske:
Hmmm works well with a new user in a "standard" environment.
In my case, I'm testing new activities so there is in my Journal some entries without existing activity related.
To reproduce: launch an activity, said Gridpaint, close it.
Then remove the Gridpaint activity directory and relaunch the Sugarizer Server.
Then try again the test you've done.

SamarthBagga:
@rahularora27 were you able to reproduce it?

rahularora27:
It works fine for me in both cases. In the standard environment everything runs as expected and even if some activity folders are removed (like after removing Gridpaint), it still works. It just shows an empty spot where the activity icon should be but nothing breaks.


https://github.com/user-attachments/assets/6bfb9377-278e-41ea-9dbb-d9071e16b825


https://github.com/user-attachments/assets/e0afa1cb-781d-4820-8d0c-5bed590b1adb

llaske:
Okay. I think the error regarding missing activity is not related, will open another issue.
Merge in dev branch on https://github.com/llaske/sugarizer/commit/b0fd195d95ff67adbd2806eab8c1db66d110bb95
Thanks.

llaske:
For information, a fallback icon has been added in a PR [here](https://github.com/llaske/sugarizer/commit/85e18feb86e0aff78299e4e87d0f66fed67577a3).
So no error should appear when unknown activity or unknown journal entry is load.

PR #1811, opened 2025-09-15
CSV view now updating real-time withou leaving/saving activity

### LINKED ISSUE #1809: CSV View data not updated in chart activity without saving
https://github.com/user-attachments/assets/e16eb778-1c90-4018-8ad1-3c818f8c8c00

Please refer to the video above however once you quit the activity and reopen, then the rows are visible

SnehalSrivastava27:
Fixed  #1809 
@llaske please review

### DIFF
--- a/activities/Chart.activity/js/activity.js
+++ b/activities/Chart.activity/js/activity.js
@@ -295,7 +295,9 @@ const app = new Vue({
 			this.updateActivityTitle(this.l10n.stringChartActivity)
 
 			const header = [this.pref.labels.x, this.pref.labels.y];
-			this.$refs.csvView.updateJsonData(this.tabularData, header, true);
+			if (this.$refs.csvView) {
+				this.$refs.csvView.updateJsonData(this.tabularData, header, true);
+			}
 		},
 		onJournalDataLoaded(data, metadata) {
 			this.LZ = this.SugarJournal.LZString;
@@ -304,7 +306,9 @@ const app = new Vue({
 				this.csvTitle = this.currentenv.activityName;
 				this.pref.chartType = "csvMode";
 				const json = CSVParser.csvToJson(data);
-				this.$refs.csvView.updateJsonData(json.data, json.headers);
+				if (this.$refs.csvView) {
+					this.$refs.csvView.updateJsonData(json.data, json.headers);
+				}
 				return;
 			}
 			const parsedData = JSON.parse(this.LZ.decompressFromUTF16(data));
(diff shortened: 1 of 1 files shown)


rahularora27:
@SnehalSrivastava27 Your solution works, but I think a few areas could be improved:

When adding or removing rows, both shouldUpdateCsv and tabularData watchers fire, causing the CSV to update twice. The second update is redundant.

The tabularData watcher fires on every keystroke when users type in input fields. Since the CSV view is read only (users can't edit data inside the table directly), these real-time updates are not necessary imo.

When using this.$refs.csvView, it's a good habit to check if it exists first to prevent potential runtime errors:

    if (this.$refs.csvView) {
        this.$refs.csvView.updateJsonData(this.tabularData, this.csvHeader, true);
    }
    
Alternative Approach:
Since CSV view is only visible when pref.chartType === "csvMode", you might want to update the CSV only when the user switches to that view, rather than continuously updating a hidden component.

Hope this helps and ping me if you need any help.

rahularora27:
@llaske 

If users can't edit data in CSV view, should we also disable the add/remove row buttons? Seems inconsistent to allow row changes but not data changes.

llaske:
> Since CSV view is only visible when pref.chartType === "csvMode", you might want to update the CSV only when the user switches to that view, rather than continuously updating a hidden component.

This approach seems to be the best.

SnehalSrivastava27:
> @SnehalSrivastava27 Your solution works, but I think a few areas could be improved:
> 
> When adding or removing rows, both shouldUpdateCsv and tabularData watchers fire, causing the CSV to update twice. The second update is redundant.
> 
> The tabularData watcher fires on every keystroke when users type in input fields. Since the CSV view is read only (users can't edit data inside the table directly), these real-time updates are not necessary imo.
> 
> When using this.$refs.csvView, it's a good habit to check if it exists first to prevent potential runtime errors:
> 
> ```
> if (this.$refs.csvView) {
>     this.$refs.csvView.updateJsonData(this.tabularData, this.csvHeader, true);
> }
> ```
> 
> Alternative Approach: Since CSV view is only visible when pref.chartType === "csvMode", you might want to update the CSV only when the user switches to that view, rather than continuously updating a hidden component.
> 
> Hope this helps and ping me if you need any help.



Thankyou will implement that way

SnehalSrivastava27:
@rahularora27 Fixed the runtime issue as well now
@llaske

Both of you can see now!
Also thankyou rahul for your suggestions

llaske:
Nice. Thanks @SnehalSrivastava27

PR #1812, opened 2025-09-23
Closes #838: add undo/redo buttons and shortcuts for gears activity

### LINKED ISSUE #838: Undo button for Gears activity

rahularora27:
closes #838 

- added undo/redo buttons to the gears activity toolbar
- history manager to track user edits
- redo stack clears on new edits
- keyboard shortcuts
- works only for the current loaded board and is not persisted to the journal

### DIFF
--- a/activities/Gears.activity/css/activity.css
+++ b/activities/Gears.activity/css/activity.css
@@ -35,6 +35,14 @@ body {
   background-image: url(../icons/edit-clear.svg);
 }
 
+#main-toolbar #undo-button {
+  background-image: url(../icons/edit-undo.svg);
+}
+
+#main-toolbar #redo-button {
+  background-image: url(../icons/edit-redo.svg);
+}
+
 #main-toolbar #help-button {
   background-image: url(../icons/help.svg);
 }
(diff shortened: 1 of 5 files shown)


llaske:
Nice work.

Two remarks:
- It will be nice to allow undo/redo removing items

[Capture vidéo du 2025-09-25 22-06-56.webm](https://github.com/user-attachments/assets/7686c1e1-237c-4985-9220-4d3ba7745e57)

- Add a vertical bar on the toolbar before the remove button

rahularora27:
> * It will be nice to allow undo/redo removing items

This actually already works fine on my end. Could you check again if the issue still occurs?

https://github.com/user-attachments/assets/252ad129-bf11-4f6d-973f-626159ceb57e

llaske:
Hmmm strange: undo/redo works when you open a Journal entry but doesn't work when open a new instance.


https://github.com/user-attachments/assets/ac9ac4f3-f0c9-440f-853e-2c4417c19edc

rahularora27:
Thank you for pointing this out. I have pushed a fix, let me know if it works as expected.

llaske:
Good job. Thanks @rahularora27

PR #1813, opened 2025-10-06
[Stickman Activity] ADD: Zoom Pallete

AdityaKrSingh26:
Added zoom pallete to zoom in / zoom out / fit to screen
image


llaske:
Very good. And with the zoom it's very usable on tablet.
Just an issue with lower zoom (see below).
Will be nice too to keep the zoom level when the activity is saved.

https://github.com/user-attachments/assets/aebe68bb-7edc-4e0d-a5ce-b87c91afa2cc

AdityaKrSingh26:
> Very good. And with the zoom it's very usable on tablet. Just an issue with lower zoom (see below). Will be nice too to keep the zoom level when the activity is saved.
> 
>  Capture.video.2025-10-07.21.23.50.mp4

Fixed. @llaske have a look at latest commit

llaske:
Very good. Thanks

PR #1827, opened 2025-12-05
Add Global Time selection palette to Clock activity

### LINKED ISSUE #1825: Add "Global Time" feature to Clock activity with timezone selection
Currently the Clock activity only shows the local device time. It would be useful to add a “Global Time” button that lets users choose a city or country (like UTC, London, New York, Tokyo, etc.), and then the clock updates to show the selected timezone’s current time. This could help learners better understand how time differs around the world.

Image

nithinshiju-dev:
Summary

This PR adds a new “Global Time” feature to the Clock activity.
Learners can choose a city/timezone from a Sugar palette and the clock updates to display the current time for the selected location.

A new Global Time toolbar button opens the palette, which lists available timezones.  
When “Local Time” is selected, the clock behaves as before using the device time.  
For other selections, the time is computed using standard IANA timezone support.

The selected timezone is displayed next to the time when it is non-local, all labels are localized, and the user’s choice is saved to the Journal so it is restored when reopening the activity.

Fixes #1825 
Screenshot 2025-12-05 054525


nithinshiju-dev:
Hi @llaske, thank you for the feedback.

I’ve updated the implementation based on your suggestions:
– Reworked the UI to use a Sugar palette instead of a combobox  
– Localized all strings  
– Displayed the selected timezone name when non-local  
– Persisted the selected timezone in the Journal  
– Recreated the PR targeting the dev branch as requested

Could you please take another look when you have time? Thanks!

PR #1829, opened 2025-12-11
Remove requirejs import from main.js dependancies

CraftierTen5:
The import was not referenced anywhere in the activity. Removing it reduces unnecessary dependencies. The activity was tested locally using the standard development workflow. All functionality works as expected without the requirejs import.


llaske:
PR should be done only on the dev branch (see https://github.com/llaske/sugarizer/blob/master/CONTRIBUTING.md).
This dependency is no longer in the code.

PR #1837, opened 2025-12-19
Declare browserName variable locally to avoid implicit global

CraftierTen5:
getBrowserVersion() assigns to browserName in multiple execution paths without declaring it. In non-strict JavaScript, this creates an implicit global variable, which can leak state across modules and cause hard-to-trace bugs.

This commit explicitly declares browserName to contain its scope within the function. No logic, control flow, or return values are changed.

### DIFF
--- a/js/sugarizer.js
+++ b/js/sugarizer.js
@@ -114,6 +114,7 @@ let sugarizer = {
 		let browserAgent = navigator.userAgent;
 		let browserVersion = '' + parseFloat(navigator.appVersion);
 		let browserMajorVersion = parseInt(navigator.appVersion, 10);
+		let browserName = ""; // declare locally
 		let Offset, OffsetVersion, ix;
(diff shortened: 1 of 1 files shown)


llaske:
Hmmm. It seems that this variable is unused neither locally, neither globally.
I think we could just remove it.

CraftierTen5:
Thanks for the clarification. I’ve updated the PR to remove all unused browserName assignments from getBrowserVersion while keeping the existing parsing logic unchanged.

llaske:
browserMajorVersion variable is also unused. Will be nice to remove it as well.

CraftierTen5:
Thanks for pointing this out. I’ve removed the unused browserMajorVersion variable. I also checked the file for additional dead code, and everything looks fine now.

llaske:
Nice. Thanks.

PR #1842, opened 2025-12-25
Improve EbookReader TTS pause/resume behavior.

nithinshiju-dev:
continuation of #1769

### DIFF
--- a/activities/EbookReader.activity/js/activity.js
+++ b/activities/EbookReader.activity/js/activity.js
@@ -15,7 +15,12 @@ var app = new Vue({
 		currentEpub: null,
 		currentView: LibraryViewer,
 		currentLibrary: {database: []},
-		timer: null
+		timer: null,
+		isSpeaking: false,
+		isPaused: false,
+	    utterance: null,
+		originalText: null,
+		currentCharIndex: 0
 	},
 
 	created: function() {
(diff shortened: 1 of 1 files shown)


llaske:
You should not include in the PR files not related to the fix: all files in directory `lib/sugar-web`

nithinshiju-dev:
> You should not include in the PR files not related to the fix: all files in directory `lib/sugar-web`

@llaske Thanks for the clarification.
I’ve updated the PR to include only the EbookReader TTS changes in activity.js and removed all unrelated files.

llaske:
Your PR is missing toolbar and icons. Did you test it on your side?
Please refresh your environment and ensure the PR contains everything need.

nithinshiju-dev:
Hi @llaske,
Initially, I included the full set of changes so the feature worked end-to-end, but some unrelated files were also included. When you pointed that out, I understood it as a request to remove the extra, unrelated files, so I trimmed the PR down. In doing so, I ended up keeping only the files I directly edited, which removed required pieces like the toolbar and icons.

At that time, the pause/resume logic was working correctly in Electron, but I wasn’t yet able to reliably run Sugarizer in Chromium/web mode. I’ve since set up Sugarizer Server properly and can now test in web mode. While testing, I’m seeing that the voice is not audible in Chromium, so I’d like to fix that and revalidate everything together.

I’ll address this properly and open a new PR once the full feature is tested end-to-end. Thanks for your guidance and patience.

PR #1845, opened 2025-12-26
Remove no-op assertions from leaveNetwork tests

CraftierTen5:
The removed assertions always evaluated to true and could never fail, so they did not verify any actual behavior or contribute to the test’s reliability.

### DIFF
--- a/test/index.html
+++ b/test/index.html
@@ -3,54 +3,145 @@
     <meta charset="utf-8"/>
     <title>Sugarizer Tests</title>
     <link href="https://cdn.rawgit.com/mochajs/mocha/2.3.0/mocha.css" rel="stylesheet" />
+    <script src="../js/sugarizer.js"></script>
   </head>
-  <body>
-  <div id="mocha"></div>
+<body>
+<div id="mocha"></div>
+<script src="https://unpkg.com/mocha@9.2.0/mocha.js"></script>
+<script src="https://unpkg.com/chai@4.3.6/chai.js"></script>
+<script>
+    mocha.setup('bdd');
+    var expect = chai.expect;
+    window.expect = expect; 
+</script>
+<script src="../js/presence.js"></script>
+<script>mocha.setup('bdd')</script>
+<script src="datastore.js"></script>
+<script src="presence.js"></script>
+  
+<script>
+    window.localStorage.clear();
+    window.initSugarizer = function(callback) {
+        window.localStorage.setItem("sugar_settings", JSON.stringify({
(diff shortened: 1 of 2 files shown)


llaske:
Look like the test page no longer work. Did you try on your side before sending the PR?
The test page should be fixed before trying to optimize it.

CraftierTen5:
I checked the test page locally.

The presence and datastore tests rely on RequireJS bootstrapping (e.g. requirejs(["sugar-web/datastore", "sugar-web/presence"], ...)). These modules are AMD modules and are not exposed as plain global variables, so loading them via <script> tags results in ReferenceError before any test logic is executed.

In addition, the test setup assumes a Sugarizer runtime environment (initialization and settings in localStorage, and for some tests a presence server). Without this environment, the test page fails.

llaske:
It has no sense to suggest an improvement to a non running page. How you can test your fix?

CraftierTen5:
That was a mistake on my part. I assumed that the page was running correctly in the Sugarizer environment, so I tested it as a standalone page. I am currently working on fixing the page, if that's okay with you.

llaske:
Sure

CraftierTen5:
I have fixed index.html so tests now run in the Sugarizer environment. 11/46 tests pass (was 0 before). Remaining 35 failures are legitimate Sugarizer bugs requiring datastore/presence implementation
Screenshot 2026-01-03 112913
.

llaske:
Oooch you've misunderstanding the issue.
The problem is just related to missing libraries. I've fixed it in https://github.com/llaske/sugarizer/commit/52067c6f643de51dbb67c65a2b2437278f23764d. Try on your side.
BTW regarding your initial suggestion I'm not in favor of removing assertions on leaveNetwork test. It's a good way to understand that the call is async.

CraftierTen5:
I see. I tested the fix 52067c6 on my system (opened index.html with live server), and it just shows a blank page. 
image

llaske:
Do you have a running instance on Sugarizer Server (https://github.com/llaske/sugarizer/tree/dev) on port 5500?

CraftierTen5:
Yes, sorry, I had missed that step. I tested running the index.html page alongside the Sugarizer server on port 5500. The page shows several 404 errors for chai.js, which is referenced from lib/chai.js but appears to be missing. Installing Chai via npm would place it in node_modules rather than lib, so the paths in the test page wouldn’t match. Is chai.js expected to be included in the Sugarizer environment, or is this a setup issue on my side?
image

PR #1849, opened 2025-12-27
Improve EbookReader TTS pause/resume behavior

nithinshiju-dev:
This PR improves the pause and resume behavior of the existing EbookReader text-to-speech feature.
Narration now pauses immediately and resumes from the exact position where it was paused, using the native Web Speech API.

Tested in Firefox and Microsoft Edge.

https://github.com/user-attachments/assets/d1a8ef7e-b2d6-42c4-8ce5-2976a4c1a386

### DIFF
--- a/activities/EbookReader.activity/css/activity.css
+++ b/activities/EbookReader.activity/css/activity.css
@@ -349,3 +349,13 @@ body {
 	box-shadow: 0 0 0 0rem rgba(158, 158, 158, .5);
 	border: 0px;
 }
+
+#read-button {
+	background-image: url("../icons/read.svg"); 
+	background-size: contain;
+	background-repeat: no-repeat;
+	width: 47px;  
+	height: 47px; 
+	border: none;
+	outline: none;
+}
--- a/activities/EbookReader.activity/css/activity.css
+++ b/activities/EbookReader.activity/css/activity.css
@@ -359,3 +359,13 @@ body {
 	border: none;
 	outline: none;
 }
+
+#read-play {
+	background-image: url("../icons/read-play.svg");
+	background-size: contain;
+	background-repeat: no-repeat;
+	width: 47px;
+	height: 47px;
+	border: none;
+	outline: none;
+  }
--- a/activities/EbookReader.activity/icons/read-play.svg
+++ b/activities/EbookReader.activity/icons/read-play.svg
@@ -1,0 +1,23 @@
+<svg width="55" height="55" xmlns="http://www.w3.org/2000/svg">
+ <metadata id="metadata18">image/svg+xml</metadata>
+ <g>
+  <title>background</title>
+  <rect fill="none" id="canvas_background" height="514" width="514" y="-1" x="-1"/>
+ </g>
+ <g>
+  <title>Layer 1</title>
+  <g fill="none" stroke="#ffffff" stroke-width="3.194291" stroke-linejoin="round" stroke-miterlimit="4" display="block" id="speaker-100" transform="matrix(1.0957048,0,0,1.0957048,1.0436598,-2.2757779)">
+   <g fill="none" stroke="#ffffff" stroke-width="3.194291" stroke-linejoin="round" stroke-miterlimit="4" id="g4">
+    <g fill="none" stroke="#ffffff" stroke-width="3.194291" stroke-linejoin="round" stroke-miterlimit="4" id="g6">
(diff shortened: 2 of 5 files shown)


nithinshiju-dev:
Hi @llaske,

This PR improves the pause/resume behavior of the existing EbookReader text-to-speech feature. Instead of relying only on speechSynthesis.pause() and resume(), it now tracks spoken progress using SpeechSynthesisUtterance.onboundary events and recreates the utterance from the remaining text when resuming. This allows narration to pause immediately and continue from the exact position where it was stopped, rather than restarting or waiting until the end.

I tested this across browsers:

Firefox and Microsoft Edge work as expected.

In Chrome/Chromium, speech starts but audio is not audible. From testing and related discussions, this appears to be a limitation in Chrome’s current SpeechSynthesis implementation rather than an issue with the pause/resume logic itself, so hopefully this will be addressed in future Chromium updates.

One known limitation with the current approach is page and chapter navigation. For example, if narration is active on page 1 and the user moves to page 2, narration continues reading the text from page 1 instead of stopping and starting on page 2. Similarly, if narration is paused on one chapter and the user navigates to another chapter, resuming continues from the previous position rather than the new chapter. These behaviors are unchanged by this PR and would require explicitly resetting or recalculating the narration state on navigation events.

There are also some possible follow-up improvements that could be explored later, such as adding speech rate control, supporting multiple languages or allowing users to choose where narration starts. I’m not sure which of these would be feasible with the current APIs, but I wanted to mention them as ideas.

I’ll have holidays from Jan 18 to Feb, so if this direction looks useful, I’d be happy to work on these improvements during that time.

Thanks for taking a look.

llaske:
Thanks for this analysis @nithinshiju-dev .
BTW I'm not convinced by the result. Due to the lack of compatibility with Chrome/Chromium and the poor quality on other platforms, this feature is unusable.
It's not related to your PR, it was the same on https://github.com/llaske/sugarizer/pull/1769.
I don't think we need to spend more time on this issue, it probably means that it's just impossible from now so I'm going to close the issue for the moment.

nithinshiju-dev:
ohh ok @llaske, Just one quick doubt from my side
Is it possible to package or integrate Python-based logic into Sugarizer, or is it strictly limited to JavaScript/HTML activities?

llaske:
> Is it possible to package or integrate Python-based logic into Sugarizer, or is it strictly limited to JavaScript/HTML activities?

Because of its [architecture](https://github.com/llaske/sugarizer/blob/dev/docs/architecture.md), Sugarizer support only HTML/JavaScript activities.

nithinshiju-dev:
@llaske  just one last question from my side before fully letting this go. I’m asking mainly because I personally really like the idea of a read-aloud feature and wanted to be sure before stopping.

I came across an alternative approach based on Piper TTS (browser-based using WASM/ONNX). It runs the TTS engine directly in the browser, converting the visible page text to audio locally without relying on the native speechSynthesis API. This would require a one-time initial download of a voice model (~10–40 MB), which gets cached locally (e.g., via IndexedDB) and then works fully offline afterward. This approach seems to avoid the Chrome/Chromium issues seen with the native API.

If this kind of initial download and additional dependency would be acceptable in principle for Sugarizer, I could try exploring it one last time. If not, I’m completely fine dropping the topic here.

Thanks again for the guidance and your time!

nithinshiju-dev:
Alternatively, we could also reuse meSpeak, which is already used in the Speak activity, for implementing read-aloud in EbookReader as well since this already works reliably in Chrome. The main drawback, as seen in Speak, is that the voice quality is quite robotic.

llaske:
> I came across an alternative approach based on Piper TTS (browser-based using WASM/ONNX). It runs the TTS engine directly in the browser, converting the visible page text to audio local. This would require a one-time initial download of a voice model (~10–40 MB), which gets cached locally (e.g., via IndexedDB) and then works fully offline afterward

Ooch sounds too complex/heavy to use WASM.

> The main drawback, as seen in Speak, is that the voice quality is quite robotic

Yes, Speak is nice to have fun but is not a solution for a reading activity.

nithinshiju-dev:
Sorry to message again @llaske.
Sugar already uses Kokoro in the Speak activity and it’s quite lightweight (around ~10 MB). Is Kokoro considered an accepted dependency?

I know Sugar is an OS itself, so the constraints are different. I thought the read-aloud feature would be nice, but I understand why it’s not possible right now.

PR #1866, opened 2026-01-01
EbookReader Add read-aloud feature with voice selection

### LINKED ISSUE #1666: Enhancing story narrating experience by introducing audio in story books

nithinshiju-dev:
This PR adds a basic read-aloud (text-to-speech) feature to the EbookReader activity.

Users can listen to the current page, pause/resume playback, and choose a voice.
The selected voice is saved and restored from the Journal.

Fixes #1666 
https://github.com/user-attachments/assets/864d6187-872c-4a77-a011-bc38bed42721

### DIFF
--- a/activities/EbookReader.activity/css/activity.css
+++ b/activities/EbookReader.activity/css/activity.css
@@ -102,14 +102,9 @@
 	box-sizing: initial;
 }
 
-.palette *, ::after, ::before {
-	box-sizing: initial;
-}
-
 body {
 	overflow-x: hidden;
 	overflow-y: hidden;
-	height: 100vh;
 }
 
 #canvas {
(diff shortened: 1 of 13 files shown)


nithinshiju-dev:
Hi @llaske, happy new year! 

Just to add some context:

Initially, audio was not playing in Chrome, while it worked in other browsers. After adjusting how speech playback is initialized and handled, audio now works correctly across Chrome, Edge, and Firefox.

One of the harder parts was extracting only the visible text from the current page. I tried a few approaches and was close to switching to Foliate at one point. While debugging this, I came across this epub.js issue:
https://github.com/futurepress/epub.js/issues/1376

That helped resolve the visible text extraction problem without needing to switch libraries.

I also added a simple voice option (male/female) for flexibility.

Please let me know if you have any suggestions or changes you’d like!

nithinshiju-dev:
One small note on possible future improvements :

I had briefly experimented with an accent option (different English accents) just for exploration and fun, to see how pronunciation varies across countries, but I’ve removed it for now. I can add it later if that’s something you’d be open to.

Other potential follow-ups could be playback speed control, allowing users to choose where narration starts from within the text, and highlighting words as they are spoken during read-aloud.

nithinshiju-dev:
Hi @llaske,

While testing the read-aloud feature across browsers, I noticed that pause/resume works as expected in Chrome and Edge, but in Firefox, resuming after pause restarts playback from the beginning.

This seems to be caused by Firefox’s incomplete support for SpeechSynthesisUtterance.onboundary, where charIndex is unreliable, making mid-utterance resume inaccurate.

My plan is to add a Firefox-specific speech handling instance that avoids mid-utterance resume and instead resumes cleanly from the next chunk, while keeping the current exact-resume behavior for Chromium-based browsers.

I currently have exams until Jan 16, so I’ll push this Firefox-specific fix right after that. Just wanted to flag the issue early.

Thanks.

llaske:
It doesn't work on my side in Chromium and it's definitively unusable on Firefox (robot voice).
Plus there is lot of restriction on this API https://caniuse.com/speech-synthesis
So I'm still not at all convinced by the interest of this feature.

nithinshiju-dev:
Ok, I understand. @llaske  Thanks for taking the time to explain the limitations and test it.

Given the browser restrictions and inconsistent behavior, I agree that this feature may not be suitable right now.

If you have any suggestions for improvements or new features in any Sugarizer activity, I’d be happy to work on something you recommend.

Issue #1863, opened 2025-12-31
[Feat] Adding mass property to objects and float sink behaviour in water.


HarleenKaurBhullar:
Currently, the Physics activity in Sugarizer mainly demonstrates gravity through object movement. I’d love to extend this by adding the concept of mass to the objects that users can control.

Along with mass, I’m thinking of adding a water option where users can drop objects and see how they behave. Depending on their mass, objects would either sink faster, sink slowly, or float, helping users understand how mass affects motion in water.

This would make the activity more interactive and help learners explore:

- How mass affects object behavior

- The difference between movement in air vs water

- Basic ideas related to buoyancy, in a simple visual way

Let me know what you think of this proposal. I'd be happy to work on this.

llaske:
Could be a good idea. Let's see if you can do something fun.

HarleenKaurBhullar:
Thanks @llaske will try my best to make a PR  with above mentioned changes as soon as possible.

karthik-dev56:
Hey @llaske If you are ok I can work on this issue also

llaske:
Look like @HarleenKaurBhullar already work on it. May be you could help him if need.

karthik-dev56:
Hey @HarleenKaurBhullar  If y are working on this issue still I will help out u or else I will fix it

HarleenKaurBhullar:
@karthik-dev56 Yes I am already working on it but if you are interested maybe we can work together.

karthik-dev56:
Hey @HarleenKaurBhullar I would love to work

HarleenKaurBhullar:
Ok @karthik-dev56 I have made some progress and was planning to open a pr today to get reviews from llaske, would love your feedback as well and then depending upon it we can work ahead.

karthik-dev56:
Sure @HarleenKaurBhullar happy to assist you and work with you

HarleenKaurBhullar:
@karthik-dev56 I have just opened a PR with changes I had made till now, you can check it out.

HarleenKaurBhullar:
Hi @karthik-dev56 if you dont mind can you help me with a thing? Despite maintainer's suggestion I am not able to find how to test orientation, I am opening debug using npm start -- --sdebug and this is image of my debug console 

<img width="1636" height="834" alt="Image" src="https://github.com/user-attachments/assets/9ccd8b20-4ae7-48bd-843e-bf8ad695afc3" />   
[...]

PR #1869, opened 2026-01-02
Adding water mode to physics game

HarleenKaurBhullar:
I have added a Water Mode to the physics game by creating a custom environment using the PhysicsJS library, It resoves #1863 
Since PhysicsJS does not provide an inbuilt water or fluid module, I implemented a custom solution based on the concepts of buoyancy and drag force to simulate objects sinking and floating in water.

This pull request includes:
- A custom environment for water physics
- Implementation of buoyant force and drag to approximate fluid behavior
- Visual demonstration of objects floating and sinking based on their properties

I have attached a short video demonstrating the current behavior and implementation.
If this aligns with the project goals, I’d be happy for this to be merged.
I would really appreciate it if you could take a look and share your feedback or suggestions for improvement. Any reviews are welcome!

[Screencast from 2026-01-02 17-48-32.webm](https://github.com/user-attachments/assets/530afd94-4406-4bad-bd7f-4da7289fcab1)

### DIFF
--- a/activities/PhysicsJS.activity/css/activity.css
+++ b/activities/PhysicsJS.activity/css/activity.css
@@ -8,6 +8,23 @@ body {
   width: 100%;
 }
 
+#viewport::after {
+    content: '';
+    position: absolute;
+    bottom: 0;
+    left: 0;
+    right: 0;
+    height: 100%;
+    background: rgba(0, 150, 255, 0.15);
+    pointer-events: none;
+    opacity: 0;
+    transition: opacity 0.3s;
+}
+
+#viewport.water-mode::after {
+    opacity: 1;
+}
+
 #main-toolbar #activity-button {
   background-image: url(../activity/activity-icon.svg);
   z-index: 3;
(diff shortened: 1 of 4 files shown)


llaske:
Nice start.
My remarks:
- Asking for density and mass for each item is too complex for children and will be impossible to use on tablet. May be you could set this value automatically depending of the size of the item.
- It will be nice to display the limit between water and air. So the board will be partially white and blue and we could better see the floating items. Ideally, items will start to fall in the air then float (or not) when in the water.
- It will be nice to support orientation change.
- Your code mix indentation using SPACE and using TAB. You should use the same indentation than the one in the file.
- You're supposed to send PR on dev branch only, see https://github.com/llaske/sugarizer/blob/master/CONTRIBUTING.md

HarleenKaurBhullar:
Thank you, yes I was also thinking of water and air barrier, I just wanted to show you work done till now. Also one question should both mass and density be automatic or should we let children write one of them.

llaske:
> both mass and density be automatic

Yes. I think so.
I don't see a solution to allow setting it without a complex UX

HarleenKaurBhullar:
Okk I will try implementing it in this way now.

PR #1872, opened 2026-01-02
Fix incorrect assignment operator in PhysicsJS activity

### LINKED ISSUE #1870: [Bug] Use of '=' instead of '==' in if else statement in Physics game activity.js
While reviewing the physics activity, I noticed a small bug in an if-else statement where = is used instead of ==. I’ve attached a screenshot for reference.
If you’d like, I can create a quick PR to fix this.

Image

chirag-patil625:
This PR fixes an issue in the PhysicsJS activity where an assignment operator
(=) was mistakenly used instead of a comparison operator in conditional
statements.

While reviewing the code, the same bug pattern was found in another
conditional and has been fixed as well.

Fixes #1870


llaske:
As mentioned in https://github.com/llaske/sugarizer/blob/master/CONTRIBUTING.md
- Do not send a Pull Request without a related issue that has been approved by a Sugarizer maintainer,
- Do not send a Pull Request on master branch

PR #1873, opened 2026-01-02
Correcting "=" to "==" in case of convex polygon in case of if-else loop.

### LINKED ISSUE #1870: [Bug] Use of '=' instead of '==' in if else statement in Physics game activity.js

HarleenKaurBhullar:
in activities.js in Physics activity i have corrected "=" to "==" in case of if-else loop. Closes #1870


llaske:
As mentioned in https://github.com/llaske/sugarizer/blob/master/CONTRIBUTING.md
- Do not send a Pull Request without a related issue that has been approved by a Sugarizer maintainer,
- Do not send a Pull Request on master branch

Issue #1871, opened 2026-01-02
[v2][Bug] Local Journal toolbar enters inconsistent state after “Copy to Remote” action


Alok-2005:
**Description**
There is a UI state management issue in the Local Journal related to item selection and toolbar state handling.

After performing a “Copy to Remote” action, the selected journal entry remains checked. When the user manually clears the selection by unchecking the item, the Journal toolbar temporarily disappears and enters an inconsistent state instead of remaining visible with its default controls.

This results in a visible UI glitch and an unclear toolbar state for the user.

**Steps to Reproduce**

- Open the Journal
- Select a single journal entry using the checkbox
- Click “Copy to Remote”
- Confirm the action by clicking “OK”
- Observe that the checkbox remains selected
- Manually uncheck the checkbox

**Actual Behavior**

- After confirming “Copy to Remote”, the journal entry remains selected
- When the selection is manually cleared , the toolbar briefly disappears.
- This creates an inconsistent and confusing UI transition

**Expected Behavior**

- After completing the “Copy to Remote” action:
- The selection state should be cleared automatically.

**Additional Notes**

- This issue was observed consistently after performing a remote copy action
- The behavior appears to be related to selection-state and toolbar visibility synchronization
- A screen recording is attached for clarity

 

https://github.com/user-attachments/assets/9d8acf72-fe3b-4f10-938d-b718c3adda53

Alok-2005:
@llaske If this is unintended behavior, I’d be happy to work on addressing the selection-state and toolbar inconsistency and submit a PR.

llaske:
You're right, it's a bug.

llaske:
Fixed in https://github.com/llaske/sugarizer/pull/1876

PR #1876, opened 2026-01-03
Fix toolbar state after copy to remote action

Alok-2005:
This PR fixes a UI state inconsistency in the Local Journal where,
after a "Copy to Remote" operation, the selection state was not
cleared and the toolbar entered an inconsistent state.

The fix ensures:
- Selection state is cleared after the operation
- Toolbar remains visible and stable

Fixes: #1871

https://github.com/user-attachments/assets/3d03a9d2-8f78-4b52-9071-4acfe160519a


llaske:
Nice. Thanks.

PR #1880, opened 2026-01-04
Fix undefined constant reference in showLocalJournal()

### LINKED ISSUE #1877: [v2][BUG]:Incorrect constant reference in showLocalJournal() early return check
## Description:
- Summary
The  **[showLocalJournal()](vscode-file://vscode-app/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html)** function in **[journal.js](vscode-file://vscode-app/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html)** references a non-existent constant **[journalType](vscode-file://vscode-app/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html)** in its early return check, causing the optimization to never execute.

## Location
**journal.js** , line 975

## Current code : 
Image

## Issue:
- The constant **this.constant.journalType** does not exist. The constant object only defines:

Image

- This causes the comparison **this.journalType === this.constant.journalType** to always evaluate to **0 === undefined**, which is **false**, preventing the early return from ever executing.
- There's NO **[journalType](vscode-file://vscode-app/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html)** property! It should be **[journalLocal](vscode-file://vscode-app/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html)**.
## Impact:

- When a user clicks the "Local Journal" button while already viewing the local journal, the function does not return early
- This causes unnecessary re-filtering and re-rendering of the journal entries
- Minor performance waste due to redundant processing

## Evidence Of Copy paste-Error
- The other two journal switching functions correctly use their respective constants:
-  Line 982 - ✅ Correct
-  Line 990 - ✅ Correct
-  Line 975 - ❌ Incorrect

Image

## Verification:
Search the entire codebase for **this.constant.journalType** - it appears only on line 975 and is never defined anywhere, confirming this is an undefined property reference.

karthik-dev56:
Fixes #1877

### DIFF
--- a/js/screens/journal.js
+++ b/js/screens/journal.js
@@ -972,7 +972,7 @@ const Journal = {
 		},
 
 		showLocalJournal() {
-			if (this.journalType === this.constant.journalType) return;
+			if (this.journalType === this.constant.journalLocal) return;
 			this.trace("show_journal", "local");
 			this.journalType = this.constant.journalLocal;
 			this.journalId = null;


karthik-dev56:
Hey @llaske I have successfully fixed this issue as you said

llaske:
Thanks.

PR #1894, opened 2026-01-08
Clean up Icon component on unmount

### LINKED ISSUE #1891: [v2][Bug] Memory leak in Icon component - Event listeners and DOM references not cleaned up
## Description:

While reviewing the Icon component in icon.js, I noticed potential memory leaks that could cause gradual memory growth in long-running sessions.

## Issues identified:

1. **Event listener not removed (Line 223-226**

Image

  This event listener is never removed. If icons are created/destroyed multiple times, these listeners accumulate in memory.

2. **Missing cleanup lifecycle hook**
   The component stores DOM references (`_element`, `_svg`, `_originalColor`) but has no `unmounted` lifecycle hook to clean them up when the component is destroyed.

## Impact:
- Gradual memory growth during long sessions
- More noticeable when frequently switching activities/views

Laxmi01345:
## Description
Fixes memory leak issues in the Icon component that could cause gradual memory growth during long-running sessions.

## Changes
- Added `_errorHandler` property to store event listener reference for proper cleanup
- Modified error event listener to use stored reference instead of anonymous function
- Added `unmounted()` lifecycle hook to clean up resources when component is destroyed
- Removed event listeners on unmount to prevent listener accumulation
- Cleared DOM references (`_element`, `_svg`, `_originalColor`) on unmount to allow garbage collection

## Impact
- Prevents memory leaks when icons are created/destroyed multiple times
- Reduces memory growth during long sessions
- More noticeable improvement when frequently switching activities/views

Fixes #1891

### DIFF
--- a/js/components/icon.js
+++ b/js/components/icon.js
@@ -56,6 +56,7 @@ const Icon ={
 			disabledData: this.disabled ? this.disabled: false,
 			_originalColor: { fill: null, stroke: null},
 			_element: null,
+			_errorHandler: null,
 			fallback: false
 		}
 	},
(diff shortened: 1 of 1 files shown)


llaske:
Nice. Thanks.

PR #1875, opened 2026-01-03
Correcting "=" to "==" in Physics activity.

HarleenKaurBhullar:
Following the review of my previous PR, I have created this new pull request targeting the dev branch to resolve #1870, as the issue appears to have been approved, according to my understanding.

Kindly review it and let me know if any changes are needed. Thank you!

PR #1900, opened 2026-01-09
Adding Water Mode to Physics js activity

### LINKED ISSUE #1863: [Feat] Adding mass property to objects and float sink behaviour in water.

HarleenKaurBhullar:
Based on the previous feedback, I’ve updated the implementation so that density is now automatically assigned to objects based on their size.
Closes #1863
Below is a short video demonstrating the changes in action.
I’d really appreciate any feedback or suggestions for further improvement.
[Screencast from 2026-01-09 12-47-27.webm](https://github.com/user-attachments/assets/3c01bd40-09cf-48bf-b2e8-77c089418a5d)
Please note that currently density and different sizes are also getting applied in gravity and newton mode but if you want I would happy to limit them to water mode.

### DIFF
--- a/activities/PhysicsJS.activity/js/activity.js
+++ b/activities/PhysicsJS.activity/js/activity.js
@@ -44,6 +44,8 @@ define(["sugar-web/activity/activity","tutorial","l10n","sugar-web/env"], functi
 		var body = document.getElementById("body");
 		var innerWidth = body.offsetWidth;
 		var innerHeight = body.offsetHeight;
+		var prevWidth = innerWidth;
+		var prevHeight = innerHeight;
 		var toolbarHeight = 55;
 		var outerWidth = 0; // Use to determine if items could disappear, could be 300;
 		var init = false;
@@ -299,24 +301,49 @@ define(["sugar-web/activity/activity","tutorial","l10n","sugar-web/env"], functi
 
 
 			// resize events
-			window.addEventListener('resize', function () {
-				if (resizeTimer) {
-					clearTimeout(resizeTimer);
-				}
-				resizerTimer = setTimeout(function() {
-					renderer.resize(body.offsetWidth,body.offsetHeight);
-					// as of 0.7.0 the renderer will auto resize... so we just take the values from the renderer
-					viewportBounds = Physics.aabb(0-outerWidth, toolbarHeight, renderer.width+outerWidth, renderer.height);
-					// update the boundaries
-					edgeBounce.setAABB(viewportBounds);
-					innerWidth = body.offsetWidth;
-					innerHeight = body.offsetHeight;
-					WATER.updateBoundary();
-					zoom();
-				}, 500);
+			// Optimized Real-Time Resize
+window.addEventListener('resize', function () {
+    const newWidth = body.offsetWidth;
+    const newHeight = body.offsetHeight;
 
-			}, true);
+    // 1. Calculate the instantaneous scale ratio
+    const scaleX = newWidth / prevWidth;
+    const scaleY = newHeight / prevHeight;
 
+    // 2. Update Physics Bounds immediately
+    viewportBounds = Physics.aabb(0 - outerWidth, toolbarHeight, newWidth + outerWidth, newHeight);
+    edgeBounce.setAABB(viewportBounds);
+
+    // 3. Update all bodies smoothly
+    var physicsBodies = world.getBodies();
+    physicsBodies.forEach(function(b) {
+        // Update positions based on the change ratio
+        b.state.pos.x *= scaleX;
+        b.state.pos.y *= scaleY;
+
+        // Scale velocity so they don't lose or gain momentum unnaturally
+        b.state.vel.x *= scaleX;
+        b.state.vel.y *= scaleY;
+
+        b.recalc(); 
+    });
+
+    // 4. Update the renderer immediately for visual feedback
+    if (renderer) {
+        renderer.resize(newWidth, newHeight);
+    }
+
+    // 5. Update global state for the NEXT resize event
+    innerWidth = newWidth;
+    innerHeight = newHeight;
+    prevWidth = newWidth;
+    prevHeight = newHeight;
+
+    // Update water and force a render frame
+    WATER.updateBoundary();
+    world.wakeUpAll();
+    world.render(); 
+}, true);
 			// handle toolbar buttons
 			document.getElementById("box-button").addEventListener('click', function (e) {
 				currentType = 1;
--- a/activities/PhysicsJS.activity/js/activity.js
+++ b/activities/PhysicsJS.activity/js/activity.js
@@ -20,11 +20,15 @@ define(["sugar-web/activity/activity","tutorial","l10n","sugar-web/env"], functi
 		var sensorButton = document.getElementById("sensor-button");
 		var gravityButton = document.getElementById("gravity-button");
 		var appleButton = document.getElementById("apple-button");
+		var waterButton = document.getElementById("water-button");
 		var runButton = document.getElementById("run-button");
 		var readyToWatch = false;
 		var sensorMode = true;
 		var newtonMode = false;
 		var resizeTimer = null;
+		var watermode = false;
+
+
 		if (useragent.indexOf('android') != -1 || useragent.indexOf('iphone') != -1 || useragent.indexOf('ipad') != -1 || useragent.indexOf('ipod') != -1 || useragent.indexOf('mozilla/5.0 (mobile') != -1) {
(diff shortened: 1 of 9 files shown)


karthik-dev56:
Hey @HarleenKaurBhullar @llaske  we can implement this feature too : 

Object Counter Display
Why it's good: Helps children understand they've created many objects
Educational value: Counting, resource awareness
Implementation: Small overlay showing "Objects: 15" or with icons per shape type
Complements Water Mode: Shows density distribution (e.g., "3 floating, 5 sinking")

HarleenKaurBhullar:
Hi @karthik-dev56 interesting but I am not sure idea does not seem to add much to the purpose of learning especially in this pr, maybe ask llaske.

karthik-dev56:
Hey @HarleenKaurBhullar you are right but new ideas and features enhance the depth of the platform usage also right.... which also helps children to learn and gain more knowledge

HarleenKaurBhullar:
Yes maybe you are right

HarleenKaurBhullar:
Hey @llaske would love if you can review the changes.

activities/PhysicsJS.activity/js/activity.js
```diff
@@ -423,7 +529,7 @@ define(["sugar-web/activity/activity","tutorial","l10n","sugar-web/env"], functi
 				} else if (savedObject.type == "rectangle") {
 					newOptions.width = savedObject.width;
 					newOptions.height = savedObject.height;
-				} else if (savedObject.type == "convex-polygon") {
+				} else if (savedObject.type = "convex-polygon") {
```
llaske: Please report fix in PR https://github.com/llaske/sugarizer/pull/1875

llaske:
My remarks:

* [ ] Not sure to understand why there is two colors for water

image

* [ ] It's no longer possible to resize items before adding it, it raise an error

image

* [ ] Even small items sink, the default density should be updated
* [ ] Increasing window size is not handle
* [ ] Changing orientation is not handle
* [ ] It should not be possible to combine Newton mode and Water mode
* [ ] I'm worried about lot of hard coded values in the code. The activity is supposed to be reactive

activities/PhysicsJS.activity/js/activity.js
```diff
@@ -96,6 +101,47 @@ define(["sugar-web/activity/activity","tutorial","l10n","sugar-web/env"], functi
 				,cof: 0.8
 			});
 
+			const WATER = {
+				y: 400,        // water surface Y
```
llaske: Is it not related to the Window size?

HarleenKaurBhullar:
Thank you @llaske for reviewing the PR , I really appreciate your time and feedback.

- I will update the implementation to make it responsive according to the window size. Regarding the resizing issue, you are absolutely right, that was my mistake. I was so focused on making the behavior automatic and overlooked this aspect, but I will fix it.

- Concerning Newton mode and Water mode, I combined them because gravity is still present in water. However, I would appreciate some clarification here: for Water mode, should gravity be redefined or handled separately from Newton mode?

- Regarding small items sinking, in my case very small items where floating so I thought this was fine as most of the items sinking seemed realistic to me but I appreciate your point I would adjust the parameters for the predefined small and medium density category depending upon radius as defined by user.
 
- Also, could you please elaborate a bit more on your concern about the use of hard-coded values?
Would it be acceptable to keep the shape resize consistent with the original code and, based on that, define an appropriate density range?

Again thank you for your time and feedback.

activities/PhysicsJS.activity/js/activity.js
HarleenKaurBhullar: Okk definitely.

activities/PhysicsJS.activity/js/activity.js
HarleenKaurBhullar: I will definitely change it to make it according to window size.

llaske:
> Concerning Newton mode and Water mode, I combined them because gravity is still present in water

In Newton the water will probably form a bubble instead of staying a the bottom of the board.
I think it better to forbid to activate Water mode at the same time than Newton mode.

> could you please elaborate a bit more on your concern about the use of hard-coded values

It's okay to use some constant but you should not use hard coded value in pixel as the screen size could vary.
So for example the size of water should be a part of the window height (2/3 of size ?) not set to a pixel size (400).

HarleenKaurBhullar:
Okk

HarleenKaurBhullar:
Hey @llaske I wanted to share my progress so far. I have addressed almost all of our previous points except the orientation-related one. Actually I wasn’t able to figure out how to test orientation changes using sdebug, and some college work came up as well. I’ll be testing this soon and will make sure it works properly.

In the meantime, could you please review the current work for the other points (excluding orientation changes)? I’ve attached a screen recording for reference. Thanks!                                                                                                                                    
[Screencast from 2026-01-21 18-04-48.webm](https://github.com/user-attachments/assets/c3d1fb41-a592-4cbd-b9f2-51202ccac5b9)

llaske:
That's better.
Testing orientation is simple. You don't need to have a specific device just click on the orientation icon in the toolbar change orientation from 90°.
Also, you should add a tutorial step to describe the new feature.

HarleenKaurBhullar:
Okk, thank you for guidance, I will soon add tutorial and orientation change.

HarleenKaurBhullar:
Hey @llaske, sorry if I’m bothering you with small doubts, but I’m unable to find the orientation icon anywhere in DevTools. If you don’t mind, could you please help me with this? I’ve attached a screenshot of my DevTools for reference.    
image
I have tried using sensors, but they are not showing any change even in default gravity mode.

llaske:
I talked about the icon in the activity toolbar. Just clicking one it change the orientation of the scene

HarleenKaurBhullar:
Ohh ohkk sorry I misunderstood. I thought it was in the debug console. But then, for orientation, like the bottom-left corner, would I need to move water as well? @llaske

llaske:
You should move the water at the same place than indicated by the floor position on the icon

HarleenKaurBhullar:
okk thank you.

HarleenKaurBhullar:
Hey @llaske I have added orientation and tutorials. Would appreciate a review when convenient.

HarleenKaurBhullar:
I just realized objects were going a little beyond bottom line due to viewport in css, so I have corrected it.

llaske:
Nice.
Two remarks:
* [ ] The water zone should never cover the toolbar.

image

* [ ] Once the water zone has been selected then unselected, it's no longer possible to change orientation

https://github.com/user-attachments/assets/b6c5aae7-1a3c-4d92-ad94-0a5d5557ebe6

HarleenKaurBhullar:
Thanks for the review @llaske . I have made the corrections. Kindly review them.

HarleenKaurBhullar:
Hey @llaske please take a review whenever you have time.

llaske:
* [ ] Water mode should be saved in the Journal. 
So the user will have the same board when reopening the activity.

https://github.com/user-attachments/assets/00ff35da-d3e4-4cfa-aad0-49d7dd85d749

* [ ] When board is resized the items position are not reevaluated.

https://github.com/user-attachments/assets/a681a1ad-ea5e-4084-ac19-e2d8b2736a3c

HarleenKaurBhullar:
Okk

HarleenKaurBhullar:
Hey @llaske, I’ve added the water mode to the journal and also implemented logic to reposition objects on window resize. The repositioning might feel a bit abrupt. If you think it needs to be smoother, I can look into alternative approaches, but currently it seems a little difficult.

llaske:
That's good.

Please take care of indentation. It should be TAB here. You've lines with SPACE indentation or with a mix of TAB and SPACE.

image

image

image

HarleenKaurBhullar:
Ok I must have missed it.

HarleenKaurBhullar:
I have made the corrections @llaske.

llaske:
Good job @HarleenKaurBhullar . It's a fun feature :+1:

HarleenKaurBhullar:
Thanks

PR #1903, opened 2026-01-09
ColorMyWorld toggle colour on repeated clicks

### LINKED ISSUE #1890: Allow de-coloring a country in Color My World
While using the Color My World activity in Sugarizer, I noticed that once a country is colored, there is no way to remove that color by clicking on the same country again. This makes it difficult to fix mistakes or change choices during use.

I’d like to suggest adding a simple way to de-color a country after it has been colored.

Use case

Users may accidentally color the wrong country

Users may want to change or redo their selections

Trial-and-error exploration becomes limited without an undo option

Suggested enhancement
Possible approaches:

Clicking on a colored country again toggles it back to the default (uncolored) state

Or providing a small erase / clear mode for removing colors

This would improve usability and make the activity more flexible.

https://github.com/user-attachments/assets/ed876a12-d0f0-43c7-bfab-0a03072cc220

nithinshiju-dev:
Allows a region to be de-colourized by clicking it again during free colouring mode.
Fixes #1890.

https://github.com/user-attachments/assets/3cee6495-82b3-4cc3-a2fc-740d13aadf4f

### DIFF
--- a/activities/ColorMyWorld.activity/lib/map.js
+++ b/activities/ColorMyWorld.activity/lib/map.js
@@ -18,37 +18,72 @@ define(["activity/ol","print","util","colormyworld","humane","flag","l10n"],
 			}),
 		});
 
-		window.map.on('click',function(evt){
-			if(colormyworld.mode==COLORING){
-				var FOUND=false;
-				dummmy=window.map.forEachFeatureAtPixel(evt.pixel,function(target_feature,layer){
-				var target_name=target_feature.get("NAME");
-				if(!target_name)target_name=target_feature.get("Name");
-				if(!target_name)target_name=target_feature.get("name");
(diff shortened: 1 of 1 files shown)


llaske:
Nice.
I've got two remarks:
- Toggle should depend of the current selected color. If your current selected color is Yellow and the current country color is Yellow, clicking on the country will remove the color. But if your current selected color is Red and the current country color is Yellow, clicking on the country will be change to Red.
- It will be nice to do also the toggle on the sea.

nithinshiju-dev:
https://github.com/user-attachments/assets/d5c6cfa6-c730-4918-8f7c-1410521692d9

@llaske I’ve implemented the suggested changes.

llaske:
Nice. It's fun. Thanks.

Issue #1897, opened 2026-01-08
[v2][Bug] Memory Leak and Console Crash in ListView component - Popup timer not cleaned up


Alok-2005:
*Description*

A critical stability issue and memory leak exists in the ListView component (js/screens/listview.js). The component uses `window.setInterval` to manage popup timers but fails to clear this interval when the component is unmounted (e.g., when switching views).

This leads to a recursive error loop where the detached component continues to execute code every 1 second, attempting to access DOM elements that no longer exist.

 `js/screens/listview.js`
Method: `showPopupTimer` (Lines 207-212) and `removePopupTimer` (Lines 243-248).

<img width="700" height="238" alt="Image" src="https://github.com/user-attachments/assets/009e9af8-e71c-45ac-b0da-f966de9b8b67" />

If the user navigates away from the List View while this timer is active (e.g., while hovering an icon), the interval triggers removePopup on a destroyed component instance.

*Steps to Reproduce*

- Open Sugarizer and log in.
- Go to List View (click the list icon in the top-right toolbar).
- Hover your mouse cursor over any activity icon (to frame a mouseover event).
- Immediately (within 1 second) click the Home View button (top right radial icon) to switch screens.
- Open the Browser Console.

Actual Behavior
The console logs an Uncaught TypeError because the timer fires in the background against a null reference:

Expected Behavior
The timer should be immediately cleared when the ListView component is unmounted. No code from ListView should execute after the user has navigated to Home.

Impact
Application Stability: Causes unhandled exceptions in the console during standard navigation flows.

A screen recording is attached for clarity

https://github.com/user-attachments/assets/79a89f0d-0f6d-45e4-bd73-be56a3022384

Alok-2005:
@llaske If this issue is valid, I’d be happy to submit a PR.

llaske:
Good point. You're right, I've reproduced on my side.

llaske:
Fixed in https://github.com/llaske/sugarizer/pull/1906

PR #1906, opened 2026-01-09
fix: resolve memory leak and crash loop in ListView popup timer

Alok-2005:
This PR fixes a critical stability issue and memory leak in the ListView component where the popup timer continued running after the component was unmounted, causing a crash loop.

The fix ensures:

- Timers are strictly cleared using a beforeUnmount lifecycle hook when leaving the view.
- Prevent infinite loops by replacing setInterval with setTimeout for one-time popup events.
- Console errors are eliminated when navigating quickly between List View and other screens.

Fixes: #1897

### DIFF
--- a/js/screens/listview.js
+++ b/js/screens/listview.js
@@ -103,6 +103,12 @@ const ListView = {
 		this.getUser();
 	},
 
+	beforeUnmount() {
+		if (this.timer) {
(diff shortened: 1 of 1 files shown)


llaske:
Nice. It's better like that.

PR #1912, opened 2026-01-11
Fix: image resize, transition lag issue

Rejah-Rabeeul:
Fix: #1698 

Implemented a cache system, batched rendering, fixed a memory leak, and enabled instant page turning.

Preview: 

https://github.com/user-attachments/assets/080fa4ef-13b5-443d-8812-235770a38269

### DIFF
--- a/activities/Fototoon.activity/js/toon.js
+++ b/activities/Fototoon.activity/js/toon.js
@@ -68,11 +68,57 @@ define(["easel","sugar-web/datastore","sugar-web/env","l10n","humane"], function
     function createAsyncBitmapButton(globe, url, callback) {
         // creates a square black button with a image inside
         // is used for the corner controls in the globe
+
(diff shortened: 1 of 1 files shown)


llaske:
You're not supposed to send a PR before an issue is declared. See https://github.com/llaske/sugarizer/blob/master/CONTRIBUTING.md
So please start by create one or more issue then wait for an approval before sending PR.

Rejah-Rabeeul:
Apologies for the confusion. This PR is actually intended to resolve the existing Issue #1551. There was a previous attempt to fix this in PR #1698 by another contributor. he couldn't fix the issues like memory leak, page transition lag, so I commented asking "can i do this task" in PR #1698. In this case also do i need to raise an issue??

Rejah-Rabeeul:
@llaske sir do i need to raise another PR for issue #1551

Issue #1921, opened 2026-01-15
[v2][BUG] Global Variable Leak in i18next module due to variable name typo


Alok-2005:
**Description:**
A naming inconsistency in the `getFormattedSizefunction` within the `i18next.js` module causes a size string to leak into the global scope. A local variable is declared as formatted, but the function subsequently assigns results to formated (missing a 't').
Because formated is never declared using var, let, or const, JavaScript attaches it to the window object in the browser, polluting the global namespace.

In `js/modules/i18next.js`:

<img width="803" height="443" alt="Image" src="https://github.com/user-attachments/assets/e2e9e8aa-fdcc-420a-b9e6-5d56c44f4473" />

**Impact:**
- Global Namespace Pollution: Every time a file size is formatted (e.g., when opening the Journal), window.formated is overwritten with a string.
- Potential for Side Effects: Other modules or future libraries that accidentally use a variable named formated will encounter unexpected state changes, leading to bugs that are extremely hard to debug.

**Reproduction Steps:**
- Launch Sugarizer and open the Browser Developer Tools.
- In the Console, type: window.formated. Observe that it is undefined.
- Navigate to the Journal view (where file sizes are calculated and displayed).
- In the Console, type: window.formated again.
- Result: The variable now contains a string, proving it has leaked into the global window scope.

Alok-2005:
@llaske If this is a valid issue, I’d be happy to submit a PR.

llaske:
That's true. It can't cause issue but it's better to fix it.

llaske:
Fixed in https://github.com/llaske/sugarizer/pull/1922

PR #1922, opened 2026-01-16
Fix: Global variable leak in i18next module due to variable typo

Alok-2005:
## Description
This PR fixes a global variable leak in the [getFormattedSize]

### The Issue
A local variable was declared as `formatted`, but the function used an undeclared variable `formated` (missing a 't') for assignments. This caused the size string to be attached to the global `window` object every time a file size was formatted.

### Changes
- Corrected all instances of `formated` to the properly declared `formatted` variable.
- Verified that the variable remains scoped to the function and no longer pollutes the global namespace.

Fixes:#1921


llaske:
Nice. Thanks.

Issue #1933, opened 2026-01-21
[BUG][Security]: Stored XSS via Buddy Name


Alok-2005:
#### **Description**
A **Stored Cross-Site Scripting (XSS)** vulnerability has been identified in the Chat Activity. A user can set a malicious string as their "Buddy Name" in the Sugarizer settings, which will then execute arbitrary JavaScript in the browsers of all participants in a shared Chat session whenever that user begins to type.

#### **Technical Analysis**
The vulnerability is located in the `TypingEffect` component. The user's name is rendered using the `v-html` directive without prior sanitization:

The `typingString` computed property constructs a string that includes the raw Buddy Name of the typing user. Because it is bound via v-html, the browser interprets any HTML tags (like <script> or <img onerror=...>) within the name as executable code.

- File: `activities/Chat.activity/js/TypingEffect.js`
- Vulnerable Line: Line 11: `<div v-html="typingString"></div>`

**Steps to Reproduce**

- Open Sugarizer and go to Settings > About Me.
- Set your name to: `<img src=x onerror=alert('XSS_VULNERABILITY')>`
- Launch the Chat activity and Share it to the Neighborhood.
- Join the shared session as a second user (using a different browser or private window).
- In the first window (Sender), start typing in the chat input area.
- Result: The second user (Receiver) will see an alert box immediately when the typing indicator appears.

**Suggested Fix**
Refactor TypingEffect.js to avoid using v-html. Instead, use standard Vue text interpolation.

 Screenshots are attached for clarity:

Sender Side View 

<img width="1898" height="877" alt="Image" src="https://github.com/user-attachments/assets/62176d93-fae5-4e47-87bb-d3021dafeb44" />

Receiver Side View 

<img width="1765" height="940" alt="Image" src="https://github.com/user-attachments/assets/d2a893d1-da7b-4021-9638-5e913b9182ca" />

<img width="1770" height="958" alt="Image" src="https://github.com/user-attachments/assets/50fa068b-6be3-42fe-b3bf-be250582cf20" />

Alok-2005:
@llaske if this is a valid issue, I would be happy to submit a PR to fix it

llaske:
You're right.

llaske:
Fixed in https://github.com/llaske/sugarizer/pull/1939

PR #1939, opened 2026-01-23
Fix Stored XSS vulnerabilities on user name

Alok-2005:
**PR Description**
This PR addresses multiple Stored XSS vulnerabilities in the Chat Activity triggered by malicious user "Buddy Names".

**Key Changes:**
- Typing Indicator: Refactored TypingEffect.js to use safe Vue text interpolation `{{ }}` instead of `v-html`.
- Join/Leave Notifications: Added an `escapeHTML` utility in `activity.js` and sanitized user names before they are rendered in the notification popups.
- User List: Updated the local `presencepalette.js` to use `textContent` instead of `innerHTML` for rendering names in the presence palette.

 Fixes:#1933

### DIFF
--- a/activities/Chat.activity/js/TypingEffect.js
+++ b/activities/Chat.activity/js/TypingEffect.js
@@ -2,15 +2,26 @@ const TypingEffect = {
 	props: ["usersTyping"],
 	template: `
 		<div class="loader-container">
-			<div class="dots" v-if="Object.keys(usersTyping).length > 0">
+			<div class="dots" v-if="keys.length > 0">
 			  <span class="dot"></span>
 			  <span class="dot"></span>
 			  <span class="dot"></span>
 			</div>
 			<div>
-			    <div v-html="typingString"></div>
+				<div v-if="keys.length === 1">
+					<strong>{{ usersTyping[keys[0]].name }}</strong> {{ l10n.stringIs }} {{ l10n.stringTyping }}
+				</div>
+				<div v-else-if="keys.length === 2">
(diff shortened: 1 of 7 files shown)


llaske:
Oh. I thought that the issue was related to Chat activity only but it's related to Sugarizer and even to Sugarizer Server.
It must not be possible to create a local user account without escaped characters.
It must not be possible to change user name without escaped characters.
So the fix should not be in the Chat activity but in Sugarizer core itself.

Plus an issue should be raised also in Sugarizer Server to not accept non escaped characters during signup and update.
Because the XSS issue exist also in Sugarizer Server dashboard.

image

Alok-2005:
Thanks for the feedback, @llaske. I agree that addressing this at the core level is much better.

I am updating this PR to:

1. Move the sanitization logic to `js/modules/user.js` to ensure the name is escaped during signup and update.
2. Revert the changes in the Chat activity files to keep the fix centralized.

I will also open an issue on the `sugarizer-server` repository for the dashboard XSS as you suggested. I'll push the updated changes to this branch shortly.

Alok-2005:
@llaske. I’ve now moved the sanitization/validation logic into the Sugarizer core (js/modules/user.js) so usernames are properly escaped during signup and update, and reverted the Chat Activity–specific changes to keep the fix centralized as suggested.

I will shortly raise a corresponding issue for the sugarizer-server dashboard XSS to ensure server-side validation is aligned with this fix.

llaske:
The invalid name alert should be raise on the name screen, not on the latest screen before creation. 

image

Alok-2005:
@llaske I have implemented the requested changes . The invalid name alert is now raised immediately on name screen itself.

llaske:
Nice. Thanks.

Issue #1935, opened 2026-01-23
Delete key deleted wrong character in Calculator


Rejah-Rabeeul:
Hey @llaske Kindly review this issue ,

1. Delete key deleted wrong character when we enter number by cursor

2. Base Conversion Fails for Floating-Point Numbers

Preview:

https://github.com/user-attachments/assets/7d187313-8295-42aa-99fe-169b3697c5aa

llaske:
It's because you have selected base 2.

Rejah-Rabeeul:
Hi @llaske 
The issue is not selecting base 2 but that floating-point numbers lose their fractional part during base conversion. For example, converting 10.10 (base 10) to base 2 should return a value with digits after the decimal point but currently only the integer part 1010 is returned.

So the problem is floating-point base conversion truncation.

There is another issue with the delete/backspace key. When entering a number and moving the cursor to edit digits in the middle, pressing backspace sometimes deletes the wrong character

llaske:
I don't think it make sense to use floating point numbers in binary. So the current behavior seems right.

You're right for the delete issue.

llaske:
Fixed in https://github.com/llaske/sugarizer/pull/1948

PR #1948, opened 2026-01-26
Fix: Delete key deletes wrong character issue 

Rejah-Rabeeul:
Fixes :  #1935 

Preview:

https://github.com/user-attachments/assets/5c21e04c-eb37-42a8-84ce-afb841384220


llaske:
Nice. Thanks.

PR #1949, opened 2026-01-27
Added instruction clarifying emoji selection is password creation

### LINKED ISSUE #1946: Suggestion: Make it clear that choosing the four emoji images is actually password creation
Hi,

Adult, experienced computer user here. Heard about the great idea of Sugar and though I'd take a look at it via Sugarizer.

One glaring issue seems to be that there is absolutely _**no indication whatsoever**_ that choosing the four or more emoji images at the start actually constitutes the password creation for the account.

Image

Although my perspective may admittedly be different from that of a child, it certainly seemed to me like the emoji choice was purely a cosmetic decision, whether for someone just testing the platform briefly or for someone making an account to use it longer-term. This would be consistent with other child-focused online platforms where the choice of profile pictures (and indeed often usernames) can only be made from a fixed number of predetermined choices (in order to prevent unsuitable pictures and/or usernames being entered onto the platform by a malicious user). As such, it's an absolute miracle that I happened to still remember my particular random choice of emojis a few minutes later when looking around and coming across something in settings requiring password entry.

I realise that in a lot of cases individual or groups of child users will probably be guided through the initial steps of account creation by a classroom teacher or other facilitator, but even so I think that this part of the process could still involve a **_much clearer on-screen indication_** that it actually constitutes password creation, whether via actually using the term 'Password' (or its other-language equivalents) or some other means such as iconified graphics of things like keys and locks.

Admittedly this issue may be more significant for the niche category of adult testing users rather than the majority group of genuine child users because presumably the nature of Sugarizer means that authorised teachers can easily access and change the passwords of users who forget them.

LH

swetalin-10:
This PR adds instructional helper text to the emoji selection step during new user creation. The message clarifies that the selected emojis form the user's password.

What problem users get?
Users may not realize that choosing emojis during setup actually creates their login password, which can cause confusion (especially for new or adult users testing the platform).

What is the solution ?
When a new account is being created, an instruction message is displayed:
"Choose 4 emojis to create your password. Remember them!"

This impact in Improves user experience and reduces confusion during account creation without changing any logic.

Fixes #1946

### DIFF
--- a/js/firstscreen.js
+++ b/js/firstscreen.js
@@ -185,6 +185,10 @@ enyo.kind({
 			this.scrollToField(this.$.passbox);
 			vpassbox = vprevious = vnext = true;
 			this.$.password.setLabel(l10n.get(this.createnew ? "ChoosePassword" : "Password", {min: util.getMinPasswordSize()}));
+			if (this.createnew) {
+				this.$.password.setPlaceholder("Choose 4 emojis to create your password. Remember them!");
+			}
+			
 			this.$.next.setText(l10n.get(this.createnew ? "Next" : "Done"));
 			break;
 
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,7 +9,7 @@
       "version": "1.9.0",
       "license": "Apache-2.0",
       "dependencies": {
-        "axios": "^0.27.2",
+        "axios": "^1.13.3",
         "electron": ">=22.3.25",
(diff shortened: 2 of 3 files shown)


llaske:
You're not supposed to open a PR before your issue has been discuted.
See https://github.com/llaske/sugarizer/blob/master/CONTRIBUTING.md

PR #1950, opened 2026-01-27
Fix Paint canvas not resizing when window size changes

### LINKED ISSUE #1944: Paint canvas does not resize when window size changes
Hi @llaske ,
The Paint activity only measures the screen size when it starts. if it is opened in a small window and later resized the canvas does not expand to fit the new space. this creates a white area that looks like part of the canvas but is actually inactive background so drawing, pasting, or adding images in that area has no effect.

Preview:

https://github.com/user-attachments/assets/9a89dba8-bb57-439d-8ba2-c2b0a5288b6a

swetalin-10:
The Paint activity only measured canvas size at startup.
When the window was resized, the canvas did not expand, leaving an inactive white area.

This fix updates canvas dimensions dynamically when the window size changes.

Fixes #1944

### DIFF
--- a/activities/Paint.activity/js/activity.js
+++ b/activities/Paint.activity/js/activity.js
@@ -125,7 +125,7 @@ define(["sugar-web/activity/activity", "tutorial", "l10n", "sugar-web/env", "act
       var canvas = PaintApp.elements.canvas;
       var ctx = canvas.getContext("2d");
 
-      // 1. Save current content before any changes
+      // 1. Save current content synchronously at the START of resize
       if (!resizeBaseCanvas) {
         resizeBaseCanvas = document.createElement('canvas');
         resizeBaseCanvas.width = canvas.width;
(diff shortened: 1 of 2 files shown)


activities/Paint.activity/js/activity.js
llaske, line 1: Please respect file indentation.

swetalin-10:
Indentation has been fixed as requested.
Please review again. Thank you!

llaske:
Resizing now reset the board.
Did you test on your side?

https://github.com/user-attachments/assets/6c04a583-e6b2-47a7-b23b-084f7ff4c1a5

swetalin-10:
Thanks for reviewing.
Yes, I tested the resize behavior. You’re right, the board was getting reset because resizing the canvas clears its content.
I’ve now updated the code to preserve the drawing by saving the canvas image data before resize and restoring it after resizing.
Could you please review the latest changes?

Rejah-Rabeeul:
Thanks for the effort but I'm still seeing the bug on my end. could you re-check the logic and share a screen recording of it working?

swetalin-10:
Thanks for checking and for the feedback 🙏
I re-tested the resize logic again carefully.
The issue was happening because resizing the canvas resets its content. To fix this, I updated the flow 
I verified that:
The canvas now resizes correctly
Existing drawings remain preserved
No white inactive area appears
I’m attaching a screen recording showing the behavior working correctly on my system.
Could you please check again? If the issue still appears, I’ll be happy to investigate further.

https://github.com/user-attachments/assets/333475cb-7335-433e-b56d-ca79d25d4f2e

Rejah-Rabeeul:
Thanks, I see the original issue is fixed now However I found a new problem that started happening with your changes. I’ve attached a recording of it here take a look at this too.


https://github.com/user-attachments/assets/efb3ccd6-2712-4262-af9e-88660ff4923d

swetalin-10:
Thanks for testing and for sharing the recording 🙏
I checked the resize behavior again on my system, but I’m not experiencing this new issue, the canvas resizes correctly and the drawing remains preserved.
It may be environment-specific. Could you please share:
• Your browser and version
• Your operating system
• The exact steps before the issue occurs
I’ll try to reproduce it with the same setup and work on a fix. Thanks for pointing this out!
https://github.com/user-attachments/assets/b79f6827-0e08-4f85-81f6-747a86010982

Rejah-Rabeeul:
I tested this on the latest versions of both Chrome and Brave and the issue is still there for me on both browsers. I'm running on Windows 11. I followed the same steps as in my recording.

swetalin-10:
Thanks again for testing and for the detailed feedback.
I’ve verified the latest changes on my side (Windows + Chromium-based browsers), and the canvas now resizes correctly with drawings preserved.

I’ve also applied additional compatibility improvements aimed at Windows 11 and Chrome/Brave:
• Synchronizing the canvas drawing buffer with the window size to avoid inactive or “dead” regions
• Forcing a paper.view.draw() refresh after resize to keep Paper.js and the canvas rendering aligned
• Explicit handling of devicePixelRatio to prevent scaling-related shifts or rendering issues on high-DPI displays

Since browser caching can sometimes cause older scripts to run, could you please test again after a hard refresh (Ctrl + F5) or clearing the browser cache?

If the issue still appears after that, I’ll continue investigating — it may be triggered by a specific resize sequence or rendering timing issue that I’ll try to reproduce.

Rejah-Rabeeul:
Its still not fixed on my end

swetalin-10:
Thanks for testing again , Since it’s still reproducible on your side but not on mine, I’d like to narrow it down.

Could you please confirm:
• Does it happen every time or only after several resizes?
• Does it occur when resizing quickly vs. slowly?
• Does it happen only after drawing, or even on a blank canvas?

I’ll add temporary logging around the resize/save/restore flow to verify whether image data is captured and restored correctly during the resize cycle. It may be a timing issue between the resize event and Paper.js rendering on some systems.
I’ll follow up with findings or a fix soon.

Rejah-Rabeeul:
It’s the same issue as shown in the video attached earlier.

swetalin-10:
https://drive.google.com/file/d/1ynIWyx4s9SoetrDixhbmVpUwmW7oOR3n/view?usp=sharing

Hi, thanks again for the feedback and testing 
I’ve updated the resize logic and reworked the implementation to properly handle canvas resizing without resetting the board.
With the latest changes:
The canvas resizes correctly when the window size changes
Existing drawings are preserved during resize
No inactive or “dead” white areas appear after resizing

I tested this on my system (Windows + Chromium-based browser) and the behavior is now consistent. I’ve attached a screen recording showing the resize working while keeping the drawing intact.
Please review the latest commit. If the issue is still reproducible on your side, I’ll be happy to investigate further with the exact steps.
Thank you!

llaske:
I think there is a misunderstanding.
Increasing the window size should not resize the current drawing, it should just increase the canvas size.

swetalin-10:
Thanks for the clarification. I misunderstood the expected behavior earlier.

The resize logic has now been updated so that increasing the window size expands the canvas without scaling or distorting the existing drawing, similar to MS Paint behavior.

Changes:
• The drawing is no longer scaled during resize.
• Resizing only expands or hides canvas space.
• A resize buffer is used to preserve content when the window is shrunk and later expanded.
• Undo/redo and drawing tools remain synchronized with the canvas state.

Verified behavior:
✓ Expanding the window adds new drawable space without stretching the drawing.
✓ Shrinking and re-expanding restores hidden content correctly.
✓ Drawing works correctly in newly expanded areas.
✓ Undo/Redo continues to work properly after resizing.

I tested this on Windows with Chromium-based browsers.
Please let me know if you’d like any additional adjustments.

https://github.com/user-attachments/assets/e2c67d34-fba7-48ea-828a-176e1eaa1805

llaske:
@swetalin-10 your PR change lot of source code just to update white spaces. I guess it's probably to the IDE you're using.
For maintenance reasons only lines directly related to the issue must be updated.

llaske:
Fixed in https://github.com/llaske/sugarizer/pull/1965

PR #1954, opened 2026-02-01
Bug in Fototoon Window resize

Rejah-Rabeeul:
Issue #559 
Fixes window resize issue in Fototoon.

Preview:

https://github.com/user-attachments/assets/23c5364e-8bd1-4c8e-8c59-dbf2e5d62041

### DIFF
--- a/activities/Fototoon.activity/js/activity.js
+++ b/activities/Fototoon.activity/js/activity.js
@@ -151,8 +151,11 @@ define(["sugar-web/activity/activity","sugar-web/datastore","sugar-web/env","tex
 
             // clone the data to remove the images
             var dataWithoutImages = {}
-            dataWithoutImages['version'] = toonModel.getData()['version'];
-            dataWithoutImages['boxs'] = toonModel.getData()['boxs'];
+            var modelData = toonModel.getData();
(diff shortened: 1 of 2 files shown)


llaske:
It solve the issue at the start of the activity but not when the window is resized (more space or less space)

https://github.com/user-attachments/assets/71f1ce93-5ecc-41ef-8882-6f9ca0f45112

Rejah-Rabeeul:
Hey @llaske, I've updated the PR to address the window resize issue.

preview:

https://github.com/user-attachments/assets/eb6317b3-68d2-4986-a7f0-cc8334775fa3

llaske:
Nice. Thanks.

llaske:
I've found a side effect of this fix. When the window is resized during text edition (for title of for globes), the update is not taken into account and there is a glitch on the screen.
Unfortunately it happens all the time on tablet because the virtual keyboard is displayed during edit so the window is resized.

https://github.com/user-attachments/assets/b050531d-7672-4955-b454-2d8b590b0ff8

@Rejah-Rabeeul could you have a look?

Rejah-Rabeeul:
I'll look into it and will raise a PR soon

Issue #1956, opened 2026-02-03
Bug in Flip Activity when resize an ended game


Rejah-Rabeeul:
Hi @llaske ,
After winning a game in Flip activity, if the user resizes the window quickly before jumping to next stage the canvas ends up showing only one column from the previous game.

preview:

https://github.com/user-attachments/assets/f13abbe5-1c6a-478a-971f-21a4fd2f1cce

The-Yearly:
Hey Can I Please Work on this issue

llaske:
@Rejah-Rabeeul you're right. And there is also an error message in the console.

<img width="1023" height="590" alt="Image" src="https://github.com/user-attachments/assets/77a0c0c4-32da-4d02-9ef7-d2643b6c3956" />

But you should be very quick to resize the window.

Rejah-Rabeeul:
Hi @The-Yearly, I'm already working on this issue. You may try fix issue #1944

Rejah-Rabeeul:
@llaske Yah if the user doesn't resize the window quickly, there wont be any issues. I'll look into fixing this bug

The-Yearly:
@Rejah-Rabeeul
Oh Im So Sorry , I Just saw the message after creating the PR

Rejah-Rabeeul:
Okay, no worries.

llaske:
Fixed in https://github.com/llaske/sugarizer/pull/1958

PR #1958, opened 2026-02-04
Fix: Fix for Issue #1956 (Bug in Flip Activity)

The-Yearly:
# Summary

Fix for Issue #1956 a canvas rendering issue in the Flip Activity that occurs when the window is resized immediately after winning a game and before moving to the next stage.

# Issue
After completing a level in the Flip Activity,If  the user resizes the window quickly before advancing to the next stage, the canvas sometimes renders incorrectly.
In this state, only a single column from the previous game remains visible instead of the correct number of column and an error is thrown along with it 
```
Uncaught TypeError: Cannot set properties of undefined (setting 'radius')
at Game.initialiseFromArray (game. 15:207:28)
at resizeCanvas (activity.is:54:6)
```

# Before

https://github.com/user-attachments/assets/2ff8b0f1-69a1-4dcf-ab9f-084bb868b990

# After

https://github.com/user-attachments/assets/5028e0de-e404-4b82-a3ac-948fa1fac407


Rejah-Rabeeul:
I was working on this and was planning to raise the PR. 🙂

The-Yearly:
Yeah imm soo sorry I just saw your comments few hours after raising my pr

package-lock.json
llaske, line 1: This file should not be updated.

The-Yearly: @llaske package-lock.json is now untracked
Apologies for the earlier commit,first time contributing here

llaske:
Sorry, I was not clear: the file should be tracked in Sugarizer but it should not be updated in your PR.

The-Yearly:
@llaske retracked package-lock.json and fetched it og state from dev

llaske:
Nice. Thanks.

PR #1965, opened 2026-02-09
Fix: Fix for issue #1944 (Paint canvas does not resize)

The-Yearly:
# Summary 
This PR is an Fix for #1944 where the canvas is not resized properly when the window dimensions Changes

# Issue
Initially the Paint activity only measures the screen size when it starts. if it is opened in a small window and later resized the canvas does not expand to fit the new space. this creates a white area that looks like part of the canvas but is actually inactive background so drawing, pasting, or adding images in that area has no effect.

# Fix
Normally, simply expanding the canvas doesn’t work, because resizing it directly will reset its contents. To solve this, a reference canvas is used: when resizing, the current image is first copied onto the reference canvas, and then the main canvas is resized. The reference canvas is also useful for keeping track of elements that extend beyond the visible canvas area when the window becomes smaller.

# Before

https://github.com/user-attachments/assets/c6cd421a-5b98-44ca-adae-9dfc96df159b

After

https://github.com/user-attachments/assets/42969baf-1e63-4159-9803-8083dc6b6547

### DIFF
--- a/activities/Paint.activity/js/activity.js
+++ b/activities/Paint.activity/js/activity.js
@@ -55,7 +55,7 @@ define(["sugar-web/activity/activity","tutorial","l10n","sugar-web/env","activit
     // Export as PNG image
     document.getElementById("save-image-button").addEventListener('click', function () {
       var mimetype = 'image/png';
-      var inputData = document.getElementById("paint-canvas").toDataURL(mimetype, 1);
+      var inputData =PaintApp.data.worldCanvas.toDataURL(mimetype, 1);
       var metadata = {
         mimetype: mimetype,
         title: l10n.get("PaintBy", { name: currentenv.user.name }),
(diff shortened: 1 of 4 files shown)


llaske:
Doesn't work on my side.

https://github.com/user-attachments/assets/9ba20e38-d0d9-4b63-91e9-4a46f766ac1f

llaske:
It works now when the window size is increase.
But when the activity is reopen on a smaller window, the non visible part of the drawing is definitively lost.

https://github.com/user-attachments/assets/b553d40e-c574-4845-acbc-d4cf4a387739

The-Yearly:
@llaske
Yeah i also noticed it that Clear Canvas , Undo/Redo and Bucket Also didnt track changes outside the visible canvas so made changes to them too to track them

activities/Paint.activity/js/paint-activity.js
```diff
@@ -50,11 +50,12 @@ function initGui() {
   PaintApp.prevScreenHeight=parseInt(window.innerHeight) - 55;
 
   const worldCanvas = document.createElement('canvas');
-  worldCanvas.width = parseInt(screen.width);
-  worldCanvas.height = parseInt(screen.height);
+  worldCanvas.width = 2000
+  worldCanvas.height = 2000
```
llaske: Not a fan of this implementation. The initial size should be the screen size and should be increase only if needed.

The-Yearly: Initially the buffer canavas was like that it but then it occurred to me if the user changes the device to a smaller device the content would get cropped when exiting, that was why I figured a more set definition of the canvas is needed

llaske: For a new instance of the activity we could initialize size with the screen size.
For an existing instance coming from the Journal the canvas size should be the one when the activity has been stopped.
In both case the canvas size should be expanded when window is expanded.

The-Yearly: Hi could you please review the changes

activities/Paint.activity/js/activity.js
```diff
@@ -94,12 +93,18 @@ define(["sugar-web/activity/activity","tutorial","l10n","sugar-web/env","activit
           return;
         }
         var data = JSON.parse(jsonData);
+        console.log(data,"Hi IM here ")
```
llaske: Please remove this trace.

llaske (changes requested):
Seems okay.
Please fix the log issue.

The-Yearly:
@llaske sir is there any other changes or implementations that I might have to do

llaske:
Nice. Thanks.

Issue #1975, opened 2026-02-18
No minimize/maximize controls when running via npm start


Shekar-77:
When running Sugarizer using:
npm start

The Electron app launches in fullscreen frameless mode by default.
Because of this:
No native OS window frame is shown, No minimize / maximize buttons, Window cannot be resized, Multitasking becomes difficult.
This behavior may not be ideal for desktop users.

How to Reproduce:
Clone the repository.
Run:
npm install
npm start
Observe that: The app opens in fullscreen.

There are no native window controls .The window cannot be minimized or resized.

Expected Behavior:
App should open in windowed mode by default.Native OS window controls (minimize/maximize) should be available.Window should be resizable

Proposed Fix:
Update BrowserWindow configuration in main.js to disable fullscreen and enable frame by default.

Shekar-77:
Hello @llaske , could you please review this?

karthik-dev56:
hey @Shekar-77  without maintaner approval you should not raise a PR

llaske:
It's on purpose. Sugarizer is targeted to children who are not familiar with windows system.
There is command line options to launch it in a window.

PR #1976, opened 2026-02-18
Disabled fullscreen and enabled os window frame.

Shekar-77:
Solves Issue: #1975

Changes Made:
- Disabled fullscreen mode by default.
- Enabled native window frame
- Removed automatic mainWindow.maximize() call
- Updated BrowserWindow configuration in main.js:

fullscreen: false,
frame: true,


llaske:
You're not supposed to send a PR without an issue validated by maintainers. See https://github.com/llaske/sugarizer/blob/dev/CONTRIBUTING.md

Issue #1977, opened 2026-02-18
Journal navbar icon sizing, dropdown alignment & hover contrast issues


cherylpinto:
**Description**
In the Journal navbar, there are several UI inconsistencies that affect usability:

- Navbar icons appear in different sizes
- Dropdown arrows are positioned below the labels instead of aligned next to them
- On hover, the icon color matches the hover background, causing icons to become hard to see or nearly invisible

**Expected Behavior**
- All navbar icons should have a uniform size
- Dropdown arrows should be horizontally aligned next to their labels/icons
- Hover states should maintain sufficient contrast, ensuring icons remain clearly visible

**Steps to Reproduce**
1. Open Sugarizer
2. Navigate to the Journal view
3. Hover over navbar icons and dropdowns

Screenshot
**Currently**

<img width="604" height="70" alt="Image" src="https://github.com/user-attachments/assets/50d364b3-55bc-42b6-9966-377253ed30ff" />

On hover
<img width="622" height="67" alt="Image" src="https://github.com/user-attachments/assets/a5fc28f4-db76-4c8f-9c33-ec5acfee1af6" />

---

**Expected:**

<img width="740" height="71" alt="Image" src="https://github.com/user-attachments/assets/3d704651-c13b-4137-bcda-22d8da36681f" />

On hover

<img width="756" height="72" alt="Image" src="https://github.com/user-attachments/assets/43a86fa5-2aaa-499f-bc30-31ac2428a8ef" />

Note
I’m a first-time contributor and would be happy to work on this fix

llaske:
I don't see the problem.

PR #1978, opened 2026-02-18
fixed icon sizing and hover color

cherylpinto:
**Fixes: #1977** 

### Problem
In the Journal toolbar:
- Icons were inconsistently sized
- Filter dropdown arrows appeared below the icon/label instead of inline
- Hover background reduced icon contrast, making icons hard to see

### Changes
- Standardized all Journal toolbar icon sizes
- Aligned filter dropdown arrows horizontally next to their icons/labels
- Improved hover styling so icons remain clearly visible
- Ensured filter icons keep a consistent size when selections change

### Testing
- Verified consistent icon sizing across the Journal toolbar
- Verified dropdown arrows align correctly and toggle as expected
- Verified hover states maintain proper icon visibility
- Verified filters continue to function normally

Screenshot
**Before**

Image

On hover
Image

---

**After:**

Image

On hover

Image

I’m happy to take feedback and make any required changes promptly

### DIFF
--- a/css/styles.css
+++ b/css/styles.css
@@ -7,6 +7,7 @@ html {
 	height: 100%;
 	-webkit-tap-highlight-color: rgba(0, 0, 0, 0);
 }
+
 body {
 	/* position: relative; */
 	height: 100vh;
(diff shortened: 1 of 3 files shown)


llaske:
You're not supposed to send a PR before there is a validated issue. See https://github.com/llaske/sugarizer/blob/dev/CONTRIBUTING.md

PR #1991, opened 2026-02-24
Fix: Hide filter button when no categories are available

adeba4343-web:
The filter option was always visible even when selected libraries
did not provide category metadata (e.g., Khan Academy, Art4Apps).

This change hides the filter button initially and prepares it to
be shown only when categories are available, improving UX for children.


llaske:
You're not supposed to propose a PR without an issue has been approved. See https://github.com/llaske/sugarizer/blob/dev/CONTRIBUTING.md

adeba4343-web:
Thanks for the clarification. I’m still learning the contribution workflow. I’ll make sure to discuss via an issue before opening any PR next time.

PR #2000, opened 2026-03-04
Fix abacus rod labels to correctly display positional values

shriidevii:
#### Problem

The abacus visualization displayed incorrect values at the bottom of the rods.
The labels were showing only the **raw bead count** (e.g., `7`) instead of the **actual positional value** of the column (e.g., `70`, `700`).

This caused a mathematical inconsistency between:

* bead positions,
* rod labels, and
* the equation displayed above the abacus.

#### Solution

Updated the logic in `updateTextItems()` within `standardabacus.js` so that rod labels now display the **correct positional value** by calculating the total value contributed by each column.

The value is now derived from:

* the number of beads in use in the column, and
* the column’s positional value.

Instead of displaying bead counts, the rod label now displays the computed column value using the existing fraction-based calculation logic.

#### Result

Rod labels now correctly reflect the contribution of each column.
For example:

```
700000000000 + 700 + 70 + 7 = 700000000777
```

This ensures consistency between the abacus bead positions, rod labels, and the equation displayed above the visualization.

#### Files Modified

* `activities/Abacus.activity/js/standardabacus.js`

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,9 +4,9 @@
 
 Sugarizer is a free/libre learning platform. The Sugarizer UI uses ergonomic principles from The [Sugar platform](https://sugarlabs.org/), developed for the One Laptop per Child project and used by more than 2 million children around the world.
 
-Sugarizer runs on every device: laptops, desktops, tiny computers, tablets or smartphones.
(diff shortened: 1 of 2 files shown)


llaske:
You're not supposed to send a PR before an issue has been validated.

Issue #2020, opened 2026-03-13
Globes not displayed when exporting a single image in fototoon


Rejah-Rabeeul:
Hi @llaske 
In the Fototoon activity globes added to a single image are not displayed when using Save as Image. The exported image only contains the base image.

Preview:

https://github.com/user-attachments/assets/8ceeb792-64c8-47b6-bf7b-ad6e27a54c3c

llaske:
You're right. It's a bug.

Saidkhusayn:
Hey! I'd like to work on this issue. 

Reviewing the Fototoon.activity. Will give an update if I find the bug.

Rejah-Rabeeul:
Hi Thanks for your interest in this issue. I’ve already raised a PR on it. You might want to explore other open issues to contribute

llaske:
Fixed in https://github.com/llaske/sugarizer/pull/2021

PR #2021, opened 2026-03-15
Fix: Globes not displayed when exporting a single image in fototoon

Rejah-Rabeeul:
Issue #2020 

Preview:

https://github.com/user-attachments/assets/4293a9d8-935b-4e4f-b7aa-a9e3c7f2ed0a

### DIFF
--- a/activities/Fototoon.activity/js/toon.js
+++ b/activities/Fototoon.activity/js/toon.js
@@ -465,12 +465,20 @@ define(["easel","sugar-web/datastore","sugar-web/env","l10n","humane"], function
                                  this._data['images'], false, {canvas: previewCanvas, order: boxOrder}, function(context) {                 
                     // Add delay to ensure rendering is fully complete (font loading, rasterization)
                     setTimeout(function () {
+                        document.body.appendChild(previewCanvas);
+                        previewCanvas.style.visibility = "hidden";
+                        previewCanvas.style.position = "absolute";
+                        previewComicBox.stage.update();
+                        setTimeout(function () {
                         // FORCE update to ensure everything (globes, images) is painted to the canvas
                         previewComicBox.stage.update();
                         var dataURL = context.canvas.toDataURL("image/png");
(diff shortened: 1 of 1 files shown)


Rejah-Rabeeul:
Hi @llaske 
just following up on this PR. please take a look when you have time. Thanks

llaske:
I'm not a big fan of adding a timer in a timer.
Could you explain your fix?
Also, please indent the code correctly.

Rejah-Rabeeul:
Apologies for the indentation earlier, I was just trying to get the logic over to you quickly.

The fix was implemented to solve a race condition where the background image would load, but the globes and text hadn't finished rendering yet.

Timer 1 : This waits for the heavy binary data (the background image) to be downloaded and decoded by the browser. We don't know if this will take 10ms or 5s.
Timer 2 : Once the image is ready, this tiny extra wait ensures the browser's layout engine has finished placing the text and vector shapes on top of that image.

Technically, you could do it with one timer by just using a very long setTimeout (like 2000ms) right at the start, but that approach is risky and slow. 
If you just set a 2000ms timer at the start, the background image might still take 3000ms to load on a slow connection. You'd end up capturing a blank screen.

So basically, Timer 1 only tells about the background. It knows nothing about the bubbles. We add the 500ms on Timer 2 for specifically to give the Render Engine a chance to notice before we take the final snapshot.

llaske:
Relying on a double timeout doesn't seem to be a good idea. It will lead to random issues.
Could you suggest a solution to replace both timeout?

Rejah-Rabeeul:
Yes, I’ll work on a better solution to replace both timeouts. I’d also appreciate if you could check my other PR when you get a chance.

Rejah-Rabeeul:
Hey @llaske 
I’ve implemented a more reliable solution by replacing both timeouts. Please check it out when you have time.

llaske:
Nice. Thanks.

PR #2034, opened 2026-03-18
Fixing snake spawning too close to the player and sprite clipping

GAMinsect:
As of right now I have fixed the first issue I pointed out.
Now the first time a snake spawn it wont' be near the position of the player (which will be at the spawn).

I also added a check that in case a snake can't find a free space to spawn in, we wait to the next iteration.
I would like to note that this issue was also present in the original version, you can try it by yourself by increasing the number of stones to +10, and you'll see than when snakes have nowhere to spawn they get teleport
at the top left of the map, with their animation clipping into one another.


llaske:

GAMinsect:
sorry i must have missed that out

Issue #2023, opened 2026-03-15
Snake spawning too close to the frog in the FoodChain Activity


GAMinsect:
Currrently it can happen that the snakes starts moving on a column too close to the starting spawn of the frog, not giving the player enough time to escape from it. 
And as long as the snake hitbox interacts with the starting position of the frog, it insta kills it, not allowing the player any chance to escape.

<img width="444" height="674" alt="Image" src="https://github.com/user-attachments/assets/e87787f1-bae5-4305-86e0-19e79d4a750b" />

Moreover if the fly moves too close to the frog, its png overwrites it

<img width="505" height="405" alt="Image" src="https://github.com/user-attachments/assets/d38558e1-7372-4cf1-b133-4de0b0638c0b" />

llaske:
Yes you're right it could be improved

GAMinsect:
I'll start to work on it, and when I'm close to the solution, I'll open the PR

llaske:
Fixed in https://github.com/llaske/sugarizer/pull/2035

GAMinsect:
I accidentally closed this issue when reading the comments.


Sorry for the inconvenience.

PR #2035, opened 2026-03-18
Fixing snake spawning to close to the player and overlapping PNG's

GAMinsect:
Fixing #2023 

As of right now I have fixed the first issue I pointed out.
Now the first time a snake spawn it wont' be near the position of the player (which will be at the spawn).

I also added a check that in case a snake can't find a free space to spawn in, we wait to the next iteration.
I would like to note that this issue was also present in the original version, you can try it by yourself by increasing the number of stones to +10, and you'll see than when snakes have nowhere to spawn they get teleport
at the top left of the map, with their animation clipping into one another.

### DIFF
--- a/activities/FoodChain.activity/playgame.js
+++ b/activities/FoodChain.activity/playgame.js
@@ -197,6 +197,7 @@ enyo.kind({
 				width: FoodChain.sprites.snake.dx, height: FoodChain.sprites.snake.dy, index: 0, sound: "audio/snake"
 			});
 			snake.alive = false;
+			snake.spawnedOnce = false;
 			this.snakes.push(snake);
 		}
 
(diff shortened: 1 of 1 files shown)


GAMinsect:
If instead you want me to strictly adhere to the contribute guideline, I'll remove the fix for the snake clipping when it has nowhere to go, and open a new issue about it, to which I'll work on after finishing this one.

The reason I added that fix was because when playing, often the only free space is the spawn and so by denying the snakes to be there, the game would frequently fall in the issue explained above, leading to poor user experience.

GAMinsect:
Now the frog png is redrawn every time a fly is drawn, to make sure that it isn't overwritten by them

GAMinsect:
In the end I decided to open a new issue regarding snake spawning, to better conform to the repository guideline.
@llaske let me know what you think

llaske:
Nice. Thanks.

GAMinsect:
Thank you

PR #2037, opened 2026-03-19
Fix eraser tool sending wrong coordinates in shared sessions

### LINKED ISSUE #2022: [Bug] Eraser tool sends incorrect coordinates in shared sessions
https://github.com/user-attachments/assets/cef42672-b450-477e-95cb-4a6c0fb5bc79

When using the eraser tool in a shared Paint activity, collaborators see broken disconnected dots instead of a smooth eraser stroke. 

### Steps to Reproduce

1. Open the Paint activity and start a shared session with at least two users.
2. User A draws a line on the canvas using the Pen tool.
3. User A switches to the Eraser tool and drags across the canvas in a continuous stroke.
4. Look at User B's screen.

### Expected Behavior

User B should see the same smooth eraser stroke that User A is drawing.

### Actual Behavior

User B sees broken dots and dashes instead of a continuous stroke. This happens because the eraser is broadcasting the wrong starting coordinates for each segment of the stroke. Every segment starts from a point one pixel away from the current cursor position, instead of from where the previous segment ended.

Bhumika-17:
Fixes #2022

In modes-eraser.js, the onMouseDrag function was using event.point.x + 1  and event.point.y + 1 as the from coordinates when broadcasting to  collaborators. 

Fixed by using PaintApp.modes.Eraser.point.x and PaintApp.modes.Eraser.point.y instead, matching how modes-pen.js handles the same thing.

### DIFF
--- a/activities/Paint.activity/js/modes/modes-eraser.js
+++ b/activities/Paint.activity/js/modes/modes-eraser.js
@@ -71,9 +71,11 @@ define([], function() {
             lineWidth: PaintApp.data.size,
             lineCap: 'round',
             strokeStyle: '#fff',
+            /*Use the previously tracked eraser position instead of the current point,
+             so collaborators receive correctly connected stroke segments */
             from: {
-              x: event.point.x + 1,
-              y: event.point.y + 1
+              x: PaintApp.modes.Eraser.point.x,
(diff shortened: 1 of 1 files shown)


llaske:
Nice. Thanks.

Issue #2040, opened 2026-03-19
Saved Chess game fails to restore difficulty setting


Rejah-Rabeeul:
When reopening a saved Chess game from the journal, the difficulty slider resets to an invalid value instead of the saved setting. As a result, the AI difficulty is not restored correctly and defaults to an incorrect value.

Preview:

https://github.com/user-attachments/assets/bec4b747-46aa-4c95-9917-159d1866ec0f

I’ve identified a fix for this issue and can implement it quickly if approved.

llaske:
You're right it will be better to keep the level.

llaske:
Fixed in https://github.com/llaske/sugarizer/pull/2041

PR #2041, opened 2026-03-20
Fix : Saved Chess game fails to restore difficulty setting

Rejah-Rabeeul:
Issue: #2040 

Preview:

https://github.com/user-attachments/assets/e9a50ee0-832a-4936-9103-835cfe77b8ea


llaske:
Nice. Thanks.

PR #2044, opened 2026-03-20
Improve Calligra performance by throttling move events

sandylangat:
This PR improves performance in the Calligra activity by throttling  mousemove/touchmove event handling.

Previously, drawing logic was triggered on every movement event, causing  frequent redraws and noticeable lag during continuous drawing.

This change limits updates to ~60fps and reduces redundant calculations,  resulting in smoother user interaction.


karthik-dev56:
Hey @sandylangat After approval from mentainer you need to raise the PR. by the way you need to raise the PR to dev branch not to master

PR #2045, opened 2026-03-20
Improve mouse move event handling performance

sandylangat:
Optimize mouse move event handling to limit updates to approximately 60 frames per second.


llaske:
You're not supposed to send a PR without an issue approved by a maintainer.
You must not send PR to the master branch.

PR #2052, opened 2026-03-22
Fix: Copy tool works in all drag directions in Paint activity

SergeantJade:
### Problem
The copy tool in Paint activity only worked when dragging from top-left to bottom-right. Reverse dragging resulted in no selection.

### Solution
- Updated selection logic to support all drag directions
- Used Math.min and Math.abs to normalize coordinates
- Adjusted selection box positioning accordingly

### Testing
- Verified copy works when dragging:
  - top-left → bottom-right
  - bottom-right → top-left
  - other directions


karthik-dev56:
Hey @SergeantJade you should not raise PR without maintener approval and also you should commit your changes to dev not to master

Issue #2049, opened 2026-03-21
Paint Activity: Copy tool only works when dragging top-left to bottom-right


karthik-dev56:
## Bug Description

The Copy tool in Paint activity only works when the user drags from **top-left to bottom-right**. Dragging in any other direction (right-to-left, bottom-to-top, or diagonally opposite) results in nothing being copied.

## Steps to Reproduce

1. Open Paint activity
2. Draw something on the canvas
3. Select the **Copy** tool from the toolbar
4. Try to select an area by dragging from **bottom-right to top-left**
5. Observe that the selection box collapses to zero size
6. Release mouse - nothing is copied, Paste tool has no data

## Expected Behavior

The Copy tool should work regardless of drag direction - users naturally drag in any direction to select areas.

## Actual Behavior

- Dragging top-left → bottom-right: **Works**
- Dragging top-right → bottom-left: **Fails**
- Dragging bottom-left → top-right: **Fails**
- Dragging bottom-right → top-left: **Fails**

https://github.com/user-attachments/assets/ee5fa2d9-1fd3-4871-865b-abd75ab03e3f

## Root Cause

In `activities/Paint.activity/js/modes/modes-copy.js`:

**Line 64-73** - `onMouseDrag` sets width/height to 0 when dragging in reverse:
```javascript
if (begin.x <= end.x) {
    PaintApp.data.currentElement.element.style.width = end.x - begin.x + 'px';
} else {
    PaintApp.data.currentElement.element.style.width = 0 + 'px';  // Bug: should handle reverse
}
```

**Line 86** - `onMouseUp` only processes when `begin < end`:
```javascript
if (begin.x < end.x && begin.y < end.y) {  // Bug: excludes 3 of 4 drag directions
```

karthik-dev56:
Hey @llaske Please have a look when you have time

llaske:
You're right, it could be improved.

llaske:
Fixed in https://github.com/llaske/sugarizer/pull/2055

PR #2055, opened 2026-03-23
Fix copy selection when dragging from top-left to bottom-right

karthik-dev56:
The copy selection in Paint.activity now works correctly in all drag directions by normalizing the selection coordinates using Math.min/max.

Fixes: #2049


karthik-dev56:
Hey @llaske Please have a look

llaske:
Nice. Thanks.

PR #2084, opened 2026-03-29
Fix: prevent crash when jsonData is null in dictstore.js

RichaSingh1505:
This PR adds a null check to the JSON.parse call in js/dictstore.js. When the Datastore is empty, jsonData returns null, causing a crash. Adding a fallback empty object string "{}" ensures the app initializes safely.


llaske:
You're not supposed to send a PR without an issue has been approved by a maintainer.

Issue #2072, opened 2026-03-26
PhysicsJS newton/gravity settings are not saved after leaving


GAMinsect:
https://github.com/user-attachments/assets/2b82cab1-98ea-479c-b828-413625a5cad5

As seen in the video, in PhysicsJS the clear button doesn't clear the canvas, and the gravity options aren't saved when leaving the activity.

Let me know what you think

llaske:
The clear button allow to remove item one by one: click the clear button, then click on item to remove.

Regarding gravity it's true, it will be better to save the settings.

GAMinsect:
I'll work on it and then open a PR

llaske:
Fixed in https://github.com/llaske/sugarizer/pull/2088

PR #2088, opened 2026-03-29
Fixing Gravity and Netwon settings not being saved after leaving the activity

GAMinsect:
Fixing #2072 

Let me know what you think


llaske:
When opening a previous (before the fix) saved instance, there is an error and the pen icon is not displayed

image

GAMinsect:
I think that i know where it is coming from, probably its the fact if the data.gravity is undefined in old instances so i need to account for that.

Also,i think you mean the gravity icon, since the I can see the pen icon on the screen you sent.

Thanks for the feedback

GAMinsect:
Done it, it was what I anticipated in the comment above, that caused that issue.

llaske:
Nice. Thanks

PR #2092, opened 2026-03-29
fix(GetThingsDone): handle null or invalid datastore data before JSON…

### LINKED ISSUE #2076: [BUG]JSON.parse(data) can crash when data is NULL
- The function directly uses JSON.parse(data) without validating whether data is null, undefined, or a valid JSON string. If data is null or invalid, JSON.parse() throws a runtime error, which can crash the application or prevent items from loading.

Image

- This can cause the application to crash if the stored data is:

abelshinev:
Fixes #2076

Handle null or invalid datastore data before calling JSON.parse in GetThingsDone.

This prevents a runtime error when the activity loads with empty, missing, or malformed saved data.

Tested with valid saved todos and with null data fallback.


llaske:
Not a bug

Issue #2090, opened 2026-03-29
Duplicate 'width' property in actionButtons across Tangram activity


karthik-dev56:
## Bug Description

The Tangram activity has a bug where the `actionButtons` object has a duplicate `width` property instead of `height`. This causes the `height` property to be `undefined`, leading to button sizing issues in the UI.

## Code Issue

```javascript
// Current bug code
actionButtons: {
  width: 30,
  width: 30,
},
```

- Description:

Affected Files
activities/Tangram.activity/js/SettingEditor.js (line 137)
activities/Tangram.activity/js/Game.js (line 224)
activities/Tangram.activity/js/DatasetList.js (line 148)
activities/Tangram.activity/js/Leaderboard.js (line 88)
activities/Tangram.activity/js/CategoryForm.js (line 52)
activities/Tangram.activity/js/Result.js (line 135)

- Impact
When the code references actionButtons.height, it returns undefined instead of 30, causing action buttons to render with incorrect dimensions.

llaske:
I don't see the problem. Could you explain how to reproduce it?

karthik-dev56:
Hey @llaske,

There's no visual steps to reproduce - it's a typo in the code.

If you open js/Game.js at line 222, you'll see width is written twice instead of width and height. The same typo exists in the other 5 files I listed.

llaske:
Ok

llaske:
Fixed in https://github.com/llaske/sugarizer/pull/2099

PR #2099, opened 2026-04-01
Fix duplicate width property in actionButtons across Tangram activity

karthik-dev56:
Change the second 'width' to 'height' in actionButtons object. This fixes the bug where actionButtons.height was undefined.

Fixes: #2090


llaske:
Thanks

PR #2110, opened 2026-04-04
Fix: Clear animation interval on stop to prevent Speak Activity memory leak

### LINKED ISSUE #2109: Bug: Memory leak in Speak activity - setInterval never cleared on activity close
## Bug Description

In `activities/Speak.activity/js/SpeakActivity.js`, a `setInterval` is created for the canvas animation loop but is never cleared when the activity closes, causing a memory leak.

## Location

**File:** `activities/Speak.activity/js/SpeakActivity.js`  
**Lines:** 74-77

## Current Code (Buggy)

```javascript
var FPS = 30;
setInterval(function() {
  updateCanvas();
}, 1000/FPS);
```

## Description:
- Problem
- The interval ID is not stored in a variable
- No cleanup when activity is closed (stop button clicked)
- The animation loop continues running **forever** in the background
- Causes memory leak and unnecessary CPU usage
## Expected Behavior
- The interval should be cleared when the activity is stopped/closed.

CodeBySayak:
### Problem
In the Speak Activity (`SpeakActivity.js`), a `setInterval` was used to run a 30 FPS canvas animation loop, but its return ID was not stored. As a result, the interval could not be cleared when the activity was closed via the stop button. This caused the animation loop to continue running in the background, leading to unnecessary CPU usage and a potential memory leak.

### Solution
- Captured the `setInterval` return ID in a new variable called `canvasInterval`.
- Added an event listener triggered by the `stop-button` to correctly run `clearInterval(canvasInterval)` when the user hits stop, successfully cleaning up the animation pipeline and eliminating the memory leak.

### How to verify
1. Open the Speak activity. 
2. Play around with the face drawing interface. 
3. Stop the activity and profile the Javascript VM's memory. Memory usage should plateau normally and stop spiking, and the CPU footprint of the `updateCanvas` loop should vanish instead of continuing to run out of view.

fixes #2109

### DIFF
--- a/activities/Speak.activity/js/SpeakActivity.js
+++ b/activities/Speak.activity/js/SpeakActivity.js
@@ -72,9 +72,12 @@ define(["sugar-web/graphics/palette","sugar-web/env","l10n","sugar-web/datastore
 				document.captureEvents(Event.MOUSEMOVE)
 			}
 			var FPS = 30;
-			setInterval(function() {
+			var canvasInterval = setInterval(function() {
 			  updateCanvas();
 			}, 1000/FPS);
+			document.getElementById("stop-button").addEventListener('click', function() {
+				clearInterval(canvasInterval);
+			});
 			window.addEventListener('localized', function() {
 				if (first) {
 					l10n.init(sugarSettings.language);


karthik-dev56:
Hey without maintener approval you should not raise the PR here also should not raise in master branch u should raise in dev branch.. and also the issue is mine here if he approved it is bug only I can raise the bug try to read Contribution guidlince thank you..

CodeBySayak:
Hi @karthik-dev56, thanks for pointing that out. I appreciate the clarification regarding the contribution guidelines and branch usage. I’ll make sure to follow the correct process (raising PRs against the dev branch and waiting for maintainer approval) going forward. Regarding the issue ownership, I understand your concern. I picked it up assuming it was open for contribution. Thanks again for the guidance!

Issue #2097, opened 2026-03-31
Undefined titles and logging occurs without pressing Record in Measure Activity


cherylpinto:
### Description
When reopening a CSV file from the Journal in the Measure Activity, the resulting entries are displayed with the title **"undefined"**.

Additionally, logging occurs even without pressing the **Record** button. According to the activity tutorial, logging should begin only after the Record button is pressed. However, entries are still being recorded and saved in the CSV file without explicitly starting recording.

### Steps to Reproduce
1. Open Measure Activity  
2. Do not press the Record button  
3. Speak for frequency variation
4. Save/export the CSV
5. Open CSV saved in journal
6. Perform any action (e.g., vizualize/ scroll)  
7. Go back to journal again

### Expected Behavior
- Journal entries should display a valid title  
- Logging should start **only after pressing the Record button**  

### Actual Behavior
- Journal entries show title as **"undefined"**  
- Data is logged and saved in CSV even without pressing the Record button  

https://github.com/user-attachments/assets/32b63912-af89-44b6-a55d-6df61e304b79

About Logging in help Tutorial

<img width="474" height="433" alt="Image" src="https://github.com/user-attachments/assets/7504e31c-1579-43f6-a85e-213a4f31e306" />

<img width="473" height="529" alt="Image" src="https://github.com/user-attachments/assets/ab7c2961-726a-45d6-b78b-8ffbb65d50f1" />

<img width="544" height="336" alt="Image" src="https://github.com/user-attachments/assets/57291ce5-3efa-4f8f-ade5-d55f553c1e90" />

### Additional Context
I would like to contribute to resolving this issue.  
@llaske Could you please share your thoughts on whether this behavior is expected or if it should be fixed?

cherylpinto:
Hi @llaske, just checking in on the issue I opened, would appreciate it whenever you have time to review it. Thanks!

cherylpinto:
Hi @llaske requesting you to please review my issue.  Thank you.

karthik-dev56:
Hey @cherylpinto 
 the "undefined" title happens at line 863 when currentenv.user.name isn't available.
However, Issue 2 needs clarification. Looking at the code:
Recording only starts when startRecord() is called (line 1027)
Data is only pushed to log_data in stopRecord() (line 1062)
The real-time waveform display is visualization, not logging
[...]

PR #2117, opened 2026-04-08
fix: undefined titles in Journal entries

cherylpinto:
**Title:** Fix "undefined" Journal title when opening CSV files in Chart Activity

**Issue:** #2097 

**Description:**

**What does this PR do?**
This PR fixes a bug where opening a previously exported CSV file from the Journal changes its entry title to "undefined". **This fix works for both the Sugarizer Electron (desktop) app and the Website version.**

**Root Cause:**
When a CSV file (such as one exported from the Measure Activity) is opened from the Journal, it is handled by the Chart Activity. Upon loading the CSV data, the activity was unintentionally overriding the entry's title using an internal activity name reference that comes back as undefined for external files. When the user closed the activity, this undefined value was saved back to the Journal, overwriting the original filename.

**Changes Made:**
* Updated the Journal data loading logic in the Chart Activity to prioritize the existing metadata title provided by the Journal when handling CSV files.
* Implemented fallback logic to ensure a valid default string is used if the metadata or internal environment properties are missing.
* Ensured the activity's internal title state is updated immediately upon loading so that the correct filename is preserved when the session is saved and closed.

**How to test:**
1. Open the Measure Activity.
2. Export the measure data as a CSV.
3. Go to the Journal and verify the CSV file has a valid title (e.g., "Measure Logging by ...").
4. Click on the CSV file to open it (it will open in the Chart Activity).
5. Stop/Close the Chart Activity.
6. Return to the Journal and verify that the title is still correctly preserved and has not been renamed to "undefined".

I verified this behavior on both the Electron app and the web version

https://github.com/user-attachments/assets/b67a2b5d-8929-437c-886a-cee2245a2fa9

### DIFF
--- a/activities/Chart.activity/js/activity.js
+++ b/activities/Chart.activity/js/activity.js
@@ -303,7 +303,8 @@ const app = new Vue({
 			this.LZ = this.SugarJournal.LZString;
 			if (metadata.mimetype === "text/csv") {
 				this.isCsvFile = true;
-				this.csvTitle = this.currentenv.activityName;
+				this.csvTitle = metadata.title || (this.currentenv ? this.currentenv.activityName : "Chart Data.csv");
+                this.updateActivityTitle(this.csvTitle);
 				this.pref.chartType = "csvMode";
 				const json = CSVParser.csvToJson(data);
 				if (this.$refs.csvView) {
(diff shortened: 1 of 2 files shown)


llaske:
Nice. Thanks.

Issue #2127, opened 2026-04-11
[v2] Error opening CSV file in Journal


llaske:
When the user click in the Journal on a CSV file exported from Write activity or uploaded to the Journal, an error page appears.
It should be forbidden to open a unknown file format like in the current version on master branch.

https://github.com/user-attachments/assets/7ce7f3ef-5e86-495e-ae41-fe0bb8e4fd3c

karthik-dev56:
Hey @llaske I will fix this working on it now

karthik-dev56:
Hey @llaske, sorry about the confusion earlier  I had opened #2128 without rebasing first, which ended up pushing some unintended commits. I’ve closed that PR and  reopened a clean one after properly rebasing.

llaske:
Fixed in https://github.com/llaske/sugarizer/pull/2129

PR #2129, opened 2026-04-11
fix: prevent crash by forbidding launch of non-PDF generic files

karthik-dev56:
fixes ; #2127 

https://github.com/user-attachments/assets/57fc370e-316a-4ede-8fc3-f8b532a465e8


llaske:
I don't understand why you've updated Write activity. It seems to be a Sugarizer Core issue.

karthik-dev56:
Hey @llaske really sorry about that, I had an extra local test file bundled in by accident! I've removed the Write.activity changes and force-pushed

llaske:
Nice. Thanks.

Issue #2125, opened 2026-04-11
Write activity .txt export squashes text and loses paragraph spacing


cherylpinto:
Description:
When exporting a document to plain text (.txt) in the Write activity, all paragraph breaks and spacing between formatted blocks are completely lost. Words and lines merge together into a single continuous string (e.g., "Hello!Welcome" instead of keeping line breaks or spaces).

Steps to Reproduce:

1. Open the Write activity.
2. Type multiple paragraphs or lines of text.
3. Export the document using the .txt export option.
4. Open the resulting file and observe that all line breaks and block spacing are missing.

Current behaviour

https://github.com/user-attachments/assets/9068e7dc-098a-430d-9bc2-dae24a63e342

Expected Exported Layout

<img width="1004" height="192" alt="Image" src="https://github.com/user-attachments/assets/5b7feb99-fa4c-44ca-be32-8bff4a38e440" />

llaske:
Yes that's true, it could be better.

karthik-dev56:
Hey @cherylpinto, I can see the issue on my side too.
It’s happening in js/activity.js around line 384 during the .txt export. Right now it uses:
var inputData = document.getElementById("editor").textContent;
This removes the spacing (like line breaks), so all the text gets joined together.
Since this uses the Quill editor, we can fix it by using:
var inputData = editor.getText();
This keeps the line breaks.

cherylpinto:
> Yes that's true, it could be better.

Okay, I will fix this issue!

llaske:
Fixed in https://github.com/llaske/sugarizer/pull/2130

PR #2130, opened 2026-04-12
fixed text editor spacing issues

cherylpinto:
Description:
Fixes a bug in the Write activity where exporting to .txt stripped all line breaks, merging paragraphs into a single unreadable block of text.

Changes made:

Updated the plain text export logic to use the editor's native text extraction  rather than the raw DOM .textContent.
This ensures all line breaks, blocks, and structural spacing are correctly preserved in the downloaded text file.

Fixes: #2125 

https://github.com/user-attachments/assets/bd2af00e-35e2-451a-8dc0-2000a145f84c


llaske:
Nice. Thanks.

PR #2149, opened 2026-04-16
Fixing rock making a wall in Foodchain Activity

### LINKED ISSUE #2135: Rocks form a wall in FoodChain Activity
As seen in the video the rocks can create a barrier that doesn't allow the player to move to the other side of the game field.

https://github.com/user-attachments/assets/96743f3e-26c1-4068-8bd5-5da36237fcdf

Although the player could still win by just waiting for the flies to fly in the area it can reach, It don't think it's intended game design.

It should be fixed by adding a new check to the intersect function:
```javascript

// Set randomely rocks
		this.rocks = [];
		for(var i = 0 ; i < FoodChain.playLevels[this.level-1].rocks ; i++) {
			// Ensure distance between blocks and between block and frog is sufficient
			var rock = FoodChain.createWithCondition(
				// Create randomly a new rock...
				function() {
					var x = 130+Math.floor(Math.random()*(FoodChain.playArea.width-300));
					var y = 130+Math.floor(Math.random()*(FoodChain.playArea.height-200));
					var h = Math.floor(Math.random()*4)*90;
					return new Sprite({
						x: x, y: y, heading: h, images: ["rock"], width: FoodChain.sprites.rock.dx, height: FoodChain.sprites.rock.dy, index: 0
					});
				},

				// ... while don't intersect with ...
				function(n, s) {
					return !n.intersect(s); // ADD here something like, && n.distance(s) > 200
				},

				// ... other rocks and frog
				enyo.cloneArray(this.rocks).concat(this.frog)
			);
			this.rocks.push(rock);
		}
```

let me know what you think.

GAMinsect:
Fixes #2135

The chance of rocks creating a wall that separates the frog from the rest of the arena is now almost zero. During testing, this only happened once, and it only affected a small portion of the bottom edge rather than the majority of the area.

The reason I still allow rocks to spawn under a certain threshold is to maintain "clumping," rather than simply forcing them to be as spread out as possible.

```javascript
return !n.intersect(s) && (n.distance(s,true) <= 100 || n.distance(s,true) >= 300);
```

If a valid position for a rock cannot be found after 24 attempts, it will not spawn. This means that when trying to spawn more than three rocks, there is a chance the fourth one won't appear.

I calculated the distance between rocks using a modified Manhattan distance (maximum norm). Since the frog cannot move diagonally, a long diagonal Euclidean distance could fall short, the rocks might still be too close for the frog to navigate between them using only four directions.

Please let me know if you would like me to address any shortcomings of this solution (such as rocks occasionally failing to spawn). 
I didn’t implement fixes for those specific issues yet to avoid unnecessarily complicating the code.

### DIFF
--- a/activities/FoodChain.activity/playgame.js
+++ b/activities/FoodChain.activity/playgame.js
@@ -151,13 +151,18 @@ enyo.kind({
 
 				// ... while don't intersect with ...
 				function(n, s) {
-					return !n.intersect(s);
+					return !n.intersect(s) && (n.distance(s,true) <=100 || n.distance(s,true) >= 300);
 				},
 
 				// ... other rocks and frog
-				enyo.cloneArray(this.rocks).concat(this.frog)
+				enyo.cloneArray(this.rocks).concat(this.frog),
+
+				// We are trying to spawn a rock
+				true
 			);
-			this.rocks.push(rock);
+			if (rock != null) {
+				this.rocks.push(rock);
+			}
 		}
 
 		// Set randomely flies
--- a/activities/FoodChain.activity/playgame.js
+++ b/activities/FoodChain.activity/playgame.js
@@ -151,7 +151,7 @@ enyo.kind({
 
 				// ... while don't intersect with ...
 				function(n, s) {
-					return !n.intersect(s) && (n.distance(s,true) <=100 || n.distance(s,true) >= 300);
+					return !n.intersect(s) && (n.distance(s,true) <= 100 || n.distance(s,true) >= 300);
 				},
 
 				// ... other rocks and frog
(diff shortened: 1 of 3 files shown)


llaske:
I'm not fully convinced. The right root cause (rock placement spacing), but there are two logic issues in the patch that make the fix unreliable:
* In sprite.js, the new “Manhattan-like” distance uses Math.max(dx, dy) instead of Math.max(Math.abs(dx), Math.abs(dy)). Negative deltas can bypass the spacing check.
* In util.js, the retry limit logic/comments are inconsistent (isRock ? 12 : 24, but text says 24 for rocks), so behavior does not match the stated fix.

GAMinsect:
yeah you are right I mixed the ternary operator, and I'll add Math.abs.

Could you be more specific to what doesn't convince you about the solution so that I can fix it?

Issue #2140, opened 2026-04-13
Pointer interaction and hitboxes misaligned for PhysicsJS Activity


cherylpinto:
### Description
In the PhysicsJS activity, pointer interactions (like clicking to delete an object) are misaligned with the visual representation of the objects on the canvas. To interact with an object, especially smaller ones, you have to click slightly outside or below it rather than directly on it.

### To Reproduce

- Open the PhysicsJS 
- Add a few small objects (like the circle or box) to the canvas.
- Select the "Delete" tool (or try to grab/interact with an object).
- Click directly on the center of the object.
- Notice the object is not deleted or targeted. Click slightly below/outside the object to see the hit register.
- Expected behavior
- Clicking directly onto an object's visual bounds should accurately trigger the interaction (grab/delete) regardless of the screen's pixel ratio.

https://github.com/user-attachments/assets/62f512a4-af76-43ec-8c88-9c8726abb84e

@llaske Can you check and let me know if this issue is valid?

karthik-dev56:
Hey @cherylpinto I have gone through this Its working fine on my side

https://github.com/user-attachments/assets/5683c718-ff0f-40b7-9e61-05aa17c6aa37

cherylpinto:
@karthik-dev56 thanks for informing, maybe its a screen display issue from my side! Will close the issue.

karthik-dev56:
Hey @cherylpinto, please don't close the issue It's a real bug we need to fix. I noticed that by reviewing code and texting with my friends laptop
Because you have a high-resolution screen, the app is automatically scaling down the visuals to keep them looking sharp. However, the invisible "hitboxes" that detect your mouse clicks aren't scaling down to match.
Since the hitboxes and the graphics no longer line up, you end up having to click slightly outside the object to grab it. 

If you need i can help u in fixing this 
Actually Its should work on every device 
Also @llaske u can review this its a Bug actually

GAMinsect:
@cherylpinto 
I only have this issue on my desktop.
When running the activity on the VM on my main PC the click behaviour works as intended.
Which strengthens the theory that it has to do with the device aspect ratio.

karthik-dev56:
@GAMinsect 
VMs usually don't use screen scaling, which is why it works perfectly there.
I took a dive into the code and found exactly where the issue is It's coming from a tiny zoom() function inside the activity.js file. This function tries to shrink the graphics to keep them sharp on nice screens, but unfortunately, it throws the hitboxes out of alignment

[...]

PR #2158, opened 2026-04-18
Fix pointer interaction and hitboxes on high DPI screens

cherylpinto:
### Description
Fixed inconsistent hit-detection behavior for the "Clear" tool across different screen resolutions.

### Fixes: #2140 

Previously, on high-DPI screens, the visual canvas scaled correctly but the underlying physics hitboxes did not. As a result, users had to click slightly outside of an object to successfully delete it. 

This fix introduces a custom, resolution-aware distance calculation for the Clear tool. Hitboxes are now perfectly aligned with the visual objects, ensuring smooth and precise deletion across all display scalings and object sizes.

### How to Test (Chrome DevTools)
You can verify the fix works perfectly across different screen scalings (DPR 1, 2, and 3) directly in Chrome:

1. Open Chrome DevTools (`F12` or `Ctrl+Shift+I`) and toggle the **Device Toolbar** (`Ctrl+Shift+M`).
2. Click the three vertical dots (⋮) in the top-right corner of the device toolbar and choose **"Add device pixel ratio"**.
3. A new **DPR** dropdown will appear in the top bar next to the screen dimensions.
4. Set the DPR to **1.0, 2.0, and 3.0** and drop an assortment of small and large objects into the scene.
5. Select the Clear tool and verify that clicking directly on the objects (including their edges or corners) deletes them accurately, without needing to click outside the object 

https://github.com/user-attachments/assets/7c9f121f-4ac1-4536-af7b-b4e0f18171be

### DIFF
--- a/activities/PhysicsJS.activity/js/activity.js
+++ b/activities/PhysicsJS.activity/js/activity.js
@@ -530,11 +530,12 @@ define(["sugar-web/activity/activity","tutorial","l10n","sugar-web/env"], functi
 				}
 				var canvas = document.getElementById("viewport").children[0];
 				var zoom = 1.0 / window.devicePixelRatio;
-				canvas.style.zoom = zoom;
 				var useragent = navigator.userAgent.toLowerCase();
-				if (useragent.indexOf('chrome') == -1) {
-					canvas.style.MozTransform = "scale("+zoom+")";
-					canvas.style.MozTransformOrigin = "0 0";
+				if (useragent.indexOf('firefox') != -1) {
+					canvas.style.transform = "scale("+zoom+")";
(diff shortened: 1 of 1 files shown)


cherylpinto:
Hey @llaske, would appreciate it if you could take a look at my PR when you have some time. Thanks!

llaske:
It solve the issue on delete feature but it should be solved on item creation also.
The new item is supposed to be created at the pointer position.

cherylpinto:
Okay @llaske will fix it

cherylpinto:
Hi @llaske, I’ve fixed the pointer interaction issue during object creation. Could you please take a look?

llaske:
Nice. Thanks

Issue #2169, opened 2026-04-20
Xmas Light Activity multiplayer not working


GAMinsect:
As shown in the video, when entering from multiplayer into another user xmas activity, the browser page keeps loading indefinitely.

https://github.com/user-attachments/assets/0d683d69-bf69-426d-82a6-ec105331809a

Let me know if you too have the same issue

llaske:
You're right, that's strange.

abelshinev:
Hey @GAMinsect are you currently working on this issue or can I take this

GAMinsect:
@abelshinev yes I am.
Locally I have also implemented the solution, waiting to open a PR to make sure it has no flaws.

abelshinev:
alright thanks for letting me know

llaske:
Fixed in https://github.com/llaske/sugarizer/pull/2173

PR #2173, opened 2026-04-24
Fixing multiplayer not working in XmasLights activity

GAMinsect:
Fixing #2169 

The issued is caused because when a user joins a game, both `initalize` and `onNetworkDataRecieved` in activity.js are called, and when `onNetworkDataRecieved` calls `vm.generateGrid` it is possible that the icons have not yet been loaded by `initialize,` which causes the function to remain in an infinite loop.

Explanation with code:
```javascript

generateGrid: function() {
			let vm = this;
			vm.gridContent = [];
			let height = document.getElementById("body").offsetHeight-(vm.$refs.SugarToolbar.isHidden()?0:55);
			let width = document.getElementById("body").offsetWidth;
			let total = Math.floor((1.0*height)/vm.size)*Math.floor((1.0*width)/vm.size);
			let keys =  Object.keys(vm.activitiesIcons); // <--- If initialize has not loaded all the icons it will be an empty array
			let count = total/keys.length;// <-- If the array is empty we will be dividing by 0, which created Infinity
			for (var i = 0 ; i < count ; i++) { // <-- Now we are iterating an infinite number of times, which results in multiplayer not working
				keys.forEach(function(id) {
					let index = vm.gridContent.length;
					let current = (vm.pattern.length==0?id:vm.pattern[index%vm.pattern.length]);
					let svg = vm.activitiesIcons[current == "random" ? keys[Math.floor(Math.random()*keys.length)]:current];
					let color = (vm.colors.length==0?index%180:vm.colors[index%vm.colors.length]);
					if (color == -1) { color = Math.floor(Math.random()*180) }
					let size = vm.size;
					vm.gridContent.push({
						name: id,
						svg: svg,
						color: color,
						size: size,
						step: (vm.colors.length==0?0:index%vm.colors.length)
					});
				});
			}
		},
```

```javascript

initialize: function() {
			let vm = this;
			return new Promise(function(resolve, reject) {
				let content = [];

				// Load activities list
				_loadActivities().then(function(activities) {
					// Load an convert to pure SVG each icon
					activities.push({directory:"activities/XmasLights.activity",icon:"icons/color.svg",id:"color"});
					...
					activities.push({directory:"icons",icon:"cloud-settings.svg",id:"cloud"});
					let len = activities.length;
					for (let i = 0 ; i < len ; i++) {
						let activity = activities[i];
						_loadIcon("../../"+activity.directory+"/"+activity.icon).then(function(svg) {
							content[activity.id] = svg;
							if (Object.keys(content).length == len) {
								let keys = [...Object.keys(content)];
								keys.sort();
								keys.forEach(function(id) {
									vm.activitiesIcons[id] = content[id]; // <-- It takes time to load all icons
									if (Object.keys(vm.activitiesIcons).length == len) {
										resolve();
									}
								});
							}
						});
					}
				});
				_loadIcon("../../activities/XmasLights.activity/icons/random.svg").then(function(svg) {
					vm.randomIcon = svg;
				});
			});
		},
```

Now we wait for `initialize` to finish before calling `transfer` (the function that transfers data from the sender to the receiver), unless we have already loaded all icons in that case we just call `transfer`

Let me know what you think.

### DIFF
--- a/activities/XmasLights.activity/js/activity.js
+++ b/activities/XmasLights.activity/js/activity.js
@@ -557,6 +557,17 @@ var app = new Vue({
 		// Handle network events
 		onNetworkDataReceived: function(msg) {
 			let vm = this;
+			if (Object.keys(vm.activitiesIcons).length == 0) {
+				vm.initialize().then(function() {
+					vm.transfer(vm,msg);
+				});
+			}
+			else{
+				vm.transfer(vm,msg);
+			}
+		},
+
+		transfer: function(vm,msg){
 			vm.pattern = msg.pattern;
 			vm.colors = msg.colors;
 			vm.size = msg.size;


llaske:
Nice. Thanks

PR #2172, opened 2026-04-23
Fix file name truncation in Journal UI by making title display flexible

karthik-dev56:
fixes : #2166 

https://github.com/user-attachments/assets/e27877d9-fafe-4930-815d-c0b42461e7f3

PR #2176, opened 2026-04-26
Fix version label alignment and input field width regressions from #2172

karthik-dev56:
fixes : #2172 

https://github.com/user-attachments/assets/a2119678-ce3b-49a6-b3e2-0fb4ad003327


karthik-dev56:
Hey @llaske I have fixed it

Issue #2175, opened 2026-04-24
Large objects in PhysicsJS water mode never reach Very Heavy density


karthik-dev56:
buoyancy simulation doesn't scale properly for really large shapes. Even if you draw a massive rectangle, it doesn't sink as fast as it should.

There is a small logic error in the `getDensityBySize` function. The code checks if an object is larger than 80 pixels to assign it a **"Heavy"** density. Immediately after that, it tries to check if the object is larger than 100 pixels to make it **"Very Heavy."**

The problem is that any object 100 pixels or larger is *already* larger than 80. Because of how the `if/else` chain is ordered, the "Heavy" check always executed, and the "Very Heavy" branch is never reached it's essentially dead code.

### The Code
You can find this in `activities/PhysicsJS.activity/js/activity.js` near line 228:

<img width="772" height="171" alt="Image" src="https://github.com/user-attachments/assets/8d4eceb4-4fc1-4a69-9007-5ea25a4594f4" />

## Steps to Reproduce

1. Open PhysicsJS and turn on Water mode.
2. Draw a very large rectangle (bigger than 100px).
3. Even though it's huge, it behaves exactly like a much smaller "Heavy" object because the density is capped at 0.0018.

https://github.com/user-attachments/assets/60876693-4f1b-49fb-8949-8f4b2da5e176

Sample Code to check in Browser console 
```javascript
world.getBodies().forEach(b => {
    let size = Math.max(b.geometry.width || 0, b.geometry.height || 0);
    if (size >= 100) console.log(`Size is ${size}, but density is stuck at ${b.density}`);
});

```
**The Fix should** : 
The checks simply need to be swapped so the more specific condition (>= 100) is evaluated before the more general one (> 80).

llaske:
That's true. Look like it's the same for convex-polygon.

karthik-dev56:
@llaske sure i will look at that even

llaske:
Fixed in https://github.com/llaske/sugarizer/pull/2177

PR #2177, opened 2026-04-26
fixes large objects sinks in waterMode in Physicsjs

karthik-dev56:
fixes : #2175 

https://github.com/user-attachments/assets/1332836e-8b9a-4c20-823b-15d8a6e7dacc

Update Sample Code to check in Browser console
```javascript
world.getBodies().forEach(b => {
    let size = 0;
    if (b.geometry.name === 'rectangle') {
        size = Math.max(b.geometry.width || 0, b.geometry.height || 0);
    } else if (b.geometry.name === 'circle') {
        size = b.radius || 0;
    } else if (b.geometry.name === 'convex-polygon') {
        const q = b.geometry.vertices[0];
        size = Math.sqrt(q.x * q.x + q.y * q.y);
    }
   console.log(`${b.geometry.name} — size: ${size.toFixed(1)}, density: ${b.density}`);
});

```


llaske:
Nice. Thanks.

PR #2181, opened 2026-05-06
Fix: Incorrect country labels and flags when coloring neighboring countries

### LINKED ISSUE #2179: Incorrect country labels and flags appear when coloring neighboring countries in Color My World activity
Description:
In the Color My World activity, there seems to be an issue with how countries are identified after coloring.

Steps to reproduce:

Open Color My World activity
Select and color India
Then select and color Bhutan

Expected behavior:
Only the selected country (Bhutan) should display its correct name and flag.

Actual behavior:
After coloring Bhutan, flags and names of other countries (such as China and Bangladesh) also appear incorrectly.

Impact:
This affects the accuracy of the activity and may confuse children learning geography.

https://github.com/user-attachments/assets/d4b261b6-32bf-4b80-b8c7-7b9a1ec81a1d

https://github.com/user-attachments/assets/81f729ea-5803-4993-bd1a-a9acf80b95bc

nithinshiju-dev:
### Description

This PR fixes an issue in the Color My World activity where incorrect country names and flags were displayed when coloring neighboring countries.

### Cause

The issue occurred because `forEachFeatureAtPixel` processed multiple overlapping features during a single click event, triggering tooltips for multiple countries.

### Fix

* Stopped iteration after the first valid feature using an early return
* Ensured tooltip is displayed only for the selected feature

### Result

Now only the selected country shows its correct name and flag when coloring, even near borders.

### Issue
Closes #2179 

https://github.com/user-attachments/assets/fdbf18a3-a7bb-4874-a353-2b89aaadaa32

### DIFF
--- a/activities/ColorMyWorld.activity/lib/map.js
+++ b/activities/ColorMyWorld.activity/lib/map.js
@@ -38,9 +38,35 @@ define(["activity/ol","print","util","colormyworld","humane","flag","l10n"],
 						return;
 					}
 
-					if (!clickedFeature) {
-						clickedFeature = target_feature;
+					clickedFeature = target_feature;
+
+					if (!me.tooltipDisplay || me.tooltipDisplay != name) {
+						me.tooltipDisplay = name;
+						humane.timeout = 1000;
+						humane.log(
+							"<img src='./flags/" + (flag[name.replace(/ /g, '_')] || 'world') + ".svg' " +
+							"style='width:auto;height:1.4em;vertical-align:middle;margin-right:8px;'>" +
+							l10n.get(name.replace(/ /g, '_')).replace(/_/g, ' ')
+						);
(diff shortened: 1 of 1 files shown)


llaske:
I can't reproduce the issue. There is sometime a shift in popup display when we click fast but I don't see an issue.

nithinshiju-dev:
@llaske Could you please try once more using the steps from #2179 ?

When clicking on two adjacent countries (for example, India and then Bhutan), sometimes the country name and flag of other neighbouring countries are also displayed, as shown in the video attached to the issue. I was able to reproduce this consistently on my setup.

If you're still unable to reproduce the issue, I'm happy to close the PR.

llaske:
Okay I've reproduced the problem but your fix doesn't solve the issue.
Because the issue disappear when the user zoom in, I guess it's related to the precision of click so I don't understand how your PR could fix the issue.

PR #2185, opened 2026-05-14
Fix axis label swapping across chart types in shared mode

### LINKED ISSUE #2184: Axis labels get swapped between chart types in shared mode
Description:
In the Chart activity, when using shared mode, editing the X-axis and Y-axis labels in one chart type affects other chart types incorrectly.

For example:
* If User A edits the X and Y labels in the Vertical Bar Chart, then for User B the Line Chart and Horizontal Bar Chart get the axes swapped (X label becomes Y and vice versa).
* Similarly, editing labels in the Line Chart or Horizontal Bar Chart causes the Vertical Bar Chart labels to become reversed for other users.

Steps to reproduce:

1. Open Chart activity in shared mode with two users.
2. Select Vertical Bar Chart.
3. Edit X-axis and Y-axis labels.
4. Observe other chart types on the second user's side.
5. Axis labels appear swapped or incorrectly mapped.

Expected behavior:
Each chart type should maintain its own correct axis mapping without affecting other chart types during synchronization.

Actual behavior:
Editing axis labels in one chart type causes other chart types to display incorrect or reversed axis labels for other users in the shared session.

https://github.com/user-attachments/assets/991d6519-ae3e-45ec-9b65-f2cc0a31398b

nithinshiju-dev:
## Summary

Fixes an issue where axis labels became swapped across chart types during shared sessions.

## Problem

The application was directly swapping `pref.labels.x` and `pref.labels.y` when switching between vertical and horizontal bar charts.

Since `pref.labels` is part of the shared synchronized state, the modified labels propagated to other users and chart types, causing incorrect axis mappings.

## Root Cause

Axis labels were being mutated inside `setChartType()` instead of being handled at the rendering layer.

This caused:

* reversed labels in other chart types
* inconsistent synchronization across users
* incorrect axis mappings during shared sessions

## Fix

* Removed direct swapping of shared label state from `activity.js`
* Added chart-type-specific axis title mapping inside `ChartView.js`

Horizontal bar charts now visually remap axis labels during rendering without modifying the synchronized shared state.

## Result

* Shared mode synchronization works correctly
* Axis labels remain consistent across users
* Horizontal bar charts still display semantically correct axis titles

Fixes #2184

### DIFF
--- a/activities/Chart.activity/js/ChartView.js
+++ b/activities/Chart.activity/js/ChartView.js
@@ -253,11 +253,26 @@ const ChartView = {
 					x: {
 						...this.chartOptions.scales.x,
 						beginAtZero: true,
+						title: {
+							...this.chartOptions.scales.x.title,
+							text: this.pref.labels.y,
+						},
 						grid: {
-							color: (ctx) => (ctx.tick && ctx.tick.value === 0 ? "rgb(50,50,50)" : "rgb(240,240,240)"),
+							color: (ctx) => (
+								ctx.tick && ctx.tick.value === 0
+									? "rgb(50,50,50)"
+									: "rgb(240,240,240)"
+							),
 						},
 					},
-					y: { ...this.chartOptions.scales.y, offset: true },
+					y: {
+						...this.chartOptions.scales.y,
+						offset: true,
+						title: {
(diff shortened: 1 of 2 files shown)


llaske:
Nice thanks.

PR #2186, opened 2026-05-22
Fix: Paint activity, painting reduces in size on reopen

Rejah-Rabeeul:
Issue  #1944


llaske:
There is another side effect: the fill feature no longer work. It fill the wrong place.

https://github.com/user-attachments/assets/9ef12a76-9ea3-4abb-8ab5-229c963053b1

PR #2188, opened 2026-06-02
[WIP] ConnectTheDots activity

Rejah-Rabeeul:
This is the initial stage of the ConnectTheDots activity. I tested it on both mobile and desktop, and also verified it by installing the APK


llaske:
Nice.

* [x] I wonder if it's a good idea to increase/decrease the number of dots when the window is resize. It will limit the board size so it will hide a part of the board. It will be specifically a problem in number and game mode but even in draw mode it will limit the drawing. See how GridPaint activity works.
* [x] The dots animation could be improved. In Pokpok, there is several dots size depending of the distance to the cursor and the size is not uniform depending of the direction.
image


* [x] I don't like the icon, could you suggest some others

Rejah-Rabeeul:
I've updated the required changes except for the icon. I'll update it soon

Rejah-Rabeeul:
Updated the canvas, dot wave, and activity icon

llaske:
* [x] The screen with figures should be responsive 

image

* [x] Miss a back button on the number board to go back to the figures screen

* [x] The back button in edit mode should be on the board, not on the toolbar. Like in Calligra

image

* [x] Clicking on the settings icon when on number board go back to the figures screen. It should display the edit figure instead.

image

* [x] Edit existing figures feature is not implemented
* [x] Adding a new category is not implemented

llaske:
Nice work.
My remarks:

* [x] The dialog popup is ugly 
image

* [x] New category name size in palette is wrong

image

* [x] When a category has one item only, the item size is bigger than the screen

image

* [x] The category name should be editable in settings mode

image

* [x] New categories/items should persist when the activity is saved in the Journal

llaske:
* [x] When switching to number mode from an existing instance the settings mode is activated

https://github.com/user-attachments/assets/383b793a-9faa-4be0-8902-65efdcb5406a

* [x] When a category is renamed, a new category is created and the old one is still there

https://github.com/user-attachments/assets/1f7c9a81-3ced-482a-baa2-b15d9c0bf82e

llaske:
Nice. Few remarks:

* [x] The timer palette now appears in draw shared mode

Capture d’écran du 2026-07-25 21-12-18

* [x] It's possible to click back button during shared mode so the user who shared can change the figure to complete

Capture d’écran du 2026-07-25 21-09-24

* [x] You do not add a label for figures

llaske:
* [x] In "No timer mode", the Restart button doesn't work

image

* [x] There is a missing icon in leader board screen

image

llaske:
* [x] Figures with several numbers at the same are difficult to read

image

* [x] Some localization doesn't work

image

llaske:
Following are my remarks for the game mode:

* [x] The icon is not adapted: not every users know what is a game pad. It's better to have an icon that show what happens. 
* [x] The game should not start automatically else it will be impossible to wait for other users in shared mode. I think a Play button should be on the toolbar and the game will start only when this button is clicked (and when there is at least one opponent: IA or user) 
* [x] A robot icon should be present on the toolbar, it should allow to ask for an IA opponent. The robot is "off" by default and become "on" when clicked (see how the robot button work in Reflection activity) 
* [x] Colors of path and fill should use the buddy colors, like in number mode. For real opponents the colors should use users colors. For IA you can use a random color pair (in the valid XO colors set). 
* [x] Changing direction should be possible using mouse/touch not only on keyboard.  See how changing direction works in Falabracman activity. 
* [x] The starting position should be on the border. Depending of number of users several borders could be used. If there is more than 4 users, borders should be shared between several users. I guess we can limit the game to 8 users.

llaske:
Some other remarks:

* [ ] I'm surprised that players can go across zone filled by others. I've double-checked: it's possible in paper.io too so that's okay. However the dots should use the color from the latest user. 

Capture d’écran du 2026-08-08 21-13-21

* [x] The reason user win or loose is a bit complex to understand. How could we improve that to let the user understand what happens: opponent crash, use crash, max surface completed, ...

llaske:
My remarks:
* [x] Colors are still wrong. As mentioned earlier colors of path and fill should use the buddy colors, like in number mode

image

* [x] As mentioned earlier, the game icon should be changed 
* [x] As mentioned earlier the reason user win or loose is a bit complex to understand. How could we improve that to let the user understand what happens: opponent crash, use crash, max surface completed, ...
* [x] If an user has more than 50% of the board, he should win right?

image
 
* [x] When more than 4 players are in the game, they should share the same board. No user should be in corner. For example in following example the two players on the left should be at 1/3 Y and 2/3 Y.

image

* [x] The speed of the game is too slow. It should be accelerated and we need to have a palette to change speed. You could reuse the one from game of life activity
* [x] I think the Restart icon could be replaced by the Play icon

llaske:
My remarks:

* [x] Export drawing into Journal should not include dots

image

* [ ] Item text for game in choose mode palette is not localized

image

* [ ] Game end condition and notification are something weird: Two players eliminate the same one, two players eliminate themselves and there are two winners

https://github.com/user-attachments/assets/1afc1d5b-ac20-4b59-a38e-2dc5f6ba1e4c

llaske:
>  Export drawing into Journal should not include dots

That's better like that : :+1: 
image

PR #2190, opened 2026-07-31
Fix incorrect country labels and flags when coloring neighboring countries

### LINKED ISSUE #2179: Incorrect country labels and flags appear when coloring neighboring countries in Color My World activity

nithinshiju-dev:
## Description

This PR fixes an issue where clicking adjacent countries (for example, India and Bhutan) could display incorrect country labels and flags.

The issue was caused by the click hit tolerance. With `hitTolerance` set to `4`, clicks near country borders could match neighboring features. Reducing it to `0` ensures that only the intended country is selected.

## Changes

- Changed `hitTolerance` from `4` to `0` in `activities/ColorMyWorld.activity/lib/map.js`.

## Testing

- Reproduced the issue by clicking adjacent countries.
- Verified that incorrect neighboring country tooltips no longer appear.
- Confirmed that the intended country's tooltip is displayed correctly.

Fixes #2179

### DIFF
--- a/activities/ColorMyWorld.activity/lib/map.js
+++ b/activities/ColorMyWorld.activity/lib/map.js
@@ -57,7 +57,7 @@ define(["activity/ol","print","util","colormyworld","humane","flag","l10n"],
 					return false;
 
 				}, {
-					hitTolerance: 4
+					hitTolerance: 0
 				});
 
 				if (clickedFeature) {


nithinshiju-dev:
@llaske  I tracked the issue to the click hitTolerance. Setting it to 0 instead of 4 prevents neighboring countries from being selected during normal clicks. The only remaining edge case is clicking exactly on the shared border, where both countries may still be detected due to the map geometry.

https://github.com/user-attachments/assets/e5de139f-bf12-43a4-bb3f-a4879282cb82

llaske:
@nithinshiju-dev did you test it on a touch device? I'm afraid reducing hit tolerance could have an impact on touch.

